PR #3, opened 2020-12-12
Use the opaque struct pattern to get types in C.

jsha:
Now the C code can use `*client_config` and `*client_session`, and will
get errors if it mixes them up.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,8 +12,9 @@ type CrustlsResult = c_int;
 pub const CRUSTLS_OK: c_int = 0;
 pub const CRUSTLS_ERROR: c_int = 1;
 
-// We use the opaque struct trick pattern to tell C about our types without
+// We use the opaque struct pattern to tell C about our types without
 // telling them what's inside.
+// https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
 #[allow(non_camel_case_types)]
 pub struct rustls_client_config {
     _private: [u8; 0],
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,5 @@
 #![crate_type = "staticlib"]
-
-use libc::{c_char, c_int, c_void, size_t, ssize_t};
+use libc::{c_char, c_int, size_t, ssize_t};
 use std::io::{Cursor, Read, Write};
 use std::slice;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -13,16 +12,27 @@ type CrustlsResult = c_int;
 pub const CRUSTLS_OK: c_int = 0;
 pub const CRUSTLS_ERROR: c_int = 1;
 
+// We use the opaque struct trick pattern to tell C about our types without
+// telling them what's inside.
+#[allow(non_camel_case_types)]
+pub struct client_config {
```
alex, line 18: I'd prefix these struct names with `rustls_` since they're pretty generic.

src/lib.rs
```diff
@@ -13,16 +12,27 @@ type CrustlsResult = c_int;
 pub const CRUSTLS_OK: c_int = 0;
 pub const CRUSTLS_ERROR: c_int = 1;
 
+// We use the opaque struct trick pattern to tell C about our types without
```
tgeoghegan, line 15: nit: you refer to _the_ opaque struct trick as though I ought to know what is. If you have a reference to some documentation or a blog post explaining this technique, this comment would be a good place for it.

src/main.c
```diff
@@ -86,7 +86,7 @@ make_conn(const char *hostname)
  * the message and return 1.
  */
 int
-send_request_and_read_response(int sockfd, void *client_session,
+send_request_and_read_response(int sockfd, rustls_client_session *client_session,
```
tgeoghegan, line 89: nit: It's a shame that the Rust type doesn't materialize in C as `struct rustls_client_session` as I think that's what the [`curl` style would prefer](https://github.com/curl/curl/blob/master/docs/CODE_STYLE.md). What _does_ this type look like in the generated C header? Is it a typedef to `void *`?

tgeoghegan (approved):

src/lib.rs
jsha: Whoops, thanks for catching! I meant to include a link here. Will add.

src/main.c
jsha: Here's what it looks like in the .h:

```
typedef struct rustls_client_session rustls_client_session;                                                                                                                                                        
```

So I probably could use the `curl` style. I'll give it a try.

PR #4, opened 2020-12-12
Make an enum for rustls_result

jsha:
Builds on top of #3.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -170,7 +170,7 @@ send_request_and_read_response(int sockfd, rustls_client_session *client_session
       }
 
       result = rustls_client_session_process_new_packets(client_session);
-      if(result != CRUSTLS_OK) {
+      if(result != rustls_result_OK) {
         fprintf(stderr, "Error in process_new_packets");
         goto cleanup;
       }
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -238,15 +238,15 @@ do_request(const void *client_config, const char *hostname, const char *path)
     goto cleanup;
   }
 
-  void *client_session = NULL;
-  int result =
+  rustls_client_session *client_session = NULL;
+  rustls_result result =
     rustls_client_session_new(client_config, hostname, &client_session);
-  if(result != CRUSTLS_OK) {
+  if(result != rustls_result_OK) {
```
tgeoghegan, line 244: nit: I don't love the mixed case here (`rustls_result_ERROR`) but I suspect this enum has to be ugly on either the C side or the Rust side, so you probably can't do much. _Unless_ there's some kind of `#[repr]` tag you can put on the Rust enum so that the enum that gets rendered into the generated C header has SCREAMING_SNAKE_CASE? Like [`serde`'s `rename_all`](https://serde.rs/container-attrs.html)?

tgeoghegan (approved):

src/main.c
jsha: Good call. This can be done with cbindgen.toml. Added.

tgeoghegan: That's great, nice find!

PR #5, opened 2020-12-16
Add is_handshaking.

jsha:


src/lib.rs
```diff
@@ -150,6 +150,16 @@ pub extern "C" fn rustls_client_session_wants_write(session: *const c_void) -> b
     }
 }
 
+#[no_mangle]
+pub extern "C" fn rustls_client_session_is_handshaking(session: *const c_void) -> bool {
```
tgeoghegan, line 154: I think this should be rebased on #3 so it can take `rustls_client_session` instead of `c_void`?

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #6, opened 2020-12-17
Fix two bugs around fully draining buffers

jsha:
Once we get a notification from epoll that there are bytes available on a given fd, we need to read all of those bytes, since we won't get notified again. To achieve that, make our connection nonblocking. For each return from epoll, we'll read bytes until we get an EAGAIN or EWOULDBLOCK, then go back into polling.

When copying TLS bytes into the client_session, loop as many times as we need to to ensure all bytes get copied into the session. Previously we were assuming every write to the session accepted all of the bytes in the buffer.

Both of these bugs manifested when downloading large files. The latter bug was dependent on the size of `buf` in main.c. If I reduced `buf` to 2032, I would never hit the bug, presumably because that made the buffers in main.c consistently smaller than the max read that rustls_client_session_read_tls would accept in one call.

This introduces a new result enum, CRUSTLS_DEMO_RESULT. It has three values, so it can indicate OK, ERROR, and AGAIN (used to tell the main loop to go back to polling rather than erroring out). I haven't converted all functions to return this enum, since I wanted to get feedback on it first. If it looks good, I will change all functions in main.c to use it.


src/main.c
```diff
@@ -37,6 +45,17 @@ write_all(int fd, const char *buf, int n)
...
 }
 
+/*
+ * Set a socket to be nonblocking.
+ */
+int
+nonblock(int sockfd)
+{
+  int flags;
+  flags = fcntl(sockfd, F_GETFL, 0);
```
tgeoghegan, line 55: Can this call ever fail? You should check if it returns `-1` before possibly mangling the flags on `sockfd`.

src/main.c
```diff
@@ -67,6 +86,7 @@ make_conn(const char *hostname)
     perror("connecting");
     goto cleanup;
   }
+  nonblock(sockfd);
```
tgeoghegan, line 89: Check the returned value here

src/main.c
```diff
@@ -80,6 +100,129 @@ make_conn(const char *hostname)
...
+ */
+int
+copy_tls_bytes_into_client_session(
+  struct rustls_client_session *client_session, uint8_t *buf, size_t len)
+{
+  ssize_t n;
+  int result;
+
+  while(len > 0) {
+    n = rustls_client_session_read_tls(client_session, (uint8_t *)buf, len);
```
tgeoghegan, line 115: Is the cast to `uint8_t *` necessary since `buf` already had that type?

src/main.c
```diff
@@ -80,6 +100,129 @@ make_conn(const char *hostname)
...
+      return 1;
+    }
+
+    result = write_all(STDOUT_FILENO, buf, n);
+    if(result != 0) {
+      return result;
+    }
+  }
+
+  return 0;
```
tgeoghegan, line 172: nit: Is this `return` reachable? I wonder if some compilers might complain about unreachable code ... but then others might complain about a function without a `return`! Anyway, more constructively, since you never, ever expect to reach this point, you might reasonably return an error or even abort here to make it easier to debug such a strange condition.

src/main.c
```diff
@@ -80,6 +100,129 @@ make_conn(const char *hostname)
...
+
+  bzero(buf, sizeof(buf));
+  n = read(sockfd, buf, sizeof(buf));
+  if(n == 0) {
+    fprintf(stderr, "EOF reading from socket\n");
+    return CRUSTLS_DEMO_ERROR;
+  }
+  else if(n < 0) {
+    if(errno == EAGAIN || errno == EWOULDBLOCK) {
+      fprintf(stderr, "reading from socket: EAGAIN or EWOULDBLOCK\n");
```
tgeoghegan, line 197: nit: consider `strerror(3)` to get a natural language description of `errno`. Then again it can often be nice to have to specific errno. Otherwise it's hard to tell `EACCES` from `EPERM`: "Permission denied" and "Operation not permitted", respectively.

tgeoghegan (changes requested):

jsha:
Thanks, fixed!

tgeoghegan (approved):

PR #9, opened 2020-12-21
Makefile: support DESTDIR

bagder:
To allow installing in a custom (non-root) dir.

Also, no longer uses sudo so use 'sudo make' if needed.


jsha (approved):
Thanks!

PR #10, opened 2020-12-21
Add rustls_version.

jsha:
I chose a "caller provides memory" style of output because that is how the corresponding `version` hook in curl handles things.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,10 @@
 #![crate_type = "staticlib"]
 use libc::{c_char, size_t, ssize_t};
-use std::io::{Cursor, Read, Write};
 use std::slice;
+use std::{
+    cmp::min,
+    io::{Cursor, Read, Write},
+};
 use std::{ffi::CStr, sync::Arc};
 use std::{io::ErrorKind::ConnectionAborted, mem};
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -26,6 +29,33 @@ pub struct rustls_client_session {
...
+#[no_mangle]
+pub extern "C" fn rustls_version(buf: *mut c_char, len: size_t) -> size_t {
+    let write_buf: &mut [u8] = unsafe {
+        if buf.is_null() {
+            eprintln!("rustls_version: buf was NULL!");
+            return 0;
+        }
+        slice::from_raw_parts_mut(buf as *mut u8, len as usize)
+    };
+    let version: String = format!(
```
alex, line 47: I was going to suggest using `write!()` to avoid the allocation, but I can't find a way to get the length from it (https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=aaefe147f79914cf10f86c0bb638db98) and it's a version function so it's very very unlikely to be performance sensitive.

src/lib.rs
```diff
@@ -26,6 +29,33 @@ pub struct rustls_client_session {
     _private: [u8; 0],
 }
 
+// Keep in sync with Cargo.toml.
+const RUSTLS_CRATE_VERSION: &str = "0.19.0";
+
+/// Write the version of the crustls C bindings and rustls itself into the
+/// provided buffer, up to a max of `len` bytes. Output is UTF-8 encoded
+/// and NUL terminated. Returns the number of bytes written before the NUL.
```
alex, line 37: I'd suggest not doing NUL terminated here.

src/lib.rs
jsha: [`write!()` says:](https://doc.rust-lang.org/std/macro.write.html)
> The macro returns whatever the write_fmt method returns; commonly a fmt::Result, or an io::Result.

[`slice` gets its `write_fmt` implementation from the `Write` trait:](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_fmt)

    fn write_fmt(&mut self, fmt: Arguments<'_>) -> Result<()>

src/lib.rs
jsha: `curl` expects NUL-terminated strings when calling `Curl_ssl_version`, which will call this function: https://github.com/curl/curl/blob/master/lib/version.c#L165.

If I don't do NUL termination on this side, I'd just have to add it on the C side. I'd rather do it on the Rust side.

tgeoghegan (approved):

PR #12, opened 2020-12-30
Use output parameters for n, and add conventions to README.

jsha:
This documents some details of how the library is designed and how to
interact with it. In particular it sets a new convention: If a function
or method is fallible, it returns rustls_result. The resulted in a
change to the four read/write functions. Previously they acted like
`read(2)`/`write(2)`, returning an ssize_t. Now, they return a
`rustls_result` and store the number of bytes read/written in an output
argument.

Corresponding changes to my PR branch against curl:
https://github.com/jsha/curl/pull/1. I'll merge that into the branch under
review once this PR lands, so that PR branch will continue to build
successfully against master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,85 @@ To install:
 To build and install in optimized mode:
 
     make PROFILE=release install
+
+# Conventions
+
+This library defines an enum, rustls_result, to indicate success or failure of
+a function call. All fallible functions return a rustls_result. If a function
+has other outputs, it provides them using output parameters (pointers to
+caller-provided objects). For instance:
+
+```rust
+rustls_result rustls_client_session_read(const rustls_client_session *session,
+                                         uint8_t *buf,
+                                         size_t count,
+                                         size_t *out_n);
+```
+
+In this example, `buf` and `out_n` are output parameters.
+
+## Structs
+
+For a given struct, all functions that start with the name of that struct are
+either associated functions or methods of that struct. For instance,
+`rustls_client_session_read` is a method of `rustls_client_session`. A function
+that takes a pointer to a struct as the first parameter is considered a method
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -19,3 +19,85 @@ To install:
...
+```
+
+In this example, `buf` and `out_n` are output parameters.
+
+## Structs
+
+For a given struct, all functions that start with the name of that struct are
+either associated functions or methods of that struct. For instance,
+`rustls_client_session_read` is a method of `rustls_client_session`. A function
+that takes a pointer to a struct as the first parameter is considerd a method
```
tgeoghegan, line 44: "considered"

README.md
```diff
@@ -19,3 +19,85 @@ To install:
...
+on that struct. Structs in this library are always created and destroyed by
+library code, so the header file only gives a declaration of the structs, not
+a definition.
+
+As a result, structs are always handled using pointers. For each struct, there
+is generally a function ending in `_new()` to create that struct. Once you've
+got a pointer to a struct, it's you're responsibility to (a) ensure no two
```
tgeoghegan, line 51: "your"

src/lib.rs
```diff
@@ -313,81 +342,113 @@ pub extern "C" fn rustls_client_session_read(
...
-            return -1;
+            return rustls_result::ERROR;
         }
         slice::from_raw_parts_mut(buf, count as usize)
     };
+    let out_n = unsafe {
+        match out_n.as_mut() {
+            Some(out_n) => out_n,
+            None => {
+                eprintln!("ClientSession::read: out_n was NULL");
```
tgeoghegan, line 440: this should be `ClientSession::write_tls`

tgeoghegan (approved):
Approved, with a couple typos and copy/paste errors

PR #16, opened 2021-01-07
Add configuration of trusted roots

jsha:
This removes the dependency on the webpki-roots crate, which bundles a set of trusted roots, in favor of the rustls-native-roots crate, which provides a way to trust the roots installed in the OS.

This also creates a new type alias `rustls_client_config_builder`, to represent the state of a ClientConfig that may still be mutated (for instance by adding trusted roots to the default, empty, set). Once the user has set up the config as desired, a builder is turned into a `rustls_client_config`, which is not mutable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crustls"
-version = "0.1.0"
+version = "0.2.0"
 authors = ["Jacob Hoffman-Andrews <github@hoffman-andrews.com>"]
 description = "C-to-rustls bindings"
 edition = "2018"
@@ -11,7 +11,7 @@ rustls = "^0.19.0"
 webpki = "0.21"
 libc = "0.2"
 env_logger = "0.8"
-webpki-roots = "0.21"
+rustls-native-certs = "0.5.0"
 
 [dev_dependencies]
 cbindgen = "*"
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
I think it's unfortunate that `rustls_client_config_builder_load_native_roots` and `rustls_client_config_builder_load_roots_from_file` can both return errors, because it's kind of a chore for the caller to have to do error handling for each call to add an option to a builder. It'd be nice if those functions were infallible and did nothing but populate a member in a config object, and then all the operations that can fail (such as loading a file, or a call into `rustls_native_certs::load_native_certs`) are done when the client calls `rustls_client_config_builder_build`, so then they only have to check errors on that one function.

However that approach would obviously require more complicated code in the Rust side, traded off against the volume of error handling on the C side, so I'm comfortable with wherever you want to land on that tradeoff.

jsha:
Ooh, that's an interesting idea to delay error handling to the `build`. One way to do that would be to go ahead and do the work of loading when one of these functions is called, but if the function fails, set an `error` field on the builder object. And then at build time, we would short-circuit if the error field is set.

I think if there turn out to be a lot of fallible methods on the builder, it would make sense to postpone error handling until the end. Right now it's just these two. Looking at the list of other fields methods on ClientConfig, I think many of them would be infallible: https://docs.rs/rustls/0.19.0/rustls/struct.ClientConfig.html#fields. But I guess even something innocuous like setting the list of ciphersuites could fail due to validation like "that's not a ciphersuite that exists."

jsha:
I'm going to go ahead with this as-is to unblock fixes on the curl PR branch, but we can still revisit this if it turns out a lot of inputs to the builder are fallible.

PR #17, opened 2021-01-07
Add rustls_result_is_cert_error.

jsha:
There are a lot of error types that amount to "failed to validate cert."
We should provide an easy way to consider them as a group.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -33,6 +33,37 @@ pub extern "C" fn rustls_error(
     }
 }
 
+#[no_mangle]
+pub extern "C" fn rustls_result_is_cert_error(result: rustls_result) -> bool {
+    match result {
+        rustls_result::CertBadDER => true,
+        rustls_result::CertBadDERTime => true,
+        rustls_result::CertCAUsedAsEndEntity => true,
+        rustls_result::CertExpired => true,
+        rustls_result::CertNotValidForName => true,
+        rustls_result::CertNotValidYet => true,
+        rustls_result::CertEndEntityUsedAsCA => true,
+        rustls_result::CertExtensionValueInvalid => true,
+        rustls_result::CertInvalidCertValidity => true,
+        rustls_result::CertInvalidSignatureForPublicKey => true,
+        rustls_result::CertNameConstraintViolation => true,
+        rustls_result::CertPathLenConstraintViolated => true,
+        rustls_result::CertSignatureAlgorithmMismatch => true,
+        rustls_result::CertRequiredEKUNotFound => true,
+        rustls_result::CertUnknownIssuer => true,
+        rustls_result::CertUnsupportedCertVersion => true,
+        rustls_result::CertUnsupportedCriticalExtension => true,
+        rustls_result::CertUnsupportedSignatureAlgorithmForPublicKey => true,
+        rustls_result::CertUnsupportedSignatureAlgorithm => true,
+        rustls_result::CertSCTMalformed => true,
+        rustls_result::CertSCTInvalidSignature => true,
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -33,6 +33,37 @@ pub extern "C" fn rustls_error(
...
+        rustls_result::CertUnsupportedSignatureAlgorithm => true,
+        rustls_result::CertSCTMalformed => true,
+        rustls_result::CertSCTInvalidSignature => true,
+        rustls_result::CertSCTTimestampInFuture => true,
+        rustls_result::CertSCTUnsupportedVersion => true,
+        rustls_result::CertSCTUnknownLog => true,
+        _ => false,
+    }
+}
+
```
tgeoghegan, line 66: It's a shame that Rust doesn't let you do something like Java enums do, where you can embed values in an enum case, so you could do something like:

```
pub enum rustls_result {
  Io(false),
  CertBadDER(true),
  ...
};
```
...where the boolean value indicates whether the enclosing enum case is a cert error (you could find a more expressive way to encode that than a bool, of course). I just think you might end up with a lot of long functions like this one as you add more attributes to these enum cases, and it's easy to end up with bugs resulting from forgetting to list all the appropriate enum cases in a new `rustls_result_is_<whatever>` function.

tgeoghegan (approved):

src/error.rs
jsha: Well, the Rust does of course allow you to embed values in enums. In your example it would be:

```rust
pub enum rustls_result {
  Io(bool),
  CertBadDER(bool),
}
```

And in fact [TLSError uses these embedded values a lot](https://docs.rs/rustls/0.19.0/rustls/enum.TLSError.html). But `rustls_result` is the C mapping of that enum, so I intentionally flattened all the enums-with-values into one namespace.

The more principled way to do this would be to define a `reverse_map_error` that takes a `rustls_result`, and turns it into a `TLSError`. Then `rustls_is_cert_error` would check if it was the variant `WebPKIError` or `InvalidSCT`.

To some extent, `result_to_string` is this reverse mapping, except instead of mapping into `TLSError` it maps into `String`. I did try initially to write the reverse mapping, but ran into some troubles. Maybe I should give it another try. You're right that giant lists invite bugs due to omitting something.

jsha: Alright, tried it with the reverse mapping. Let me know what you think!

src/error.rs
```diff
@@ -218,141 +227,142 @@ pub(crate) fn map_error(input: rustls::TLSError) -> rustls_result {
...
+// whether a rustls_result is part of some top-level variant that maps to
+// several rustls_results.
+fn result_to_tlserror(input: &rustls_result) -> Either {
     use rustls::internal::msgs::enums::AlertDescription as alert;
     use rustls_result::*;
     use sct::Error as sct;
     use webpki::Error as webpki;
 
-    let msg = match input {
+    match input {
```
tgeoghegan, line 256: I like the `Either` enum, but I'm not sure about the two `match` blocks here. It's not great having all the `unreachable!` cases in the second one. Why not one giant `match`? Was it to avoid having dozens of cases that all get wrapped in `Either::TlsError`?

tgeoghegan (approved):

src/error.rs
jsha: Yep, exactly - I didn't want to have to do tons of wrapping.

tgeoghegan: I don't have a suggestion of how to do better so I'll spare you my aesthetic griping. This change definitely achieves what it intends.

PR #19, opened 2021-01-08
Check in crustls.h.

jsha:
Previously, crustls.h was only generated at build time. However, this
made it hard for visitors to the repo to see what the C API looks like.

Now, crustls.h is checked in, with a CI test to verify the current
version matches what `make` produces.


tgeoghegan (approved):
That's quite a nice header for generated code!

PR #20, opened 2021-01-13
Add InvalidDnsNameError.

jsha:
Covers https://doc.rust-lang.org/std/str/struct.Utf8Error.html and
https://briansmith.org/rustdoc/webpki/struct.InvalidDNSNameError.html


src/error.rs
```diff
@@ -48,6 +48,7 @@ pub enum rustls_result {
     Ok = 7000,
     Io,
     NullParameter,
+    InvalidDnsNameError,
```
tgeoghegan, line 51: Adding this here will shift every other existing error's value by one, right? `Ok = 7000`, hence `Io = 7001` and so on? Are the error enum values part of your ABI with clients? It's likely fine to change these values at this stage but in the future, you would want to be careful about where you add new errors to avoid breaking clients who built against an older header but dynamically link a newer library. Unless you expect clients to always statically link crustls, in which case it doesn't matter.

tgeoghegan (approved):

src/error.rs
jsha: Yeah, agreed this breaks ABI. I'm willing to do it at this point because the library is still under such active development.

I've been intending to explicitly number each of these so it's easier to keep ABI in the future. One thought I had was to group them by component, so this cluster would be 7000, 7001, etc; the next cluster would be 7100, 7101, etc.

Thoughts on that approach? The main downside I can think of is that people would start treating the number ranges as a part of the the API and say, e.g. "7200-7300 means some sort of TLS alert."

tgeoghegan: I like the grouping approach, though of course you might eventually run out of error codes in a particular range. As for people abusing them: I don't know that there's very much you can do about that short of hiding the actual error code in private struct fields, which isn't even really possible in C. I read somewhere that Microsoft randomizes error codes or other integer constants with every release of Windows to force clients to recompile against headers consistent with the DLL they're going to link, but I think that would be overkill here.

PR #26, opened 2021-01-18
looking for TCP explicitly

icing:
Giving hints to getaddinfo() that we want something TCP like. On MacOS Big Sur we get UDP returned first otherwise.


jsha (approved):
Thanks!

PR #28, opened 2021-01-22
Make better use of macros.

jsha:
Add checked_ptr_ref! and checked_ptr_mut! to eliminate the
repeated null-pointer checking code we had at the beginning of each
function.

Revise the ffi_panic_boundary_* macros to be implemented in terms of
ffi_panic_boundary_generic!, which takes a return value as an argument.

Remove some `const` modifiers on parameters that are actually treated
as mutable (rustls_client_session_read, rustls_client_session_write,
etc).

I think overall this improves clarity by removing boilerplate. For instance
the wants_read and is_handshaking methods are much more clearly
plain passthroughs now. But feel free to tell me "you're getting carried
away with macros!"

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,16 +99,21 @@ macro_rules! ffi_panic_boundary_unit {
     }
 }
 
-/// If the provided pointer is non-null, convert it to a reference
-/// to the type in the second argument.
+/// If the provided pointer is non-null, convert it to the reference
+/// type in the second argument.
 /// Otherwise, return NullParameter (in the two-argument form) or the provided
 /// value (in the three-argument form).
+/// Examples:
+///   let config: &mut ClientConfig = try_ref_from_ptr!(builder, &mut ClientConfig,
+///        null::<rustls_client_config>());
+///   let session: &ClientSession = try_ref_from_ptr!(session, &ClientSession);
+///
 #[macro_export]
-macro_rules! checked_ptr_ref {
-    ( $var:ident, $typ:ty ) => {
-        checked_ptr_ref!($var, $typ, rustls_result::NullParameter)
+macro_rules! try_ref_from_ptr {
+    ( $var:ident, & $typ:ty ) => {
+        try_ref_from_ptr!($var, &$typ, rustls_result::NullParameter)
     };
-    ( $var:ident, $typ:ty, $retval:expr ) => {
+    ( $var:ident, & $typ:ty, $retval: expr ) => {
         unsafe {
             match ($var as *const $typ).as_ref() {
                 Some(c) => c,
@@ -116,15 +121,10 @@ macro_rules! checked_ptr_ref {
             }
         };
     };
-}
-
-/// Like checked_ptr, but provides mutable references.
-#[macro_export]
-macro_rules! checked_ptr_mut {
-    ( $var:ident, $typ:ty ) => {
-        checked_ptr_mut!($var, $typ, rustls_result::NullParameter)
+    ( $var:ident, &mut $typ:ty ) => {
+        try_ref_from_ptr!($var, &mut $typ, rustls_result::NullParameter)
     };
-    ( $var:ident, $typ:ty, $retval:expr ) => {
+    ( $var:ident, &mut $typ:ty, $retval:expr ) => {
         unsafe {
             match ($var as *mut $typ).as_mut() {
                 Some(c) => c,
@@ -180,7 +180,7 @@ pub extern "C" fn rustls_client_config_builder_build(
     builder: *mut rustls_client_config_builder,
 ) -> *const rustls_client_config {
     ffi_panic_boundary_ptr! {
-        let config: &mut ClientConfig = checked_ptr_mut!(builder, ClientConfig,
+        let config: &mut ClientConfig = try_ref_from_ptr!(builder, &mut ClientConfig,
              null::<rustls_client_config>());
         let b = unsafe { Box::from_raw(config) };
         Arc::into_raw(Arc::new(*b)) as *const _
@@ -194,7 +194,7 @@ pub extern "C" fn rustls_client_config_builder_load_native_roots(
     config: *mut rustls_client_config_builder,
 ) -> rustls_result {
     ffi_panic_boundary! {
-        let mut config: &mut ClientConfig = checked_ptr_mut!(config, ClientConfig);
+        let mut config: &mut ClientConfig = try_ref_from_ptr!(config, &mut ClientConfig);
         let store = match rustls_native_certs::load_native_certs() {
             Ok(store) => store,
             Err(_) => return rustls_result::Io,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -48,63 +48,100 @@ pub struct rustls_client_session {
...
+/// value (in the three-argument form).
+#[macro_export]
+macro_rules! checked_ptr_ref {
+    ( $var:ident, $typ:ty ) => {
+        unsafe {
+            match ($var as *mut $typ).as_ref() {
+                Some(c) => c,
+                None => return rustls_result::NullParameter,
+            }
+        };
```
alex, line 114: Is there a reason not to implement this in terms of the other variant? So: `checked_ptr_ref!($var, $typ, rustls_result::NullParameter)`?

jsha: Good call, done.

src/lib.rs
```diff
@@ -48,63 +48,90 @@ pub struct rustls_client_session {
...
+/// Otherwise, return NullParameter (in the two-argument form) or the provided
+/// value (in the three-argument form).
+#[macro_export]
+macro_rules! checked_ptr_ref {
+    ( $var:ident, $typ:ty ) => {
+        checked_ptr_ref!($var, $typ, rustls_result::NullParameter)
+    };
+    ( $var:ident, $typ:ty, $retval:expr ) => {
+        unsafe {
+            match ($var as *mut $typ).as_ref() {
```
tgeoghegan, line 113: Does this need to be `as *mut $typ` or would `as *$typ` suffice? Put otherwise: since we're ultimately getting an immutable reference, why do we need a mutable reference to `var`?

src/lib.rs
```diff
@@ -153,7 +180,9 @@ pub extern "C" fn rustls_client_config_builder_build(
     builder: *mut rustls_client_config_builder,
 ) -> *const rustls_client_config {
     ffi_panic_boundary_ptr! {
-        let b = unsafe { Box::from_raw(builder as *mut ClientConfig) };
+        let config: &mut ClientConfig = checked_ptr_mut!(builder, ClientConfig,
+             null::<rustls_client_config>());
+        let b = unsafe { Box::from_raw(config) };
```
tgeoghegan, line 185: 👍🏻 I like that the `unsafe` section is made smaller.

src/lib.rs
```diff
@@ -165,12 +194,7 @@ pub extern "C" fn rustls_client_config_builder_load_native_roots(
     config: *mut rustls_client_config_builder,
 ) -> rustls_result {
     ffi_panic_boundary! {
-        let mut config: &mut ClientConfig = unsafe {
-            match (config as *mut ClientConfig).as_mut() {
-                Some(c) => c,
-                None => return rustls_result::NullParameter,
-            }
-        };
+        let mut config = checked_ptr_mut!(config, ClientConfig);
```
tgeoghegan, line 197: nit: for consistency with other usages of `checked_ptr_ref` and `checked_ptr_mut`, and what I think is your preferred style, you might do `let config: &mut ClientConfig = checked_ptr_mut!(config, ClientConfig);`. On the other hand, that would repeat the type `ClientConfig` twice, so perhaps this usage is better.

src/lib.rs
```diff
@@ -228,24 +247,19 @@ pub extern "C" fn rustls_client_config_builder_load_roots_from_file(
...
-                        "rustls_client_config_free: invariant failed: arc.strong_count was < 1: {}. \
-                        You must not free the same client_config multiple times.",
-                        strong_count
-                    );
-                }
-            } else {
-                eprintln!("rustls_client_config_free: config was NULL");
-            }
-        };
+        let config: &ClientConfig = checked_ptr_ref!(config, ClientConfig, ());
```
tgeoghegan, line 250: Looking at this usage, I don't love that `checked_ptr_ref!` can `return` an error value and bail out of the function, because I don't think the control flow is obvious when I read this function. Since there's no `?` or other error handling or branching, my assumption if I were scanning this code, perhaps seeking the line that produced the error message I found in a log, is that `checked_ptr_ref!` can't fail. What if instead of either evaluating to a `T` value or returning some value, `checked_ptr_ref!` evaluated to `Result<T>`, and then you'd do:
`let config = checked_ptr_ref!(config, ClientConfig, ())?;`
Which I think makes it more obvious that there's a potential failure case on this line. If you end up having to reintroduce the `match` statement at each callsite, then I think that would defeat the point of macro-ing away the checked conversion, but if it's just an extra `?`, I think it's worth it for clarity.

src/lib.rs
```diff
@@ -344,12 +350,8 @@ pub extern "C" fn rustls_client_session_is_handshaking(
     session: *const rustls_client_session,
 ) -> bool {
     ffi_panic_boundary_bool! {
-        unsafe {
-            match (session as *const ClientSession).as_ref() {
-                Some(cs) => cs.is_handshaking(),
-                None => false,
-            }
-        }
+        let session: & ClientSession = checked_ptr_ref!(session, ClientSession, false);
```
tgeoghegan, line 353:
```suggestion
        let session: &ClientSession = checked_ptr_ref!(session, ClientSession, false);
```

src/lib.rs
```diff
@@ -48,63 +48,90 @@ pub struct rustls_client_session {
...
+        ffi_panic_boundary_generic!((), $($tt)*)
+    }
+}
+
+/// If the provided pointer is non-null, convert it to a reference
+/// to the type in the second argument.
+/// Otherwise, return NullParameter (in the two-argument form) or the provided
+/// value (in the three-argument form).
+#[macro_export]
+macro_rules! checked_ptr_ref {
```
tgeoghegan, line 107: Galaxy brain review comment: could you write a single macro that matches on the literal `mut` in front of the `$typ:ty`?

```
macro_rules! checked_ptr {
    ( $var:ident, mut $typ:ty, $retval:expr ) => {
        // handle mutable case with .as_mut()
    };
    ( $var:ident, $typ:ty, $retval:expr ) => {
       // handle immutable case with .as_ref()
    }
}
```

Or maybe there's some other way to write arms that match on either `mut T` vs `T`? Then you wouldn't have to define two nearly identical macros for mut vs. ref. This is definitely code golf, and I'm more curious about whether it's possible than advocating for it as a good idea.

tgeoghegan (approved):
Approving because there's a couple typos that I'm sure you can fix without me doing another review, and then some ideas that you should feel free to ignore, so nothing that should block merging. This is neat, I'm enjoying learning more about Rust macros!

src/lib.rs
jsha: FWIW, it's not really. The `checked_ptr_mut!` implementation is one big unsafe blocks. So now we just have two smaller unsafe blocks rather than one big one.

src/lib.rs
jsha: I would love to do this. `?` only works in functions that return Result or Option. So it wouldn't work in this function that returns unit.

It seems like it *might* be possible for the functions that return rustls_result to return `Option<rustls_result>`, and get Rust to represent `None` as rustls_result::Ok (not sure how I would go about this but would do some research).

Another tweak that would provide much smaller benefit but much cheaper: I could rename these to start with `try_`, to remind people of the `try` macro that predated `?`: https://doc.rust-lang.org/std/macro.try.html. But given that's deprecated, it will probably resonate with fewer people over time.

src/lib.rs
jsha: Good catch. This actually shouldbe `*const $typ`.

src/lib.rs
jsha: This was a cool idea and it worked! I didn't wind up de-duplicating the macro definitions (probably could but I felt like it would be too fancy). But by adding `&` and `&mut` to the matches, the call sites for the macro more clearly express the type being cast to.

PR #30, opened 2021-01-23
Add server support

jsha:
I've mainly adapted the code from the client side, with the main changes being in the configuration code (entirely different, though I've kept the builder pattern) and how sessions are created (no hostname is needed). To see a diff of the two files, try `diff src/client.rs src/server.rs`.

I've considered ways to abstract the common code and reduce duplication. The natural way to do this is to express the `Session` trait in C, since the `Session` methods are mainly what is common between `ServerSession` and `ClientSession`. However, expressing this in C's type system would require C code to do a cast from `rustls_client_session*` to `rustls_session*` and so on. Right now I think the code duplication in this library is preferable to requiring casts on the C side. I'm interested in other ideas.

One notable place where this library diverges from the rustls API: `ServerConfig::set_single_cert` expects its inputs as DER, but this library offers `rustls_server_config_builder_set_single_cert_pem`, which takes them in PEM. There are a few reasons for this. First, a `Vec<Certificate>` (which is really a `Vec<Vec<u8>>`) is awkward to express in C. I would have to take a pointer to an array of pointers to byte arrays, a length of that array, and a pointer to an integer arrays containing the lengths of those byte arrays. PEM allows passing a single byte array, because it is a delimited format. Secondly, PEM is a popular format, and most client code will be parsing PEM anyhow. We may as well do parsing in Rust.

Not tested yet.


src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+                Err(()) => return rustls_result::PrivateKeyParseError,
+            }
+        };
+        let private_key: PrivateKey = match private_keys.pop() {
+            Some(p) => p,
+            None => return rustls_result::PrivateKeyParseError,
+        };
+        let parsed_chain: Vec<Certificate> = match certs(&mut cert_chain) {
+            Ok(v) => v,
+            Err(()) => return rustls_result::CertificateParseError,
```
tgeoghegan, line 113: Oof, [`Result<Vec<Certificate>, ()>`](https://docs.rs/rustls/0.19.0/rustls/internal/pemfile/fn.certs.html) is a strange return type.

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+        // representing the C code's copy. When it drops, that refcount will go down to 0
+        // and the inner ServerConfig will be dropped.
+        let arc: Arc<ServerConfig> = unsafe { Arc::from_raw(config) };
+        let strong_count = Arc::strong_count(&arc);
+        if strong_count < 1 {
+            eprintln!(
+                "rustls_server_config_free: invariant failed: arc.strong_count was < 1: {}. \
+                You must not free the same server_config multiple times.",
+                strong_count
+            );
```
tgeoghegan, line 156: I forget if we talked about this on a previous PR, but I vote for `panic` here. If we can detect a double free, the best thing to do is to crash to make it as easy as possible for the programmer who misused `rustls_server_config_free` to debug the problem.

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+            Ok(()) => rustls_result::Ok,
+            Err(e) => map_error(e),
+        }
+    }
+}
+
+/// Turn a *rustls_server_config_builder (mutable) into a *rustls_server_config
+/// (read-only).
+#[no_mangle]
+pub extern "C" fn rustls_server_config_builder_build(
```
tgeoghegan, line 125: Am I ever going to call `rustls_server_config_builder_build` without then passing the result to `rustls_server_session_new`? I wonder if you could provide a function that lets the caller trade `rustls_server_config_builder` for a `rustls_server_session` and then save themselves an extra function call + error handling.

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+        session.wants_write()
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn rustls_server_session_is_handshaking(
+    session: *const rustls_server_session,
+) -> bool {
+    ffi_panic_boundary_bool! {
+        let session: &ServerSession = try_ref_from_ptr!(session, &ServerSession, false);
```
tgeoghegan, line 213: nit/galaxy brain/definitely not worth trying: it feels like having to specify both `ffi_panic_boundary_bool!` and `false` is redundant, and also all these functions have a nearly identical preamble where you use `try_ref_from_ptr!` to get a `ServerSession`. I wonder if you could have a macro `server_ffi_panic_boundary_bool!` that you could use like so:

```
server_ffi_panic_boundary_bool! {
    session.is_handshaking()
}
```

I suspect I just have a bad case of macro madness, though.

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+        let b = Box::new(server);
+        unsafe {
+            *session_out = Box::into_raw(b) as *mut _;
+        }
+
+        return rustls_result::Ok;
+    }
+}
+
+#[no_mangle]
```
tgeoghegan, line 192: The next three functions could stand to have doccomments

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+        };
+        let out_n: &mut size_t = unsafe {
+            match out_n.as_mut() {
+                Some(out_n) => out_n,
+                None => return NullParameter,
+            }
+        };
+        let n_written: usize = match session.write(write_buf) {
+            Ok(n) => n,
+            Err(_) => return rustls_result::Io,
```
tgeoghegan, line 281: Kind of a shame to drop whatever information might be in the error returned from `session.write()`. I know you have thought a ton about error handling so I apologize if you've already made up your mind about this but couldn't `rustls_result` contain an arbitrary error string?

src/server.rs
```diff
@@ -0,0 +1,409 @@
...
+        // zero it out just in case. TODO: use Initializer once it's stabilized.
+        // https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.initializer
+        for c in read_buf.iter_mut() {
+            *c = 0;
+        }
+        let n_read: usize = match session.read(read_buf) {
+            Ok(n) => n,
+            // The CloseNotify TLS alert is benign, but rustls returns it as an Error. See comment on
+            // https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#impl-Read.
+            // Log it and return EOF.
```
tgeoghegan, line 326: It doesn't seem like you log anything here.

src/lib.rs
```diff
@@ -72,28 +73,17 @@ macro_rules! ffi_panic_boundary_unit {
...
-        try_ref_from_ptr!($var, &mut $typ, rustls_result::NullParameter)
-    };
-    ( $var:ident, &mut $typ:ty, $retval:expr ) => {
-        unsafe {
-            match ($var as *mut $typ).as_mut() {
-                Some(c) => c,
-                None => return $retval,
-            }
-        };
-    };
```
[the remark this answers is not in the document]
tgeoghegan: It doesn't compile when I build it locally, so I think you *do* need the `&mut` arms.

tgeoghegan (changes requested):
re: code duplication: I'll take your word for it that exposing the `Session` trait to C is a chore. However you can deduplicate the virtually identical implementations in `server.rs` and `client.rs` by introducing functions such as:

```
fn rustls_session_read(
    session: &mut dyn Session,
    buf: *mut u8,
    count: size_t,
    out_n: *mut size_t,
) -> rustls_result;
```
...and then have `rustls_server_session_read` and `rustls_client_session_read` dispatch to `rustls_session_read` after converting the `ServerSession` or `ClientSession` from C-land. I gave it a try for that particular triple of functions and it at least compiles ([diff.txt](https://github.com/abetterinternet/crustls/files/5876989/diff.txt) attached).

src/client.rs
```diff
@@ -1,14 +1,17 @@
 use libc::{c_char, size_t};
+use std::io::ErrorKind::ConnectionAborted;
 use std::io::{BufReader, Cursor, Read, Write};
```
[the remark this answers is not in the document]
djc: `cargo fmt` is fine with the older version, and I prefer it to your proposal.

src/server.rs
```diff
@@ -0,0 +1,409 @@
+use libc::size_t;
+use std::io::ErrorKind::ConnectionAborted;
+use std::io::{Cursor, Read, Write};
+use std::ptr::null;
+use std::slice;
+use std::sync::Arc;
+
+use rustls::{
+    internal::pemfile::{certs, pkcs8_private_keys, rsa_private_keys},
```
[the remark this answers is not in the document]
djc: There's now a separate `rustls_pemfile` crate that exposes this without all the warnings.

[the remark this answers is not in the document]
tgeoghegan: Great! That crate's versions of [`pkcs_private_keys`](https://docs.rs/rustls-pemfile/0.2.0/rustls_pemfile/fn.pkcs8_private_keys.html) and other functions have much nicer signatures, too.

src/server.rs
tgeoghegan: [`rusttls_pemfile::certs`](https://docs.rs/rustls-pemfile/0.2.0/rustls_pemfile/fn.certs.html) is a little nicer.

src/client.rs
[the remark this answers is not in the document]
tgeoghegan: Fair enough, anything that's good enough for `cargo fmt` is good enough for me.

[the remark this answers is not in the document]
jsha: My personal style is: no nested qualifiers in `use` statements. Or another way of putting it: A `use` statement should end in at most one set of braces, and those braces should not contain any `::`.

I sometimes deviate from this ideal by accident, since I let rust-analyzer fill in my `use` statements for me, and it is not opinionated in the same way. In general `rustfmt` seems to allow a lot of variation here and not try very hard to canonicalize.

src/server.rs
jsha: `panic`'ing across the C-to-rust boundary is undefined behavior, so I want to avoid it. Some discussion at https://users.rust-lang.org/t/feedback-on-c-to-rustls-bindings/54201/6. A related alternative would be to abort instead, but Daniel Stenberg has correctly pointed out that libraries should never abort, but should leave the choice to their caller.

That said, I've been convinced that this branch is not actually meaningful. If strong_count is 0, `Arc`'s internal invariants mean it will already have dropped. Once it's dropped, accessing the related memory (including the strong_count) is, I think, undefined. So I think this is just one of those cases where we are unavoidably dependent on the C caller not to to invoke undefined behavior by freeing something multiple times. I'd be interested in ideas on how we could be robust to multiple frees, though!

src/server.rs
jsha: I think we don't want this. `*rustls_server_config` represents an `Arc<ServerConfig>`. We want to build one once and then use it for multiple calls to `rustls_server_session_new`.

src/server.rs
jsha: This does seem like macro madness! I'll pass for now but it's a cool idea. I bet it would work.

src/server.rs
jsha: I definitely considered some way of returning errors with arbitrary strings. It would be nice in various ways! But there are some annoying downsides. For instance, say we turn `rustls_result` into a `struct { int code; size_t msglen; const char* msg }`. We have to choose one of "crustls owns the memory, and your rustls_result becomes invalid after any subsequent rustls call" (doesn't work for concurrent programs) or "caller takes ownership of the memory and has to remember to free it after getting an error" (easy for caller to leak memory; not leaking memory requires possibly-awkward `goto cleanup`). I'm interested in learning about best C practices here. From what I can tell, it seems like most libraries rely on enum-valued errors.

By the way, I think in the specific `Io` error cases in this library we don't need to worry. We're not doing real I/O, we're writing to a session's in-memory buffer. Internally this is infallible: https://docs.rs/rustls/0.19.0/src/rustls/client/mod.rs.html#787-800.

jsha:
> you can deduplicate the virtually identical implementations in server.rs and client.rs by introducing functions such as:

This is a good point. I started work on this and it was looking rather clean. One problem: your version, with `&mut dyn Session`, introduces dynamic dispatch. Maybe that's not so bad? I haven't measured the performance impact.

Also, I get this error:

```
the type `dyn rustls::Session` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
within `&dyn rustls::Session`, the trait `std::panic::RefUnwindSafe` is not implemented for `dyn rustls::Session`
required because it appears within the type `&dyn rustls::Session`
required because of the requirements on the impl of `std::panic::UnwindSafe` for `&&dyn rustls::Session`
required because it appears within the type `[closure@src/lib.rs:17:42: 19:10 session:&&dyn rustls::Session]`rustcE0277
panic.rs(378, 40): required by this bound in `std::panic::catch_unwind`
lib.rs(17, 15): Error originated from macro here
```

I tried a version with static dispatch:

```
fn rustls_session_read(
    session: &mut impl Session,
    buf: *mut u8,
    count: size_t,
    out_n: *mut size_t,
) -> rustls_result;
```

But I get this error:

```
the type `&mut &mut impl Session` may not be safely transferred across an unwind boundary
within `[closure@src/lib.rs:29:42: 31:10 buf:&*const u8, count:&usize, out_n:&*mut usize, session:&mut &mut impl Session]`, the trait `std::panic::UnwindSafe` is not implemented for `&mut &mut impl Session`
required because it appears within the type `[closure@src/lib.rs:29:42: 31:10 buf:&*const u8, count:&usize, out_n:&*mut usize, session:&mut &mut impl Session]`rustcE0277
session.rs(77, 5): within this `[closure@src/lib.rs:29:42: 31:10 buf:&*const u8, count:&usize, out_n:&*mut usize, session:&mut &mut impl Session]`
panic.rs(378, 40): required by this bound in `std::panic::catch_unwind`
lib.rs(29, 15): Error originated from macro here
```

jsha:
I was able to get the deduplication working like so:

session.rs:
```
pub(crate) fn rustls_session_wants_read(session: &impl Session) -> bool {
    session.wants_read()
}
```

client.rs:

```
use crate::session;

#[no_mangle]
pub extern "C" fn rustls_client_session_wants_read(session: *const rustls_client_session) -> bool {
    ffi_panic_boundary_bool! {
        let session: &ClientSession = try_ref_from_ptr!(session, &ClientSession, false);
        session::rustls_session_wants_read(session)
    }
}
```

My understanding from the error messages above is that I can't pass a reference type across the catch_unwind boundary (`ffi_panic_boundary_*`) unless it's explicitly given the marker trait UnwindSafe. I think in practice that means the `ffi_panic_boundary_` macro needs to go on the outer func, so there's not that much useful sharing.

I tried pushing the try_ref_from_ptr into the inner, shared func, but that doesn't work because there we need to cast to a specific struct type - `ClientSession` or `ServerSession`.

One other option to reduce code duplication would be to define one giant `session` macro, then evaluate it with an arg of `client` and again with an arg of `server`.

djc:
In my understanding, if we know in advance that only `ServerSession` and `ClientSession` are used as implementers of `dyn Session`, we can probably safely assert that `dyn Session` is `RefUnwindSafe` (since there should be no interior mutability involved). I wouldn't worry too much about the performance impact of dynamic dispatch, it seems likely to be lost in the noise of crypto operations and networking stuff. That said, a macro that deduplicates bindings to all `Session` methods for both types of sessions also seems like a decent solution.

src/server.rs
tgeoghegan: That's what I'd forgotten, that `curl` has rules about libraries crashing intentionally. I have no objection to complying with that requirement, and your point about UB makes it moot anyway.

tgeoghegan:
The difficulty is that the `Session` and `RefUnwindSafe` traits are both defined outside `crustls`  so you're not allowed to make the one require the other. Perhaps you could define a further trait `CrustlsSession` that is `Session + RefUnwindSafe`?

IMO even if you have to leave the `ffi_panic_boundary_*` usage in the outer function, it's worth it for unifying the inner logic. However as the library maintainer I think it's your prerogative to choose which one you like best.

tgeoghegan (approved):

jsha:
I'm convinced sharing the common `Session` code for `ClientSession` and `ServerSession` is worthwhile. I tried to get this done with a macro, but it turns out Rust's macro system can't concatenate, e.g. `rustls_`, `client`, and `_session_read` to get `rustls_client_session_read`. There's https://doc.rust-lang.org/std/macro.concat_idents.html, but it doesn't work for defining new identifiers. It looks like https://github.com/dtolnay/paste solves this problem, but I at this point I start to think it's too much macro.

I'm going to merge for now and file a separate issue for factoring out the common bits via nice, boring function calls.

PR #32, opened 2021-01-30
Remove check on strong_count < 1.

jsha:
Previously, we tried to check against the possibility that C code calls
rustls_client_config_free twice. We did this by doing `Arc::from_raw`,
then checking if strong_count is < 1.

However, this was undefined behavior: deferencing a dangling pointer.
https://doc.rust-lang.org/reference/behavior-considered-undefined.html

If strong_count went to zero on some previous call, `Arc` would have
dropped its contents.  That means the pointed-to memory is no longer
valid to access, and its contents are undefined. So we might see
strong_count of 1,000,000, or -1,000,000, or any other value; or monkeys
could fly out of our noses.

The C caller can still invoke undefined behavior by calling
`rustls_client_config_free` twice, but the previous change tried to
detect undefined behavior by invoking undefined behavior, which doesn't
work.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -135,18 +135,10 @@ pub extern "C" fn rustls_client_config_builder_load_roots_from_file(
 pub extern "C" fn rustls_client_config_free(config: *const rustls_client_config) {
     ffi_panic_boundary_unit! {
         let config: &ClientConfig = try_ref_from_ptr!(config, &mut ClientConfig, ());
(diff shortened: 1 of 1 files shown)


jsha:
Thanks to @alex for pointing this out in https://github.com/abetterinternet/crustls/pull/1#discussion_r541430723.

tgeoghegan (approved):

PR #33, opened 2021-01-30
Use address sanitizer and UB sanitizer for demo.

jsha:


Makefile
```diff
@@ -10,6 +10,11 @@ CFLAGS := -Werror -Wall -Wextra -Wpedantic -g
 PROFILE := debug
 DESTDIR=/usr/local
 
+ifeq ($(CC), clang)
+	CFLAGS += -fsanitize=address -fPIE -fsanitize=undefined
```
tgeoghegan, line 14: Is position independent executable required for the sanitizers? Do we not want PIE when building with GCC?

tgeoghegan (approved):

Makefile
jsha: When testing this, I ran into errors that I thought (based on Googling) were due to lacking `-fPIE`. However, I think that was incorrect. Removing `-fPIE` and re-running, this works.

PR #34, opened 2021-01-30
Add clang and GCC to test matrix.

jsha:


.github/workflows/test.yaml
```diff
@@ -7,10 +7,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest]
+        cc: [clang, gcc]
```
tgeoghegan, line 10: Are you sure there will be a `gcc` installed on `macOS-latest`, or a `clang` installed on `ubuntu-latest`?

tgeoghegan (commented):
This LGTM provided the requested compilers are available on all the requested operating systems. I can't tell if this works because I think your `.github/workflows/test.yaml` is broken on `main`: https://github.com/abetterinternet/crustls/actions

.github/workflows/test.yaml
jsha: Yep!

PR #35, opened 2021-02-01
fix GitHub action YAML

tgeoghegan:
The `jobs` key was defined twice, causing test jobs to fail
(https://github.com/abetterinternet/crustls/actions/runs/522584929).


jsha:
Thanks!

PR #38, opened 2021-02-09
Add rustls_server_session_sni_hostname_get() to the API

icing:
Give C access to the SNI hostname provided by the TLS client. Follows naming and parameter patterns from other calls.

Feel free to discard this PR if you already worked on your side on this, of course.

### DIFF
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -393,7 +393,7 @@ rustls_result rustls_server_session_write_tls(rustls_server_session *session,
                                               size_t *out_n);
 
 /**
- * Write up to `count` characters of the SNI hostname into `buf`. If the
+ * Write up to `len` characters of the SNI hostname into `buf`. If the
  * handshake has not been performed yet or if the client does not support SNI,
  * the returned length in out_n will be 0.
  * https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -392,4 +392,13 @@ rustls_result rustls_server_session_write_tls(rustls_server_session *session,
                                               size_t count,
                                               size_t *out_n);
 
+/**
+ * Write up to `count` characters of the SNI hostname into `buf`. If the
+ * handshake has not been performed yet or if the client does not support SNI,
+ * the returned length in out_n will be 0.
+ * https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
+ */
+rustls_result rustls_server_session_sni_hostname_get(rustls_server_session *session,
+                                                     char *buf, size_t len, size_t *out_n);
+
 #endif /* CRUSTLS_H */
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,4 +1,5 @@
-use libc::size_t;
+use libc::{c_char, size_t};
+use std::cmp::min;
 use std::io::ErrorKind::ConnectionAborted;
 use std::io::{Cursor, Read, Write};
 use std::ptr::null;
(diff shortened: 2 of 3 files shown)


src/crustls.h
```diff
@@ -392,4 +392,13 @@ rustls_result rustls_server_session_write_tls(rustls_server_session *session,
                                               size_t count,
                                               size_t *out_n);
 
+/**
+ * Write up to `len` characters of the SNI hostname into `buf`. If the
+ * handshake has not been performed yet or if the client does not support SNI,
+ * the returned length in out_n will be 0.
+ * https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
+ */
+rustls_result rustls_server_session_sni_hostname_get(rustls_server_session *session,
```
tgeoghegan, line 401: nit: `rustls_server_session_get_sni_hostname` seems like a more natural way to translate the `rustls` method.

src/server.rs
```diff
@@ -402,3 +403,41 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+/// https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
+#[no_mangle]
+pub extern "C" fn rustls_server_session_sni_hostname_get(
+    session: *mut rustls_server_session,
+    buf: *mut c_char,
+    len: size_t,
+    out_n: *mut size_t,
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let session: &mut ServerSession = try_ref_from_ptr!(session, &mut ServerSession);
```
tgeoghegan, line 419: [`ServerSession::get_sni_hostname`](https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname) takes `&self`. Can `session` be an immutable reference?

src/server.rs
```diff
@@ -402,3 +403,41 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+                return NullParameter;
+            }
+            slice::from_raw_parts_mut(buf as *mut u8, len as usize)
+        };
+        let hostname_str = match session.get_sni_hostname() {
+            Some(hostname_str) => hostname_str,
+            None => "",
+        };
+        let len: usize = min(write_buf.len() - 1, hostname_str.len());
+        write_buf[..len].copy_from_slice(&hostname_str.as_bytes()[..len]);
```
tgeoghegan, line 436: Would this truncate the `hostname` if the `buf` provided by the C side is too small? Is this the contract between `curl` and TLS backends? Silently truncating the hostname seems risky to me. Can we instead return an error indicating to the caller explicitly that they need to provide a bigger buffer?

jsha:
The PR looks in order to me. The only question in my mind is one of API design. In Rust `&str` means "a const reference to memory owned by someone else." We could also do an API like this:

```
// Store a pointer to the SNI hostname in `buf`, and the length of that hostname in `out_n`.
// The string is stored in UTF-8 with no terminating NUL byte.
// The caller must not hold a reference to `buf` after `session` has been freed.
rustls_result rustls_server_session_sni_hostname_get(rustls_server_session *session,
                                                     const char **buf, size_t *out_n);
```

As a C developer and the first client of this API, what do you think? It saves the caller having to set up a buf to write into, but it makes the caller think more about lifetimes.

Also, I'd prefer to call it rustls_server_session_get_sni_hostname, moving the `get` towards the front, since that matches the underlying method.

tgeoghegan (commented):
Thanks for the contribution! I'm leaving this as a "comment" review since this is primarily @jsha's project so I want to let him approve or not.

icing:
That would work for me as well. Unnecessary copying when the lifetime is safe for "a while", so the C part can do a  strndup() on it.
As to the method name, please change it as you see fit and in the scheme of things. I thought it was like something_something_verb(), but having it as the rustls method has worth in itself.

src/server.rs
icing: This would truncate, but the new way @jsha proposed would get rid of that buffer anyway.

icing:
I'll make a new PR.

PR #42, opened 2021-02-10
Rework of new method rustls_server_session_get_sni_hostname()

icing:
Rework of the method to access the SNI hostname of the session after
review by others.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -62,6 +62,22 @@ pub extern "C" fn rustls_server_config_builder_new() -> *mut rustls_server_confi
     }
 }
 
+/// With `ignore` != 0, the server will ignore the client ordering of cipher
+/// suites, aka preference, during handshake and respect its own ordering
+/// as configured.
+/// https://docs.rs/rustls/0.19.0/rustls/struct.ServerConfig.html#fields
+#[no_mangle]
+pub extern "C" fn rustls_server_config_builder_set_ignore_client_order(
+    builder: *mut rustls_server_config_builder,
+    ignore: i8,
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let config: &mut ServerConfig = try_ref_from_ptr!(builder, &mut ServerConfig);
+        config.ignore_client_order = ignore != 0;
+        rustls_result::Ok
+    }
+}
+
 /// Sets a single certificate chain and matching private key.
 /// This certificate and key is used for all subsequent connections,
 /// irrespective of things like SNI hostname.
--- a/src/server.rs
+++ b/src/server.rs
@@ -425,7 +425,7 @@ pub extern "C" fn rustls_server_session_write_tls(
 /// Returns RUSTLS_RESULT_INSUFFICIENT_SIZE if the SNI hostname is longer than `count`.
 /// https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
 #[no_mangle]
-pub extern "C" fn rustls_server_session_sni_hostname_get(
+pub extern "C" fn rustls_server_session_get_sni_hostname(
     session: *const rustls_server_session,
     buf: *mut u8,
     count: size_t,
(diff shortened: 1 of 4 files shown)


icing:
Has the cbindgen updated its code generation? I just let the Makefile run its course...

src/server.rs
```diff
@@ -402,3 +418,47 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+/// https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#method.get_sni_hostname
+#[no_mangle]
+pub extern "C" fn rustls_server_session_get_sni_hostname(
+    session: *const rustls_server_session,
+    buf: *mut u8,
+    count: size_t,
+    out_n: *mut size_t,
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let session: &ServerSession = try_ref_from_ptr!(session, &mut ServerSession, NullParameter);
```
tgeoghegan, line 435:
```suggestion
        let session: &ServerSession = try_ref_from_ptr!(session, &ServerSession, NullParameter);
```

src/server.rs
```diff
@@ -62,6 +62,22 @@ pub extern "C" fn rustls_server_config_builder_new() -> *mut rustls_server_confi
...
 }
 
+/// With `ignore` != 0, the server will ignore the client ordering of cipher
+/// suites, aka preference, during handshake and respect its own ordering
+/// as configured.
+/// https://docs.rs/rustls/0.19.0/rustls/struct.ServerConfig.html#fields
+#[no_mangle]
+pub extern "C" fn rustls_server_config_builder_set_ignore_client_order(
+    builder: *mut rustls_server_config_builder,
+    ignore: i8,
```
tgeoghegan, line 72: The field on `ServerConfig` is `bool`, so why not plumb a `bool` all the way to the C interface? We already return `bool` in a few places so I don't think there's a concern about `stdbool.h` not being available.

src/server.rs
```diff
@@ -402,3 +418,47 @@ pub extern "C" fn rustls_server_session_write_tls(
         rustls_result::Ok
     }
 }
+
+/// Copy the SNI hostname to `buf` wich can hold up  to `count` bytes,
```
tgeoghegan, line 422:
```suggestion
/// Copy the SNI hostname to `buf` which can hold up  to `count` bytes,
```

src/server.rs
```diff
@@ -402,3 +418,47 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+/// and the length of that hostname in `out_n`. The string is stored in UTF-8
+/// with no terminating NUL byte.
+/// Returns RUSTLS_RESULT_INSUFFICIENT_SIZE if the SNI hostname is longer than `count`.
```
jsha, line 425: Docs should also say "Returns Ok with *out_n == 0 if there is no SNI hostname available on this session because it hasn't been processed yet, or because the client did not send SNI."

src/server.rs
```diff
@@ -402,3 +418,47 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+            },
+        };
+        let len: usize = sni_hostname.len();
+        if len <= write_buf.len() {
+            write_buf[..len].copy_from_slice(&sni_hostname.as_bytes());
+            unsafe {
+                *out_n = len;
+            }
+            return rustls_result::Ok
+
```
jsha, line 460:
```suggestion
```

Stray newline.

src/server.rs
```diff
@@ -402,3 +418,47 @@ pub extern "C" fn rustls_server_session_write_tls(
...
+            slice::from_raw_parts_mut(buf as *mut u8, count as usize)
+        };
+        let sni_hostname = match session.get_sni_hostname() {
+            Some(sni_hostname) => sni_hostname,
+            None => {
+                return rustls_result::Ok
+            },
+        };
+        let len: usize = sni_hostname.len();
+        if len <= write_buf.len() {
```
jsha, line 454: Style-wise, I would invert the condition: If the len is too big, error out and return early. Otherwise, go on with the copying and return Ok. I generally like to see Ok be the fallthrough case at the bottom of the function, mostly because that's what's common.

jsha (changes requested):
Looking good! Thanks for the PR.

jsha:
Looks like cbindgen did change its style. I merged #43 so you can do this change without spurious diffs to the header file.

jsha (approved):
Thanks!

PR #44, opened 2021-02-10
Support Winsock in crustls-demo

gvanem:
As per the issue https://github.com/abetterinternet/crustls/issues/15, here is my feeble attempt at this demo.
Tested with MSVC / clang-cl only


src/main.c
```diff
@@ -30,23 +61,45 @@ print_error(char *prefix, rustls_result result)
...
+const char *ws_strerror (int err)
+{
+  static char ws_err[50];
+
+  if (err >= WSABASEERR) {
+    snprintf(ws_err, sizeof(ws_err), "Winsock err: %d", err);
+    return ws_err;
+  }
+  /* Assume a CRT error */
+  return (strerror)(err);
```
jsha, line 74: Why is strerror in parentheses here?

src/main.c
```diff
@@ -30,23 +61,45 @@ print_error(char *prefix, rustls_result result)
...
+}
+#endif
+
 /*
  * Write n bytes from buf to the provided fd, retrying short writes until
  * we finish or hit an error. Assumes fd is blocking and therefore doesn't
  * handle EAGAIN. Returns 0 for success or 1 for error.
+ *
+ * For Windsock we cannot use a socket-fd in write().
```
jsha, line 83:
```suggestion
 * For Winsock we cannot use a socket-fd in write().
```

src/main.c
```diff
@@ -30,23 +61,45 @@ print_error(char *prefix, rustls_result result)
...
+ * Call send() if fd > STDOUT_FILENO.
```
jsha, line 84: I think `send` and `recv` are supported on both Linux and macOS. I'd be happier splitting this function into two: `write_all` (for FDs) and `send_all` (for sockets).

jsha (changes requested):
Awesome, thanks for writing this!

src/main.c
gvanem: To get the original CRT `strerror()` back after I changed it.

src/main.c
gvanem: That's what I was thinking too. But that would be too many diffs.

PR #46, opened 2021-02-12
Make rustls_server_config_builder_set_single_cert_pem() work with RSA…

icing:
… PRIVATE KEY pem files.

  The alternative recognition of rsa_private_keys in the parsed item list was not
  working and alwas resulted in a rustls_result::PrivateKeyParseError.


jsha (approved):
Thanks!

PR #50, opened 2021-02-16
C callback implemented for client hello inspection

icing:
Adding capabilites to setup a list of certificates or to install a callback for certificate selection during TLS handshake processing in the server.

Users of the API can create `rustls_certified_key` from the PEMs of certificate (chain) and key and install a list of `rustls_certified_key` in a server configuration. `rustls_certified_key` can be freed as soon as the application no longer uses them as automatic reference counting manages the lifetimes when shared among sessions. When more than one certificate is installed in a server config, the first one that is supported by the client's signature verification capabilities is used.

A callback to inspect the `ClientHello` information can be installed as an alternative. The callback is expected to return the `rustls_certified_key` to be used for  the TLS session or return `NULL` in which case the handshake will fail.

In combination, these two setups allow the handling of serving several DNS domains on the same port and configuring TLS correctly based on the client hello information (most often the SNI hostname).


jsha (commented):
Generally looks good! We will probably want to change this API once a newer solution comes out upstream, but this looks like the right direction for now. You'll need to add output args to sni_callback to return a CertifiedKey, and you'll need to define a C analog of CertifiedKey for the callback to return.

In https://github.com/abetterinternet/crustls/compare/server-cert-verifier I picked "userdata" for the name of the data parameter, which I think is clearer. However, if C style is `_callback_data` I'm happy to go with that.

The fn arg should be Option<...>, to represent nullability. Then we can check the param and return NullParameter if needed.

Thoughts on `*mut c_void` vs `*const c_void` for the data parameter? I went with `*const c_void` on my ServerCertVerifier branch out of conservatism, but could be convinced either way. Maybe some implementations will want to maintain a cache, and thus need mutability?

We need to define the lifetime of the parameters. It should be something like "the data parameter must be valid to dereference until after (a) the config has been freed, and (b) all sessions created with that config have been freed."

icing:
Naming is generally not an issue with me. `userdata` is very common and will work nicely.

Since this part of rustls seems likely to be changed, I would not invest too much into the return of the callback. In mod_tls, I will use the dual session approach until rustls comes with another API. The nice thing about that is that I can select the complete config for the second session, including all other parameters and there is less need to pass some `CertifiedKey`-C-alike back and forth.

For the `const void *`, commonly just a `void*` is used. The keeper of the pointer has to treat it as opaque and the const'ness of where it points to has no relevance. In fact, it often points to something mutable for the callee.

jsha:
Good point about the "dual session approach." Do make sure to document that as part of this work, since it's not something someone would guess automatically. By the way, do you have a public branch where you're working on mod_tls? I would be curious to see the code.

Thanks for the advice on `void*`. I'll update my branch to match.

icing:
Just added it as one of my repositories: https://github.com/icing/mod_tls

icing:
Inspired by your work on the ClientVerifier I adjusted the ClientHello callback in this PR. Looks much better now, I think. Thanks a lot for the example!

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+/// Any error returned by the callback will abort the TLS handshake and give
+/// an error in the called rustls function (most likely
+/// `rustls_server_session_write_tls`).
+#[no_mangle]
+pub extern "C" fn rustls_server_config_builder_set_hello_callback(
+    builder: *mut rustls_server_config_builder,
+    callback: Option<extern "C" fn(
+        userdata: ClientHelloUserData,
+        hello: *const rustls_client_hello
+    ) -> rustls_result>,
```
jsha, line 564: Let's use `Option<ClientHelloCallback>` (or `Option<rustls_client_hello_callback>`) here.

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+        };
+        match &self.previous_resolver {
+            Some(r) => r.resolve(client_hello),
+            None => None
+        }
+    }
+}
+
+unsafe impl marker::Sync for ClientHelloResolver {}
+unsafe impl marker::Send for ClientHelloResolver {}
```
jsha, line 547: Send and Sync are in [`std::prelude`](https://doc.rust-lang.org/std/prelude/index.html), so you don't need to preface them with `marker::` (and it's somewhat unidiomatic to do so).

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+        userdata: ClientHelloUserData,
+        previous_resolver: Option<Arc<dyn rustls::ResolvesServerCert>>,
+    ) -> ClientHelloResolver {
+        ClientHelloResolver { callback, userdata, previous_resolver }
+    }
+}
+
+impl rustls::ResolvesServerCert for ClientHelloResolver {
+    fn resolve(&self, client_hello: ClientHello<'_>) -> Option<CertifiedKey> {
+        let sni_name = {
```
jsha, line 520:
```suggestion
        let sni_name: &str = {
```

I prefer lots of type annotations, particularly when relying on type inference like `c.into()`.

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+            sni_name: sni_name.as_ptr() as *const c_char,
+            sni_name_len: sni_name.len() as size_t,
+            signature_schemes: sigschemes.as_ptr(),
+            signature_schemes_len: sigschemes.len(),
+        };
+        let cb = self.callback;
+        let result: rustls_result = unsafe { cb(self.userdata, &hello) };
+        match result {
+            rustls_result::Ok => rustls_result::Ok,
+            r => panic!("client hello callback returned: {}", r)
```
jsha, line 537: We'd like to avoid panicking as much as possible in this library. It will get transformed by one of the ffi_panic_boundaries anyhow. In this situation let's do an early-return.

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+    sni_name_len: size_t,
+    signature_schemes: *const c_ushort,
+    signature_schemes_len: size_t,
+}
+
+/// Any context information the callback will receive when invoked.
+type ClientHelloUserData = *mut c_void;
+
+/// Called when the TLS ClientHello message was received and successfully decoded.
+type ClientHelloCallback = unsafe extern "C" fn(
```
jsha, line 495: These typedefs wind up getting exported in the .h, so let's use C-style naming for them: `rustls_client_hello_user_data` and `rustls_client_hello_callback` (this is something I missed in my own branch; I'll update it to follow my own advice :-D)

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+    userdata: ClientHelloUserData,
+    hello: *const rustls_client_hello,
+) -> rustls_result;
+
+struct ClientHelloResolver {
+    /// Implementation of rustls::ResolvesServerCert that passes values
+    /// from the supplied ClientHello to the callback function.
+    pub callback: ClientHelloCallback,
+    pub userdata: ClientHelloUserData,
+    pub previous_resolver: Option<Arc<dyn rustls::ResolvesServerCert>>,
```
jsha, line 505: If we get rid of the `replace` semantics, this field can go away.

src/server.rs
```diff
@@ -469,4 +473,112 @@ pub extern "C" fn rustls_server_session_get_sni_hostname(
...
+}
+
+/// Any context information the callback will receive when invoked.
+type ClientHelloUserData = *mut c_void;
+
+/// Called when the TLS ClientHello message was received and successfully decoded.
+type ClientHelloCallback = unsafe extern "C" fn(
+    userdata: ClientHelloUserData,
+    hello: *const rustls_client_hello,
+) -> rustls_result;
```
jsha, line 498: I think the current version of the callback should return void. Reasoning: The callback can't return a certificate, so to be useful it has to be something like "inform the C context (via userdata) about the contents of the Client Hello." That should be infallible in most cases.

src/cipher.rs
```diff
@@ -0,0 +1,67 @@
...
+    for scheme in ALL_SIGNATURE_SCHEMES {
+        if scheme.get_u16() == n {
+            return format!("{:?}", scheme)
+        }
+    };
+    String::from("Unknown")
+}
+
+pub(crate) fn map_signature_schemes(schemes: &[rustls::SignatureScheme]) -> Vec<u16> {
+    let mut mapped_schemes :Vec<u16> = Vec::new();
```
jsha, line 34:
```suggestion
    let mut mapped_schemes: Vec<u16> = Vec::new();
```

Note: rustfmt should catch this (and I should add rustfmt checking to CI). If you're using an IDE, I recommend turning on automatic rustfmt.

src/cipher.rs
```diff
@@ -0,0 +1,67 @@
...
+    for s in schemes {
+        mapped_schemes.push(s.get_u16());
+    }
+    mapped_schemes
+}
+
+#[no_mangle]
+pub extern "C" fn rustls_signature_scheme_name(
+    scheme: c_ushort,
```
jsha, line 43: I'd like to use a type alias, like `type rustls_signature_scheme = c_ushort;`.

jsha (changes requested):

jsha:
By the way, reminder to add doccomments for most of the new items, particularly those that are exported. In particular I think it's important to comment the callback type explaining how you can use it even though it doesn't return a certificate. And the doccomments should include descriptions of how long the userdata must stay valid, and that the callback should not capture copies of any of the pointers it receives.

> Currently, each connection has to start with a new builder, as the callback needs to be registered before the config is created. This cannot be done before the connection exists, as the connection needs to be passed to the callback. It would be nicer, if the application could attach the callback data on creation of the session from a config.

I think this is okay. It seems likely that this will be a temporary API until rustls offers more explicit support for this pattern. In fact, perhaps we should explicitly call out in the API that this is likely to change with a future rustls release?

Also, if we are going to land this, I would like the callback data to include ALPN info, if it's not too complicate to add.

One open question: Should the callback data struct be passed as a pointer or as a value? I have it as a pointer in my server-cert-verifier branch, but it seems like it could just as well be a value, since the struct is not opaque.

src/server.rs
icing: Ack.

src/server.rs
icing: Then we can make the callback `void`, if it cannot fail. Or did I misread?

icing: Ah you suggested it below. Ack.

src/server.rs
icing: I would, but then the generated `crustls.h` no longer contains `rustls_client_hello_callback` and realted things, but only an opaque

```
typedef struct Option_rustls_client_hello_callback Option_rustls_client_hello_callback;
```

Can one make `cbindgen` understand?

src/base.rs
```diff
@@ -0,0 +1,108 @@
...
+/// needed for longer.
+///
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct rustls_vec_ushort {
+    data: *const c_ushort,
+    len: size_t,
+}
+
+impl From<&Vec<u16>> for rustls_vec_ushort {
```
jsha, line 101: Similar to the comment above about `&String` and `&str`, this should be `From<&[u16]>`, since `Vec<u16>` Deref's to `[u16]` https://doc.rust-lang.org/std/vec/struct.Vec.html#deref-methods.

src/cipher.rs
```diff
@@ -0,0 +1,83 @@
+use std::{cmp::min, slice};
+
+use crate::{ffi_panic_boundary_generic, ffi_panic_boundary_unit};
+use libc::{c_char, size_t};
+use std::os::raw::c_ushort;
+
+/// All SignatureScheme currently defines in rustls.
```
jsha, line 7:
```suggestion
/// All SignatureScheme currently defined in rustls.
```

jsha (changes requested):
> typedef struct Option_rustls_client_hello_callback Option_rustls_client_hello_callback;
> Can one make cbindgen understand?

I struggled with this a bit yesterday. Here's the solution I wound up at: https://github.com/abetterinternet/crustls/pull/51/files#diff-7f93c4e263c4e9ec748f804c7fd04a3b2fde86ffd741fb5516d67e1097bae4c1R121-R134. It involves duplicating the definition, but just once:

```rust
#[allow(non_camel_case_types)]
type rustls_verify_server_cert_callback = Option<
    unsafe extern "C" fn(
        userdata: rustls_verify_server_cert_user_data,
        params: *const rustls_verify_server_cert_params,
    ) -> rustls_result,

// This is the same as a rustls_verify_server_cert_callback after unwrapping
// the Option (which is equivalent to checking for null).
type NonNullVerifyCallback = unsafe extern "C" fn(
    userdata: rustls_verify_server_cert_user_data,
    params: *const rustls_verify_server_cert_params,
) -> rustls_result;
```

I also found this interesting thread: https://github.com/eqrion/cbindgen/issues/326


> pub struct rustls_string {

I like this idea, but a couple of tweaks: Let's have separate types for rustls_string vs rustls_bytes, because strings are guaranteed to be UTF-8, and also generally have different semantics than byte arrays. Also, rather than exposing these types directly to the user, I'd like to define type aliases where appropriate. For instance in #51 I have an equivalent type `rustls_certificate`. I would still like to have a type named `rustls_certificate` because the type name indicates useful semantics. But it can become a type alias for `rustls_bytes`.

src/base.rs
icing: This does get complains:

```
the trait `From<&Vec<u16>>` is not implemented for `rustls_vec_ushort`
```

icing:
Made the changes apart from the Vec deref that I cannot get to work.

Re exposing `rustls_bytes` as `rustls_certificate`. As long as it is not opaque, I see no benefit from it. To the contrary: an API user might write functions to copy `rustls_bytes` into its own, allocated data structures or convert `rustls_string` to a C-string. 

Having individual types for each use case would cause the user either duplicate this code for each type, or merely cast it everywhere. After all, `rustls` itself exposes the ALPN values as `&[&[u8]]`.

jsha (changes requested):
Alright, I think we're getting close!

I realized I made a bit of a mistake in naming. What I called `rustls_vec_slice_bytes` should really be `rusts_slice_slice_bytes`. Vec's are growable and owned, and are not `#[repr(C)]`. So we can never really pass a Vec to C except as opaque. What we're really passing to C is a slice of slices. Same applies to `rustls_vec_str`. It should be `rustls_slice_str` (and we should clarify in the docs that it is a slice of multiple strs, not a sub-slice of a str).

`rustls_vec_slice_bytes_keeper` is not `#[repr(C)]` and is only used within Rust code, so we should give it a CamelCase name, e.g. `VecSliceBytes`. Also, it doesn't need to have `view` as a field: instead it can have `fn view(&'a self) -> rustls_slice_slice_bytes<'a>` (I think you can probably elide the lifetime here though).

For the signature schemes: Thoughts on giving them enums or const names?

BTW you've got a session.rs checked in that looks unrelated.

icing:
The easy things fist:

 * the `session.rs` came with our PR yesterday, but I'l happily remove it again. 
 * the renaming: whatever you think fits best. I have no large stake in this.
 * Keeper with `fn` for `view`: does that not introduce the problem again that something leads a separate life with Rust-does-not-help pointers to the other struct? I fail to see the benefit in making this a function. Please explain.
 * Related to this: Rust complains about the "unused" `data` member in the keeper. I had to add the `allow(dead_code)` to silence it. Is there another way to tell Rust that it does not know everything and move on to acceptance?

icing:
I found a way to create `rustls_server_session`s with the flexibility I need. There is now a `rustls_server_config_builder_from_config()` that takes an existing config and clones it into a new builder. I use this in `mod_tls` now to set the correct ALPN response value. Very convenient since all other configurations done for the virtual host at server startup remain in effect.

jsha:
> Keeper with fn for view: does that not introduce the problem again that something leads a separate life with Rust-does-not-help pointers to the other struct? I fail to see the benefit in making this a function. Please explain.

The problem before was that `rustls_vec_str` (for example) did not take ownership of the `Vec` and did not convey lifetime information with a `PhantomData`, even though `rustls_str` did. Here's a variant where I define `RustlsVecStr` to hold ownership of the `Vec`, and offer a `view` function:

```rust
#[repr(C)]
pub struct rustls_vec_str<'a> {
    data: *const rustls_str<'a>,
    len: size_t,
    phantom: PhantomData<&'a str>,
}

impl<'a> From<&'a [String]> for rustls_vec_str<'a> {
    fn from(input: &'a [String]) -> Self {
        let mut output: Vec<rustls_str> = vec![];
        for b in input {
            let b: &str = b;
            output.push(b.into());
        }
        RustlsVecStr(output).view()
    }
}

struct RustlsVecStr<'a>(Vec<rustls_str<'a>>);

impl<'a> RustlsVecStr<'a> {
    fn view(&'a self) -> rustls_vec_str<'a> {
        rustls_vec_str {
            data: self.0.as_ptr(),
            len: self.0.len(),
            phantom: PhantomData,
        }
    }
}
```

This errors out as expected with:
```
error[E0515]: cannot return value referencing temporary value
   --> src/base.rs:104:9
    |
104 |         RustlsVecStr(output).view()
    |         --------------------^^^^^^^
    |         |
    |         returns a value referencing data owned by the current function
    |         temporary value created here
```

I'd like to spend some time looking around at what other Rust projects have done for Vecs of Vecs, since it's presumably a fairly common problem. I'll get back to you on this one; no need for changes now.

> I found a way to create rustls_server_sessions with the flexibility I need. There is now a
> rustls_server_config_builder_from_config() that takes an existing config and clones it into a new builder. 

Nice! Would it work to provide a `rustls_server_config_builder_clone()` instead? That would align a little more straightforwardly with Rust methods. The idea being that you'd make the config builder you want once, then clone it, tweak the clone, and build the original and the clone into finalized configs.

icing:
> 
> I'd like to spend some time looking around at what other Rust projects have done for Vecs of Vecs, since it's presumably a fairly common problem. I'll get back to you on this one; no need for changes now.
> 

Sure, nothing against a proven approach to this.

> > I found a way to create rustls_server_sessions with the flexibility I need. There is now a
> > rustls_server_config_builder_from_config() that takes an existing config and clones it into a new builder.
> 
> Nice! Would it work to provide a `rustls_server_config_builder_clone()` instead? That would align a little more straightforwardly with Rust methods. The idea being that you'd make the config builder you want once, then clone it, tweak the clone, and build the original and the clone into finalized configs.

That would mean a builder, which is mutable, hangs around after the server startup and gets cloned on every connection. As long as no one does anything else with it, it will probably be fine, however I feel a bit uncomfortable with this. 

Currently the config shared is readonly and we use it to create a new mutable builder. That happens per connection, so we do not share any mutable data. I prefer that. In regard to naming, I have nothing against giving it a more rusty name, of course.

src/server.rs
```diff
@@ -93,44 +157,53 @@ pub extern "C" fn rustls_server_config_builder_set_single_cert_pem(
...
-            Ok(v) => v.into_iter().map(Certificate).collect(),
-            Err(_) => return rustls_result::CertificateParseError,
+        config.cert_resolver = Arc::new(ResolvesServerCertFromChoices::new(&[Arc::new(certified_key)]));
+        rustls_result::Ok
+    }
+}
+
+/// Provide the configuration a list of certificates where the session
+/// will select the first one that is compatible with the client's signing
+/// capabilities. Servers that want to support ECDSA and RSA certificates
```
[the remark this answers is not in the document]
icing: Correct. Changed.

src/server.rs
```diff
@@ -83,6 +143,10 @@ pub extern "C" fn rustls_server_config_builder_set_ignore_client_order(
 /// first.
 /// private_key must point to a byte array of length private_key_len containing
 /// a private key in PEM-encoded PKCS#8 or PKCS#1 format.
+///
+/// EXPERIMENTAL: installing a client_hello callback will replace any
+/// configured certified keys and vice versa. Same holds true for the
+/// set_single_cert variant.
 #[no_mangle]
 pub extern "C" fn rustls_server_config_builder_set_single_cert_pem(
```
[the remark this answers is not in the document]
icing: I can remove that function in the PR, if you like.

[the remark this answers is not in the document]
icing: And I removed it, saving us a potential ping-pong across time zones.

jsha (approved):
Thanks for all the work on this! I have a couple more tweaks I would like to make, but I think it's time to merge and build on top.

PR #51, opened 2021-02-18
Add support for custom server certificate verifier

jsha:
C code can pass a callback to be used as a verifier. Note that since we don't yet provide bindings for certificate parsing or path building, the main thing you can do with this callback is turn off certificate verification (or link in another library for parsing).

I've also added a binding for setting the enable_sni field in ClientConfig, so clients that want to disable certificate verification for IP addresses can do so.


src/main.c
```diff
@@ -470,6 +470,11 @@ do_request(const struct rustls_client_config *client_config,
   return ret;
 }
 
+enum rustls_result
+verify(__attribute((unused)) void *userdata, __attribute((unused)) const rustls_verify_server_cert_params *params) {
```
tgeoghegan, line 474: Since this program is the integration test for `crustls`, perhaps this should verify that the `userdata` and `params` that get passed to the C into Rust and back to C don't get mangled somehow.

src/client.rs
```diff
@@ -77,6 +77,173 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+///
+/// The userdata pointer must stay valid until (a) all sessions created with this
+/// config have been freed, and (b) the config itself has been freed.
+/// The callback must not capture any of the pointers in its
+/// rustls_verify_server_cert_params.
+///
+/// The callback must be safe to call on any thread at any time, including
+/// multiple concurrent calls. So, for instance, if the callback mutates
+/// userdata (or other shared state), it must use synchronization primitives
+/// to make such mutatation safe.
```
tgeoghegan, line 207:
```suggestion
/// to make such mutation safe.
```

src/client.rs
```diff
@@ -77,6 +77,173 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+/// (a) always succeed (or fail), or (b) compare the certificates against
+/// static bytes. We plan to export parsing code in the future to make it
+/// possible to implement other strategies.
+///
+/// If you intend to write a verifier that accepts all certificates, be aware
+/// that special measures are required for IP addresses. Rustls currently
+/// (0.19.0) doesn't support building a ClientSession with an IP address
+/// (because it's not a valid DNSNameRef). One workaround is to detect IP
+/// addresses and rewrite them to `example.invalid`, and _also_ to disable
+/// SNI vial rustls_client_config_builder_set_enable_sni (IP addresses don't
```
tgeoghegan, line 221:
```suggestion
/// SNI via rustls_client_config_builder_set_enable_sni (IP addresses don't
```

src/client.rs
```diff
@@ -77,6 +77,173 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+#[no_mangle]
+pub extern "C" fn rustls_client_config_builder_dangerous_set_certificate_verifier(
+    config: *mut rustls_client_config_builder,
+    callback: rustls_verify_server_cert_callback,
+    userdata: rustls_verify_server_cert_user_data,
+) {
+    ffi_panic_boundary_unit! {
+        let callback: NonNullVerifyCallback = match callback {
+            Some(cb) => cb,
+            None => return,
```
tgeoghegan, line 239: If I'm reading this right, this means passing `NULL` for the callback is a no-op. What if the caller wants to clear a previously set callback?

src/client.rs
```diff
@@ -77,6 +77,173 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+// According to the nomicon https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization):
+// > Option<extern "C" fn(c_int) -> c_int> is a correct way to represent a
+// > nullable function pointer using the C ABI (corresponding to the C type int (*)(int)).
+// So we use Option<...> here. This is the type that is passed from C code.
+#[allow(non_camel_case_types)]
+type rustls_verify_server_cert_callback = Option<
+    unsafe extern "C" fn(
+        userdata: rustls_verify_server_cert_user_data,
+        params: *const rustls_verify_server_cert_params,
+    ) -> rustls_result,
```
tgeoghegan, line 126: Can this be `Option<NonNullVerifyCallback>` to avoid the duplication of the inner definition? Also I think `NonNullVerifyCallback` could just be `VerifyCallback`--the fact that it's non-null is implicit because it's not possible to have a value of `NonNullVerifyCallback` that isn't, well, a value. But I see the merit in disambiguating between the Rust callback type and the nullable type exposed to C so whatever you prefer here is fine by me.

src/client.rs
```diff
@@ -77,6 +77,173 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+) -> rustls_result;
+
+// An implementation of rustls::ServerCertVerifier based on a C callback.
+struct Verifier {
+    callback: NonNullVerifyCallback,
+    userdata: rustls_verify_server_cert_user_data,
+}
+
+unsafe impl Send for Verifier {}
+unsafe impl Sync for Verifier {}
```
tgeoghegan, line 143: What does it mean to provide an `unsafe impl`? Does this mean this type is _not_ actually `Send` or `Sync` but we're telling the compiler to pretend that it is?

tgeoghegan (changes requested):

src/client.rs
jsha: They would have to define their own do-nothing callback and pass it in. Note that rustls itself doesn't have a concept of "no ServerCertVerifier," just "the default ServerCertVerifier."

Also keep in mind this is a builder object. You can't use a config till you've finished building it, so it would be somewhat pointless to set and then clear a certificate verifier callback.

tgeoghegan: Oh, right, builder pattern. Yeah, I agree there's no point to supporting clearing the callback.

src/client.rs
tgeoghegan: That's a shame about `cbindgen`. These definitions are part of the stable API and aren't likely to change so the duplication isn't so bad.

src/client.rs
```diff
@@ -74,6 +77,165 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+    _private: [u8; 0],
+}
+
+/// Input to a custom certificate verifier callback. See
+/// rustls_client_config_builder_dangerous_set_certificate_verifier().
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct rustls_verify_server_cert_params<'a> {
+    end_entity: rustls_slice_bytes<'a>,
+    intermediates: rustls_slice_slice_bytes<'a>,
```
tgeoghegan, line 93: I'm assuming `end_entity` and `intermediates` are certificates, but that's not obvious since the types are byte slices. This would be more clear as `end_entity_cert` and `intermediate_certs`.

src/client.rs
```diff
@@ -74,6 +77,165 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+        roots: &RootCertStore,
+        presented_certs: &[Certificate],
+        dns_name: DNSNameRef<'_>,
+        ocsp_response: &[u8],
+    ) -> Result<ServerCertVerified, TLSError> {
+        let cb = self.callback;
+        let dns_name: &str = dns_name.into();
+        let dns_name: rustls_str = match dns_name.try_into() {
+            Ok(r) => r,
+            Err(_) => return Err(TLSError::General("NUL byte in SNI".to_string())),
```
tgeoghegan, line 149: nit: I know this is the only possible failure case from `rustls_str::from` but it'd be nice if that `From` impl's associated error type was `String` and it returned "NUL byte in SNI" on failure so that you wouldn't have to assume that was the cause of the error here in this method.

tgeoghegan (approved):
A couple of nits but LGTM!

src/client.rs
jsha: I updated the `Error` associated type for `TryFrom` to be a type alias `NulByte`, as a way of documenting that's the only error that will come out. I wasn't able to find a way to make that clear at this match size. `Err(())` works but `Err(NulByte)` doesn't.

src/crustls.h
```diff
@@ -223,6 +249,45 @@ struct rustls_client_config_builder *rustls_client_config_builder_new(void);
...
+ * multiple concurrent calls. So, for instance, if the callback mutates
+ * userdata (or other shared state), it must use synchronization primitives
+ * to make such mutation safe.
+ *
+ * The callback receives certificate chain information as raw bytes.
+ * Currently this library offers no functions for C code to parse the
+ * certificates, so it's only possible to implement verifiers that either
+ * (a) always succeed (or fail), or (b) compare the certificates against
+ * static bytes. We plan to export parsing code in the future to make it
+ * possible to implement other strategies.
```
tgeoghegan, line 270: IIUC, the "raw bytes" will point to a `[rustls::Certificate]`. I think that rather than commit to extending `crustls`'s API surface with a rich representation of an X.509 certificate, you should provide this callback the DER encoded certificate chain, because then the client can bring whatever library they want (such as OpenSSL, or maybe a project like `curl` already has a bunch of code for working with certs and won't want to adopt a new interface), parse the certificates and do whatever they want without `crustls` needing to specifically provide an API affordance to enable their use case. Note that I'm not arguing for committing to that API contract now, but rather removing the promise to eventually provide a certificate API in `crustls`.

src/client.rs
```diff
@@ -74,6 +76,164 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+/// Input to a custom certificate verifier callback. See
+/// rustls_client_config_builder_dangerous_set_certificate_verifier().
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct rustls_verify_server_cert_params<'a> {
+    end_entity_cert_der: rustls_slice_bytes<'a>,
+    intermediate_certs_der: &'a rustls_slice_slice_bytes<'a>,
+    roots: *const rustls_root_cert_store,
+    dns_name: rustls_str<'a>,
+    ocsp_response: rustls_slice_bytes<'a>,
```
tgeoghegan, line 95: nit: It's interesting that these struct fields are public from the point of view of a C client, but a Rust client using `crustls` (for some reason) wouldn't be able to access them. Should they all be `pub` if only for clarity of intent? Does that confuse `cbindgen` in any way?

src/client.rs
```diff
@@ -74,6 +76,164 @@ pub extern "C" fn rustls_client_config_builder_build(
...
+    callback: rustls_verify_server_cert_callback,
+    userdata: rustls_verify_server_cert_user_data,
+) {
+    ffi_panic_boundary_unit! {
+        let callback: VerifyCallback = match callback {
+            Some(cb) => cb,
+            None => return,
+        };
+        let config: &mut ClientConfig = try_ref_from_ptr!(config, &mut ClientConfig, ());
+        let verifier: Verifier = Verifier{callback: callback, userdata};
```
tgeoghegan, line 232:
```suggestion
        let verifier = Verifier{callback, userdata};
```
I think the `Verifier` literal is clear enough without the type annotation.

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #54, opened 2021-03-01
Add rustls_str, rustls_slice_bytes, and friends

jsha:
For the basic slice types we need, this introduces `rustls_str`, `rustls_slice_bytes`, and `rustls_slice_u16`. For slices of slices, we have `rustls_slice_str` and `rustls_slice_slice_bytes`. Each of these types have a lifetime parameter that ensures they don't outlive the object they are borrowing from. 

The `str` types have an additional guarantee on top of Rust's guarantee that strings contain UTF-8: They can't contain `\0`. This makes it possible to interpolate them into C strings and compare them against C strings using `strncmp`.

Built off @icing's #50.

### DIFF
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -342,7 +342,7 @@ bool rustls_result_is_cert_error(enum rustls_result result);
  * Retrieve the nth element from the input slice of slices. If the input
  * pointer is NULL, returns 0.
  */
-uintptr_t rustls_slice_slice_bytes_len(const struct rustls_slice_slice_bytes *input);
+size_t rustls_slice_slice_bytes_len(const struct rustls_slice_slice_bytes *input);
 
 /**
  * Retrieve the nth element from the input slice of slices. If the input
@@ -350,20 +350,20 @@ uintptr_t rustls_slice_slice_bytes_len(const struct rustls_slice_slice_bytes *in
  * rustls_slice_slice_bytes, returns rustls_slice_bytes{NULL, 0}.
  */
 struct rustls_slice_bytes rustls_slice_slice_bytes_get(const struct rustls_slice_slice_bytes *input,
-                                                       uintptr_t n);
+                                                       size_t n);
 
 /**
  * Retrieve the nth element from the input slice of slices. If the input
  * pointer is NULL, returns 0.
  */
-uintptr_t rustls_slice_str_len(const struct rustls_slice_str *input);
+size_t rustls_slice_str_len(const struct rustls_slice_str *input);
 
 /**
  * Retrieve the nth element from the input slice of slices. If the input
  * pointer is NULL, or n is greater than the length of the
  * rustls_slice_slice_bytes, returns rustls_str{NULL, 0}.
  */
-struct rustls_str rustls_slice_str_get(const struct rustls_slice_str *input, uintptr_t n);
+struct rustls_str rustls_slice_str_get(const struct rustls_slice_str *input, size_t n);
 
 /**
  * Create a rustls_server_config_builder. Caller owns the memory and must
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -147,7 +147,9 @@ typedef struct rustls_server_session rustls_server_session;
 typedef struct rustls_slice_slice_bytes rustls_slice_slice_bytes;
 
 /**
- * A read-only view of a slice of Rust `&str`.
+ * A read-only view of a slice of multiple Rust `&str`'s (that is, multiple
+ * strings). Like `rustls_str`, this guarantees that each string contains
+ * UTF-8 and no NUL bytes. Strings are not NUL-terminated.
  *
  * This is used to pass data from crustls to callback functions provided
  * by the user of the API. Because Vec and slice are not `#[repr(C)]`, we
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -130,6 +130,76 @@ typedef struct rustls_server_config_builder rustls_server_config_builder;
 typedef struct rustls_server_session rustls_server_session;
 
 /**
+ * A read-only view of a slice of Rust byte slices.
+ *
+ * This is used to pass data from crustls to callback functions provided
+ * by the user of the API. Because Vec and slice are not `#[repr(C)]`, we
+ * provide access via a pointer to an opaque struct and an accessor method
+ * that acts on that struct to get entries of type `rustls_slice_bytes`.
+ * Internally, the pointee is a `&[&[u8]]`.
+ *
+ * The memory exposed is available as specified by the function
+ * using this in its signature. For instance, when this is a parameter to a
+ * callback, the lifetime will usually be the duration of the callback.
+ * Functions that receive one of these must not call its methods beyond the
+ * allowed lifetime.
+ */
+typedef struct rustls_slice_slice_bytes rustls_slice_slice_bytes;
+
+/**
+ * A read-only view of a slice of Rust `&str`.
+ *
+ * This is used to pass data from crustls to callback functions provided
+ * by the user of the API. Because Vec and slice are not `#[repr(C)]`, we
+ * can't provide a straightforward `data` and `len` structure. Instead, we
+ * provide access via a pointer to an opaque struct and accessor methods.
+ * Internally, the pointee is a `&[&str]`.
+ *
+ * The memory exposed is available as specified by the function
+ * using this in its signature. For instance, when this is a parameter to a
+ * callback, the lifetime will usually be the duration of the callback.
+ * Functions that receive one of these must not call its methods beyond the
+ * allowed lifetime.
+ */
+typedef struct rustls_slice_str rustls_slice_str;
+
+/**
+ * A read-only view on a Rust byte slice.
+ *
+ * This is used to pass data from crustls to callback functions provided
+ * by the user of the API.
+ * `len` indicates the number of bytes than can be safely read.
+ *
+ * The memory exposed is available as specified by the function
+ * using this in its signature. For instance, when this is a parameter to a
+ * callback, the lifetime will usually be the duration of the callback.
+ * Functions that receive one of these must not dereference the data pointer
+ * beyond the allowed lifetime.
+ */
+typedef struct rustls_slice_bytes {
+  const uint8_t *data;
+  size_t len;
+} rustls_slice_bytes;
+
+/**
+ * A read-only view on a Rust `&str`. The contents are guaranteed to be valid
+ * UTF-8. As an additional guarantee on top of Rust's normal UTF-8 guarantee,
+ * a `rustls_str` is guaranteed not to contain internal NUL bytes, so it is
+ * safe to interpolate into a C string or compare using strncmp. Keep in mind
+ * that it is not NUL-terminated.
+ *
+ * The memory exposed is available as specified by the function
+ * using this in its signature. For instance, when this is a parameter to a
+ * callback, the lifetime will usually be the duration of the callback.
+ * Functions that receive one of these must not dereference the data pointer
+ * beyond the allowed lifetime.
+ */
+typedef struct rustls_str {
+  const char *data;
+  size_t len;
+} rustls_str;
+
+/**
  * Write the version of the crustls C bindings and rustls itself into the
  * provided buffer, up to a max of `len` bytes. Output is UTF-8 encoded
  * and NUL terminated. Returns the number of bytes written before the NUL.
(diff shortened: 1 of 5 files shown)


src/rslice.rs
```diff
@@ -0,0 +1,184 @@
...
+/// pointers beyond the allowed lifetime.
+#[repr(C)]
+pub struct rustls_slice_bytes<'a> {
+    pub data: *const u8,
+    pub len: size_t,
+    phantom: PhantomData<&'a [u8]>,
+}
+
+impl<'a> From<&'a [u8]> for rustls_slice_bytes<'a> {
+    fn from(s: &[u8]) -> Self {
```
tgeoghegan, line 24: My intuition is that you need `s` (which is a reference) to at least outlive `'a`. Is that lifetime bound implicit here?

src/rslice.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+/// An owned Vec<rustls_slice_bytes>, where the inner slices have lifetime 'a.
+/// If we want to share a view of a `Vec<Vec<_>>` with C, we can't do it
+/// directly because `Vec` is not #[repr(C)]. So we build a new Vec containing
+/// rustls_slice_bytes<'a> as an owned object. We can then expose slices of
+/// that Vec to C.
+pub(crate) struct VecSliceBytes<'a>(Vec<rustls_slice_bytes<'a>>);
```
tgeoghegan, line 38: Could this be a type alias `type VecSliceBytes = Vec<rustls_slice_bytes<'a>>;` and save us the enclosing envelope struct?

src/rslice.rs
```diff
@@ -0,0 +1,184 @@
...
+/// The memory exposed is available as specified by the function
+/// using this in its signature. For instance, when this is a parameter to a
+/// callback, the lifetime will usually be the duration of the callback.
+/// Functions that receive one of these must not retain copies of any of the
+/// pointers beyond the allowed lifetime.
+#[repr(C)]
+pub struct rustls_slice_slice_bytes<'a> {
+    data: *const rustls_slice_bytes<'a>,
+    len: size_t,
+    phantom: PhantomData<&'a [&'a [u8]]>,
```
tgeoghegan, line 68: nit: could be `PhantomData<&'a [rustls_slice_bytes<'a>]>` for consistency, but IIUC it doesn't matter what goes in the angle brackets as long as it includes `'a`.

src/rslice.rs
```diff
@@ -0,0 +1,184 @@
...
+            data: s.as_ptr() as *const c_char,
+            len: s.len() as size_t,
+            phantom: PhantomData,
+        })
+    }
+}
+
+/// An owned Vec<rustls_str>, where the inner slices have lifetime 'a.
+/// See comment on VecSliceBytes for more information.
+pub(crate) struct VecStr<'a>(Vec<rustls_str<'a>>);
```
tgeoghegan, line 116: It seems like `VecStr` and `VecSliceBytes` are manual monomorphizations of `Vec<T>`. But I think `VecStr` (for instance) only gets used in `rustls_slice_str::from<&'a VecStr<'a>>`, so it's never exposed to C. What I wonder is: could you implement `From<&'a Vec<rustls_str<'a>>` on `rustls_slice_str`, move the implementation of `VecStr::new` into `rustls_slice_str::from` and eliminate the need for `VecStr` altogether? I think that's pretty much what you did for `rustls_slice_u16`.

src/rslice.rs
jsha: Yep, this is lifetime elision. https://doc.rust-lang.org/nomicon/lifetime-elision.html

> If there is exactly one input lifetime position (elided or not), that lifetime is assigned to all elided output lifetimes.

src/rslice.rs
jsha: Unfortunately you can't implement traits on type aliases. That's one of the main reasons for using the newtype pattern: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types

jsha:
> Could VecStr and VecSliceBytes be eliminated?

You also mention this in https://github.com/abetterinternet/crustls/pull/54#discussion_r585039954, so I'll respond here: Unfortunately they can't. We would get a "temporary value dropped here while still borrowed." That is, we need to create a new owned `Vec` that holds `rustls_str`'s instead of `&str`s. Given we need an owned object, who owns it? If we create it within a `From` implementation, it will be dropped before `from` returns, so `rustls_slice_str` can't hold a reference to it. Ultimately, whatever function is going to pass this off into C needs to hold the owned `Vec`, to ensure that it lives long enough for the callback to complete.

> Why don't rustls_str etc. appear in an updated crustls.h?

[cbindgen publishes functions, globals, and constants](https://github.com/eqrion/cbindgen/blob/master/docs.md#writing-your-c-api) - plus any types required by those functions, globals and constants. Since these types aren't (yet) used by any published functions, cbindgen does not publish them. You can see rendered .h files [here](https://github.com/abetterinternet/crustls/pull/50/files#diff-d78a232a38a96236f8e8bbe0acb353b20997674d6555e2ab75b16cbc0348d9ecR139-R214) and [here](https://github.com/abetterinternet/crustls/pull/51/files#diff-d78a232a38a96236f8e8bbe0acb353b20997674d6555e2ab75b16cbc0348d9ecR144-R196).

tgeoghegan:
Thanks for the clarifications and links; re-approved!

src/rslice.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/// An owned Vec<rustls_slice_bytes>, where the inner slices have lifetime 'a.
+/// If we want to share a view of a `Vec<Vec<_>>` with C, we can't do it
+/// directly because `Vec` is not #[repr(C)]. So we build a new Vec containing
+/// rustls_slice_bytes<'a> as an owned object. We can then expose slices of
+/// that Vec to C.
+pub(crate) struct VecSliceBytes<'a>(Vec<rustls_slice_bytes<'a>>);
+
+impl<'a> VecSliceBytes<'a> {
```
alex, line 40: I'm a bit confused on how this is intended to be used, can you sketch out an example?

I also don't love needing to allocate for this. It seems preferable to keep the original `Vec<Vec<u8>>` and have C functions to extract a `rustls_slice_bytes` at a given index?

jsha: One example here: https://github.com/abetterinternet/crustls/pull/51/files#diff-7f93c4e263c4e9ec748f804c7fd04a3b2fde86ffd741fb5516d67e1097bae4c1R149-R183

And one here: https://github.com/abetterinternet/crustls/pull/50/files#diff-48d27698cc708c7efe770fd897e4885efaa0a15cae30d54936068603ba03bbd9R696-R711

Good point on accessor functions! I recall that you had recommended that before and I forgot to pursue it. I'll give that a try.

jsha: For the accessor version, I'm thinking of a function signature like this:

```
pub extern "C" fn rustls_slice_slice_bytes_get(
    input: rustls_slice_slice_bytes, n: usize) -> rustls_slice_bytes {
```

But this requires there be some sort of sentinel value of `rustls_slice_bytes` that means "n was too big." I guess that can be `rustls_slice_bytes{data: null(), len: 0}`. Interested in other ideas! I could also do:

```
pub extern "C" fn rustls_slice_slice_bytes_get(
    input: rustls_slice_slice_bytes, n: usize, out: *mut rustls_slice_bytes) -> rustls_result {
```

But that feels pretty awkward to use in a loop.

src/rslice.rs
divergentdave: Moreover, the lifetimes are defined to be equal in the impl line above (the lifetime of the slice that's the parameter to From, and the lifetime parameter of rustls_slice_bytes), and the compiler will check that the function signature here (after applying lifetime elision rules) matches the signature of From::from with the type parameters substituted in.

src/rslice.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub data: *const u16,
+    pub len: size_t,
+    phantom: PhantomData<&'a [u16]>,
+}
+
+impl<'a> From<&'a [u16]> for rustls_slice_u16<'a> {
+    fn from(s: &[u16]) -> Self {
+        rustls_slice_u16 {
+            data: s.as_ptr() as *const u16,
+            len: s.len() as size_t,
```
divergentdave, line 184: FYI, both these casts are redundant, and the from/new functions aren't consistent on whether they do or don't cast.

src/rslice.rs
divergentdave: I agree on the accessor recommendation. They would allow encapsulating some memory layouts, cutting down on the number of types needed on the Rust size, plus calls to the accessors could be inlined when using a cross-language LTO-capable toolchain, making this a cheap abstraction.

The first accessor signature looks good to me, NULL checks seem like idiomatic C IMHO.

jsha:
I've pushed an update with the accessor style. It's not correct yet - see the segfault in https://github.com/abetterinternet/crustls/pull/51 when I tried to incorporate it into that branch. But I figured I'd push up what I've got so far for folks to take a look!

I suspect I'm mixing up types between a thin pointer (`*const T`) and a fat pointer (`&[T]`), because the pointer value is 1, which happens to also be the length of the particular chain we're dealing with.

```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555580426 in crustls::rslice::rustls_slice_slice_bytes_get (input=0x7fffffffa848, n=0) at /home/jsha/crustls/src/rslice.rs:95
95              Some(rsb) => (*rsb).into(),
(gdb) bt
#0  0x0000555555580426 in crustls::rslice::rustls_slice_slice_bytes_get (input=0x7fffffffa848, n=0) at /home/jsha/crustls/src/rslice.rs:95
#1  0x00005555555800d2 in verify (userdata=0x55555573b425, params=0x7fffffffaa48) at src/main.c:487
(gdb) p rsb
$1 = (*mut &[u8]) 0x1
```

divergentdave:
The first thing I noticed is that `rustls_slice_slice_bytes` is now a ZST, since it only contains a PhantomData. I think the issue is that `rustls_slice_slice_bytes_new` is returning a stack pointer to the reference it was passed in. It first takes a reference to the input slice reference, then casts it to one pointer type, then another. Passing an opaque pointer to either a slice (a fat pointer itself) or a Vec does sound like a promising way to go, but we'll need to be careful that the pointee stays put while it is in use.

jsha:
Good catch on the stack pointer, thanks! Pushed a fix. I thought of another approach: we could make `rustls_slice_slice_bytes` a newtype holding `&[&[u8]]`, but make it opaque to C by not making it `pub` (which causes cbindgen to generate an opaque struct declaration).

jsha:
Alright, I switched to a newtype pattern for `_slice_slice_bytes` and `_slice_str`, and I like it a lot better. I don't have to do nearly as much casting of thing to `*const T`. I've also added some tests, and updated #51 to use the latest version of `rslice`. Ready for another round of reviews!

src/crustls.h
```diff
@@ -268,6 +340,33 @@ void rustls_error(enum rustls_result result, char *buf, size_t len, size_t *out_
 
 bool rustls_result_is_cert_error(enum rustls_result result);
 
+/**
+ * Retrieve the nth element from the input slice of slices. If the input
+ * pointer is NULL, returns 0.
+ */
+size_t rustls_slice_slice_bytes_len(const struct rustls_slice_slice_bytes *input);
```
icing, line 347: Comment is copy+pasta. Been there, done that.

src/crustls.h
```diff
@@ -268,6 +340,33 @@ void rustls_error(enum rustls_result result, char *buf, size_t len, size_t *out_
...
+ * rustls_slice_slice_bytes, returns rustls_slice_bytes{NULL, 0}.
+ */
+struct rustls_slice_bytes rustls_slice_slice_bytes_get(const struct rustls_slice_slice_bytes *input,
+                                                       size_t n);
+
+/**
+ * Retrieve the nth element from the input slice of slices. If the input
+ * pointer is NULL, returns 0.
+ */
+size_t rustls_slice_str_len(const struct rustls_slice_str *input);
```
icing, line 361: Description.

icing (approved):
Nice design the never construct the vec of struct thing, but to pass out views on the elements as value!

divergentdave (commented):
FYI, I did a Miri spot-check of the tests here, and they passed. I changed the crate-type in Cargo.toml from staticlib to lib, then ran `cargo +nightly miri test`.

jsha:
> FYI, I did a Miri spot-check of the tests here, and they passed. I changed the crate-type in Cargo.toml from staticlib to lib, then ran cargo +nightly miri test.

Why change from staticlib to lib? When I run with `crate-type = ["staticlib"]` things also work.

divergentdave:
Ah, when I did that at first, it didn't pick up any test targets. Perhaps it was fixed recently, and my nightly channel as installed was too old to have the fix.

src/rslice.rs
```diff
@@ -0,0 +1,305 @@
...
+    n: size_t,
+) -> rustls_slice_bytes<'a> {
+    let input: &rustls_slice_slice_bytes = unsafe {
+        match input.as_ref() {
+            Some(c) => c,
+            None => {
+                return rustls_slice_bytes {
+                    data: null(),
+                    len: 0,
+                    phantom: PhantomData,
```
tgeoghegan, line 91: nit: you could save yourself a little bit of repetition by providing an `std::Default` implementation that constructs a "zero valued" `rustls_slice_bytes`.

src/crustls.h
tgeoghegan: This is odd, because the corresponding doccomment in `rslice.rs` is correct.

tgeoghegan (approved):
Couple nits and a doccomment remark but LGTM!

src/rslice.rs
jsha: This is a good point, but I've found on another project that `std::Default` hides things in a way that can cause trouble. I'd like to favor explicitness and obviousness on this project.

jsha:
on `crustls.h`:
> This is odd, because the corresponding doccomment in rslice.rs is correct.

I hadn't updated `crustls.h`, and also the test phase that was supposed to check for that was broken (I needed `--exit-code` on `git diff`)

icing:
\o/

PR #56, opened 2021-03-08
Icing/enums

icing:
Not sure how we want to merge all this. I think it is time to merge something of our changes into `main` and have a common base where to add future changes to.


jsha (commented):
I appear to have made two mistakes with this PR; my apologies. The first mistake is that I appear to have auto-closed it when I merged the `rslice` branch. Not sure why that happened. But since the `rslice` branch is also deleted, GitHub won't allow me to reopen it. Please go ahead and reopen against `main`.

The other mistake was that I wrote up some feedback yesterday but neglected to send it! Sending this now. I think it's still worthwhile having some conversation about implementation while we also discuss how much configuration to expose, though I also understand if you want to postpone investment on this side of the code until that is settled.

icing:
I'll make a new PR against `main` which incorporates changes in the light of your review. Continuing the conversation here without the possibility to update this seems less effective.

PR #60, opened 2021-03-11
Do some tweaks and cleanups

jsha:
Order imports as: std, third-party, crate.

Use `try_ref_from_ptr!` to turn `out_n`'s into refs. This moves unsafe blocks from later in the function to earlier in the function, closer to where their safety properties are checked. Also, remove pre-emptive zeroing of `out_n`'s. According to the conventions listed in our README, there are no partial successes or partial failures, so we shouldn't write to `out_n` if we early-return.

Remove `strong_count < 1` checks on Arc. Follow-up from #32.

Add some missing comments and fix some formatting.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -1,18 +1,17 @@
+use libc::{c_char, size_t};
+use std::io::Cursor;
+use std::os::raw::c_ushort;
+use std::sync::Arc;
 use std::{cmp::min, slice};
+
+use rustls::sign::CertifiedKey;
+use rustls::{Certificate, PrivateKey};
+use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 
 use crate::error::rustls_result;
 use crate::{
     ffi_panic_boundary, ffi_panic_boundary_generic, ffi_panic_boundary_unit, try_ref_from_ptr,
 };
-use libc::{c_char, size_t};
-use std::io::Cursor;
-use std::os::raw::c_ushort;
-use std::sync::Arc;
-
-use rustls::{Certificate, PrivateKey};
-use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
-
-use rustls::sign::CertifiedKey;
 use rustls_result::NullParameter;
 
 /// All SignatureScheme currently defined in rustls.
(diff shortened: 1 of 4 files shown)


src/cipher.rs
```diff
@@ -64,35 +62,45 @@ pub extern "C" fn rustls_cipher_get_signature_scheme_name(
...
-            let out_n: &mut size_t = match out_n.as_mut() {
-                Some(out_n) => out_n,
-                None => return,
-            };
-            *out_n = 0;
             if buf.is_null() {
                 return;
             }
             slice::from_raw_parts_mut(buf as *mut u8, len as usize)
         };
```
tgeoghegan, line 69: I think this is the pattern we were talking about in Slack earlier. How about:
```
if buf.is_null() {
    return;
}
let write_buf: &mut [u8] = unsafe { slice::from_raw_parts_mut(buf as *mut u8, len as usize) };
```

src/cipher.rs
```diff
@@ -64,35 +62,45 @@ pub extern "C" fn rustls_cipher_get_signature_scheme_name(
...
+/// https://docs.rs/rustls/0.19.0/rustls/sign/struct.CertifiedKey.html
 pub struct rustls_certified_key {
     // We use the opaque struct pattern to tell C about our types without
     // telling them what's inside.
     // https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
     _private: [u8; 0],
 }
 
+/// Build a `rustls_certified_key` from a certificate chain and a private key.
+/// `cert_chain` must point to an array of `cert_chain_len` bytes, containing
```
tgeoghegan, line 90:
```suggestion
/// `cert_chain` must point to a buffer of `cert_chain_len` bytes, containing
```
Per discussion in Slack earlier today

src/server.rs
```diff
@@ -105,6 +105,16 @@ pub extern "C" fn rustls_server_config_builder_set_ignore_client_order(
     }
 }
 
+/// Set the ALPN protocol list to the given protocols. `protocols` must point
+/// to a buffer of `rustls_slice_bytes` (built by the caller) with `len`
+/// elements. Each element of th buffer must point to a slice of bytes that
```
tgeoghegan, line 110:
```suggestion
/// elements. Each element of the buffer must point to a slice of bytes that
```

src/cipher.rs
```diff
@@ -64,35 +62,45 @@ pub extern "C" fn rustls_cipher_get_signature_scheme_name(
...
+/// a series of PEM-encoded certificates, with the end-entity (leaf)
+/// certificate first.
+///
+/// `private_key` must point to an array of `private_key_len` bytes, containing
```
tgeoghegan, line 94:
```suggestion
/// `private_key` must point to a buffer of `private_key_len` bytes, containing
```

tgeoghegan (changes requested):

src/cipher.rs
djc: (Is the Slack instance accessible to me? Might be useful...)

tgeoghegan (approved):

src/cipher.rs
jsha: Yep! Josh sent you an invite. Let me know if you didn't receive it.

PR #64, opened 2021-03-16
session storage in the API

icing:
Adding session storage handling from `rustls` to the API.

The 3 traits functions for server storage are mapped into 2 C callbacks and used also for the client version. There is a `SessionStoreBroker` which can be used in both configs.

This is the first time we have the C side copying bytes into a Rust provided buffer, but I see no better way to handle it in this scenario. The lookup happens on every new connection and needs to be thread safe. A global buffer is therefore out of the question.

I added test cases for the server side in `mod_tls`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -579,6 +579,9 @@ pub extern "C" fn rustls_client_session_write_tls(
 /// keys and values are highly sensitive data, containing enough information
 /// to break the security of the sessions involved.
 ///
+/// `userdata` must live as long as the config object and any sessions
+/// or other config created from that config object.
+///
 #[no_mangle]
 pub extern "C" fn rustls_client_config_builder_set_persistence(
     builder: *mut rustls_client_config_builder,
--- a/src/client.rs
+++ b/src/client.rs
@@ -13,6 +13,11 @@ use rustls::{
 
 use crate::error::{self, map_error, result_to_tlserror, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_slice_slice_bytes, rustls_str};
+use crate::session::{
+    rustls_session_store_get_callback, rustls_session_store_put_callback,
+    rustls_session_store_userdata, SessionStoreBroker, SessionStoreGetCallback,
+    SessionStorePutCallback,
+};
 use crate::{
     arc_with_incref_from_raw, ffi_panic_boundary, ffi_panic_boundary_bool,
     ffi_panic_boundary_generic, ffi_panic_boundary_ptr, ffi_panic_boundary_unit, rslice::NulByte,
@@ -565,6 +570,35 @@ pub extern "C" fn rustls_client_session_write_tls(
             Err(_) => return rustls_result::Io,
         };
         *out_n = n_written;
+        rustls_result::Ok
+    }
+}
+
+/// Register callbacks for persistence of TLS session data. This means either
+/// session IDs (TLSv1.2) or . Both
+/// keys and values are highly sensitive data, containing enough information
+/// to break the security of the sessions involved.
+///
+#[no_mangle]
+pub extern "C" fn rustls_client_config_builder_set_persistence(
+    builder: *mut rustls_client_config_builder,
+    userdata: rustls_session_store_userdata,
+    get_cb: rustls_session_store_get_callback,
+    put_cb: rustls_session_store_put_callback,
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let get_cb: SessionStoreGetCallback = match get_cb {
+            Some(cb) => cb,
+            None => return rustls_result::NullParameter,
+        };
+        let put_cb: SessionStorePutCallback = match put_cb {
+            Some(cb) => cb,
+            None => return rustls_result::NullParameter,
+        };
+        let config: &mut ClientConfig = try_ref_from_ptr!(builder, &mut ClientConfig);
+        config.set_persistence(Arc::new(SessionStoreBroker::new(
+            userdata, get_cb, put_cb
+        )));
         rustls_result::Ok
     }
 }
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -299,6 +299,55 @@ typedef struct rustls_client_hello {
 typedef const struct rustls_certified_key *(*rustls_client_hello_callback)(rustls_client_hello_userdata userdata, const struct rustls_client_hello *hello);
 
 /**
+ * Any context information the callback will receive when invoked.
+ */
+typedef void *rustls_session_store_userdata;
+
+/**
+ * Prototype of a callback that can be installed by the application at the
+ * `rustls_server_config` or `rustls_client_config`. This callback will be
+ * invoked by a TLS session when looking up the data for a TLS session id.
+ * `userdata` will be supplied as provided when registering the callback.
+ *
+ * The `buf` points to a buffer of `count` consecutive bytes where the
+ * callback is expected to copy the result to. The number of copied bytes
+ * needs to be written to `out_n`.
+ *
+ * If the value to copy is larger than `count`, the callback should never
+ * do a partial copy but instead remove the value from its store and
+ * act as if it was never found.
+ *
+ * The callback should return != 0 to indicate that a value was retrieved
+ * and written in its entirety into `buf`.
+ *
+ * When `remove_after` is != 0, the returned data needs to be removed
+ * from the store.
+ *
+ * NOTE: the passed in `key` and `buf` are only availabe during the
+ * callback invocation.
+ * NOTE: callbacks used in several sessions via a common config
+ * must be implemented thread-safe.
+ */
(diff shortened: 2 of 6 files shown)


src/session.rs
```diff
@@ -0,0 +1,159 @@
...
+            get_cb,
+            put_cb,
+        }
+    }
+
+    fn retrieve(&self, key: &[u8], remove: bool) -> Option<Vec<u8>> {
+        let key: rustls_slice_bytes = key.into();
+        // TODO: we need a buffer where th client can store the retrieved
+        // session value. What size should it have? 10k seems excessive...
+        let mut data: Vec<u8> = vec![0 as u8; 10 * 1024];
```
jsha, line 103: I think `as u8` is not needed here because of the type annotation.

(Some fiddly thoughts below):

For this use case I think we can allocate a fixed-size array on the stack, and then after the callback is done, we can turn that array into a Vec of exactly the right size before returning it.

Now there's the question of: is that better, or worse? Checking the code, it looks like the returned `Vec<u8>` is [immediately treated as a slice to be read into a session](https://github.com/ctz/rustls/blob/e3e626d4dc6f1ffa3ee99e43893582ec5e62ca21/rustls/src/server/tls13.rs#L529-L532), and then dropped.

So, disregard the suggestion about using a fixed-size array. Making a potentially-oversized Vec here is fine because it gets thrown away shortly. And it saves us a copy.

src/session.rs
```diff
@@ -0,0 +1,159 @@
...
+            }
+            None
+        }
+    }
+
+    fn store(&self, key: Vec<u8>, value: Vec<u8>) -> bool {
+        let key: rustls_slice_bytes = key.as_slice().into();
+        let value: rustls_slice_bytes = value.as_slice().into();
+        unsafe {
+            let cb = self.put_cb;
```
jsha, line 128: The `let` should be outside the `unsafe` block.

src/session.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.store(key, value)
+    }
+
+    fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
+        return self.retrieve(key, false);
+    }
+}
+
+unsafe impl Sync for SessionStoreBroker {}
+unsafe impl Send for SessionStoreBroker {}
```
jsha, line 159: We should have a comment documenting why we consider this `Sync` and `Send` (even if the answer is just "we have documented that the C code needs to be threadsafe.")

src/client.rs
```diff
@@ -568,3 +573,32 @@ pub extern "C" fn rustls_client_session_write_tls(
         rustls_result::Ok
     }
 }
+
+/// Register callbacks for persistence of TLS session data. This means either
+/// session IDs (TLSv1.2) or . Both
+/// keys and values are highly sensitive data, containing enough information
+/// to break the security of the sessions involved.
+///
```
jsha, line 581: This should document something about the lifetime of the `userdata` pointer. I.e. if it's non-NULL, the pointed-to data must live at least as long as the config object and any sessions created from that config object.

src/session.rs
```diff
@@ -0,0 +1,159 @@
...
+            put_cb,
+        }
+    }
+
+    fn retrieve(&self, key: &[u8], remove: bool) -> Option<Vec<u8>> {
+        let key: rustls_slice_bytes = key.into();
+        // TODO: we need a buffer where th client can store the retrieved
+        // session value. What size should it have? 10k seems excessive...
+        let mut data: Vec<u8> = vec![0 as u8; 10 * 1024];
+        let buffer = data.as_mut_slice();
```
jsha, line 104: Vec has [`as_mut_ptr`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_ptr), so we can skip this `let` and just do `data.as_mut_ptr()` / `data.len()` below.

src/crustls.h
```diff
@@ -235,6 +235,55 @@ typedef struct rustls_verify_server_cert_params {
...
+ * Prototype of a callback that can be installed by the application at the
+ * `rustls_server_config` or `rustls_client_config`. This callback will be
+ * invoked by a TLS session when a TLS session has been created and an id
+ * for later use is handed to the client/has been received from the server.
+ * `userdata` will be supplied as provided when registering the callback.
+ *
+ * The callback should return != 0 to indicate that the value has been
+ * successfully persisted in its store.
+ *
+ * NOTE: the passed in `key` and `val` are only availabe during the
```
jsha, line 280:
```suggestion
 * NOTE: the passed in `key` and `val` are only available during the
```

jsha (changes requested):
This is very nicely done. Thanks!

I think it's fine for C to write into a Rust-provided buffer. It's allowed and safe.

jsha:
As for how big the buffer should be: It will receive [`ServerSessionValue`s](https://github.com/ctz/rustls/blob/e3e626d4dc6f1ffa3ee99e43893582ec5e62ca21/rustls/src/msgs/persist.rs#L186-L195) and [`ClientSessionValue`s](https://github.com/ctz/rustls/blob/e3e626d4dc6f1ffa3ee99e43893582ec5e62ca21/rustls/src/msgs/persist.rs#L55-L67). Both of those are probably dominated by [`CertificatePayload`](https://github.com/ctz/rustls/blob/e3e626d4dc6f1ffa3ee99e43893582ec5e62ca21/rustls/src/msgs/handshake.rs#L1287-L1296), which is limited to 64KB. So we can be safe if we allocate 64KB + some headroom; maybe 65KB? That's an awfully large buffer. I would say let's make the big buffer and not worry about if for now; if profiling shows it is an issue, we can do various optimizations, like pooling buffers for reuse, or limiting the max size for sessions to something lower.

src/crustls.h
```diff
@@ -235,6 +235,55 @@ typedef struct rustls_verify_server_cert_params {
...
+ *
+ * When `remove_after` is != 0, the returned data needs to be removed
+ * from the store.
+ *
+ * NOTE: the passed in `key` and `buf` are only availabe during the
+ * callback invocation.
+ * NOTE: callbacks used in several sessions via a common config
+ * must be implemented thread-safe.
+ */
+typedef int (*rustls_session_store_get_callback)(rustls_session_store_userdata userdata, const struct rustls_slice_bytes *key, uint8_t *buf, size_t count, int remove_after, size_t *out_n);
```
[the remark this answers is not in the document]
icing: No big deal. Separating in and out clearly would give `userdata, key, remove_after, count, buf, out_n` but that would not be very readable either.

Made the return value a `rustls_result`, but looks weird to just convert it to a `bool` for rustls trait return value.

If `remove_after` has only a boolean value, let's not complicate things. I think we don't want `do_it` callbacks with an `rustls_operation` enum for all things now and future.

As to using `bool` in C: that touches on a sore sport, namely that Apache is still on `C89`(it's also why you do not see revolutionary C things as intermittent variable declarations in my code). 

Of course, `crustls` could declare C99 as its base.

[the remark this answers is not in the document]
jsha: > Separating in and out clearly would give userdata, key, remove_after, count, buf, out_n but that would not be very readable either.

This is a specific exception to the coding conventions in the README:

> As a minor exception to the above: When an output parameter is a byte buffer
> (*uint8_t), the next parameter will always be a size_t denoting the size of
> the buffer. This is considered part of the output parameters even though it is
> not directly modified.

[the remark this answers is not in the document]
jsha: > If remove_after has only a boolean value, let's not complicate things. I think we don't want do_it callbacks with an rustls_operation enum for all things now and future.

Fair enough. My reasoning for preferring enums over booleans / ints in function arguments is to make it clear at the call site what's happening. For example if I see:

`get_session(client_session, true, false, true)`

I don't know what all those fields do! But if I see:

`get_session(client_session, KEEP, ALLOW_EXPIRED, REQUIRE_SNI_MATCH)`

It's more clear. I wouldn't propose to have a single `rustls_operation` that tried to encompass all such values, but a specific enum for a specific situation.

Still, in this situation, there are no call sites in the C code. There is only one call site on the Rust side, and it is fairly close to where the variable is set, so the meaning is clear. And in the C code that defines the implementation, the variable will have a name chosen by the implementer (probably `remove_after`).

jsha (approved):

src/crustls.h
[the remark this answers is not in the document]
icing: I agree that it hinders readability on the callers side. Python's named args are really a nice way of tackling this.

We could change `remove_after: bool` to a `afterwards: rustls_post_op[KEEP, DELETE]`. The complication this brings is that the C implementer cannot  be 100% sure that these are the only `afterwards` values it will ever receive. And if it does, should it return a `NOT_IMPLEMENTED` or ignore it. That would need to be documented.

An alternative would be to adhere to the rustls pattern and introduce a 3rd callback type `take`, so the parameter is no longer needed in the `get` callback.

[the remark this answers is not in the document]
jsha: I'm satisfied with the current version; plenty of other work to do. :-)

PR #65, opened 2021-03-17
Icing/tls protocol

icing:
* Exposing known constants
 * Inquiring about support at the builder
 * setting the set of protocols to use at the builder
 * forcing export of otherwise unused `rustls_tls_version`
 * adding Makefile dependency on `cbindgen.toml`

### DIFF
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -594,6 +594,30 @@ void rustls_server_config_builder_free(struct rustls_server_config_builder *conf
 struct rustls_server_config_builder *rustls_server_config_builder_from_config(const struct rustls_server_config *config);
 
 /**
+ * Set the TLS protocol versions to use when negotiating a TLS session.
+ *
+ * `tls_version` is the version of the protocol, as defined in rfc8446,
+ * ch. 4.2.1 and end of ch. 5.1. Some values are defined in
+ * `rustls_tls_version` for convenience.
+ *
+ * `versions` will only be used during the call and the application retains
+ * ownership. `len` is the number of consecutive `ui16` pointed to by `versions`.
+ */
+enum rustls_result rustls_server_config_builder_set_versions(struct rustls_server_config_builder *builder,
+                                                             const uint16_t *tls_versions,
+                                                             size_t len);
+
+/**
+ * Find out if the given TLS protocol version is suported in the
+ * `rustls_server_config` that is being built.
+ * `tls_version` is the version of the protocol, as defined in rfc8446,
+ *  ch. 4.2.1 and end of ch. 5.1. Some values are defined in
+ * `rustls_tls_version` for convenience.
+ */
+bool rustls_server_config_builder_supports_version(struct rustls_server_config_builder *builder,
+                                                   uint16_t tls_version);
+
+/**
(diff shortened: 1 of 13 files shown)


src/crustls.h
```diff
@@ -593,6 +605,30 @@ void rustls_server_config_builder_free(struct rustls_server_config_builder *conf
...
 
+/**
+ * Set the TLS protocol versions to use when negotiating a TLS session.
+ *
+ * `tls_version` is the version of the protocol, as defined in rfc8446,
+ * ch. 4.2.1 and end of ch. 5.1. Some values are defined in
+ * `rustls_tls_version` for convenience.
+ *
+ * `versions` will only be used during the call and the application retains
+ * ownership. `len` is the number of consecutive `ui16` pointed to by `versions`.
```
jsha, line 616:
```suggestion
 * ownership. `len` is the number of consecutive `uint16_t` pointed to by `versions`.
```

src/server.rs
```diff
@@ -89,6 +90,55 @@ pub extern "C" fn rustls_server_config_builder_from_config(
...
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let config: &mut ServerConfig = try_ref_from_ptr!(builder, &mut ServerConfig);
+        config.versions.clear();
+        unsafe {
+            let x: &[u16] = slice::from_raw_parts(tls_versions, len);
+            for i in x {
+                match rustls_tls_version_from_u16(*i) {
+                    Some(pversion) => config.versions.push(pversion),
+                    None => return rustls_result::AlertIllegalParameter
```
jsha, line 115: The `Alert` values in `rustls_result` are meant for conveying a TLS alert. I'd prefer to allocate a new `rustls_result` value rather than use this one outside its original purpose.

src/server.rs
```diff
@@ -89,6 +90,55 @@ pub extern "C" fn rustls_server_config_builder_from_config(
...
+/// Find out if the given TLS protocol version is suported in the
+/// `rustls_server_config` that is being built.
+/// `tls_version` is the version of the protocol, as defined in rfc8446,
+///  ch. 4.2.1 and end of ch. 5.1. Some values are defined in
+/// `rustls_tls_version` for convenience.
+#[no_mangle]
+pub extern "C" fn rustls_server_config_builder_supports_version(
+    builder: *mut rustls_server_config_builder,
+    tls_version: u16,
+) -> bool {
```
jsha, line 132: Out of curiosity, is this something you need or anticipate needing in mod_tls?

icing: I'll remove this again. I was originally thinking that this method would be sufficient for `mod_tls`. I have to issue open at https://github.com/ctz/rustls/issues/577 to allow the API user to detect and iterate the supported versions.

It might seem far fetch to prepare for future protocol versions, but I guess people felt the same at the time `SSLv3` was hot.

src/server.rs
icing: Changed. Added rustls_result::InvalidParameter.

src/server.rs
icing: Removed the function again.

jsha (approved):

PR #67, opened 2021-03-18
Pass through close_notify alerts

jsha:
In rustls, close_notify is expressed as an `io::Error` rather than a `TLSError`, presumably to be compatible with the `Read` trait:

https://github.com/ctz/rustls/blob/75d08697ab8ddbe3c0051dee195674b187c1717c/rustls/src/session.rs#L623-L628

https://github.com/ctz/rustls/blob/75d08697ab8ddbe3c0051dee195674b187c1717c/rustls/src/session.rs#L845-L850

In the C wrapper, we don't have to follow the `Read` trait, and can return the appropriate corresponding rustls_result. The functions no longer set `*out_n = 0` because of our "no partial failures" convention.

Update crustls-demo apropriately.

https://tools.ietf.org/html/rfc8446#section-6.1

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,8 +1,8 @@
 use libc::{c_char, size_t};
+use std::convert::TryInto;
 use std::io::{BufReader, Cursor, Read, Write};
 use std::ptr::null;
 use std::slice;
-use std::{convert::TryInto, io::ErrorKind::ConnectionAborted};
 use std::{ffi::CStr, sync::Arc};
 use std::{ffi::OsStr, fs::File};
 use webpki::DNSNameRef;
@@ -15,8 +15,8 @@ use crate::error::{self, map_error, result_to_tlserror, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_slice_slice_bytes, rustls_str};
 use crate::{
     arc_with_incref_from_raw, ffi_panic_boundary, ffi_panic_boundary_bool,
-    ffi_panic_boundary_generic, ffi_panic_boundary_ptr, ffi_panic_boundary_unit, rslice::NulByte,
-    try_ref_from_ptr,
+    ffi_panic_boundary_generic, ffi_panic_boundary_ptr, ffi_panic_boundary_unit, is_close_notify,
+    rslice::NulByte, try_ref_from_ptr,
 };
 use rustls_result::NullParameter;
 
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -396,7 +398,7 @@ send_request_and_read_response(int sockfd,
         if(result == CRUSTLS_DEMO_AGAIN) {
           break;
         }
-        else if(result == CRUSTLS_DEMO_EOF) {
+        else if(result == CRUSTLS_DEMO_CLOSE_NOTIFY) {
```
tgeoghegan, line 401:
```suggestion
        else if(result == CRUSTLS_DEMO_CLOSE_NOTIFY || result == CRUSTLS_DEMO_EOF) {
```
`do_read` can still return `CRUSTLS_DEMO_EOF`, right?

src/server.rs
```diff
@@ -378,12 +378,10 @@ pub extern "C" fn rustls_server_session_read(
         let out_n: &mut size_t = try_ref_from_ptr!(out_n, &mut size_t);
         let n_read: usize = match session.read(read_buf) {
             Ok(n) => n,
-            // The CloseNotify TLS alert is benign, but rustls returns it as an Error. See comment on
-            // https://docs.rs/rustls/0.19.0/rustls/struct.ServerSession.html#impl-Read.
-            // Log it and return EOF.
+            // Rustls turns close_notify alerts into `io::Error` of kind `ConnectionAborted`.
+            // https://docs.rs/rustls/0.19.0/rustls/struct.ClientSession.html#impl-Read.
             Err(e) if e.kind() == ConnectionAborted && e.to_string().contains("CloseNotify") => {
```
tgeoghegan, line 383: Could you factor this predicate into a function `is_close_notify(error: std::io::Error) -> bool` that could be shared between `server.rs` and `client.rs`? Or maybe provide `impl From<io::Error> for rustls_result` so you can just do
```
match result {
    Ok(n) => n,
    Err(e) => return e.into();
};
```

tgeoghegan (changes requested):

src/main.c
jsha: It can return EOF, but we actually want that to be an error. It falls through to the `else if(result != CRUSTLS_DEMO_OK)` on the next line.

It's an error because in theory we should always get `CRUSTLS_DEMO_CLOSE_NOTIFY` before we get an EOF.

tgeoghegan: That makes sense. Then the remaining nit is that instead of exiting successfully on `CRUSTLS_DEMO_CLOSE_NOTIFY`, this program could instead enforce that it then gets `CRUSTLS_DEMO_EOF` in a timely manner.

src/client.rs
```diff
@@ -486,12 +486,10 @@ pub extern "C" fn rustls_client_session_read(
         let out_n: &mut size_t = try_ref_from_ptr!(out_n, &mut size_t);
         let n_read: usize = match session.read(read_buf) {
             Ok(n) => n,
-            // The CloseNotify TLS alert is benign, but rustls returns it as an Error. See comment on
+            // Rustls turns close_notify alerts into `io::Error` of kind `ConnectionAborted`.
             // https://docs.rs/rustls/0.19.0/rustls/struct.ClientSession.html#impl-Read.
```
tgeoghegan, line 490: final nit: this comment belongs in the `is_close_notify` function.

tgeoghegan (approved):

PR #68, opened 2021-03-18
querying and configuring tls ciphers

icing:
This PR adds the following:

 * query the negotiated tls cipher from a client/server session
 * inspect all ciphers supported by rustls using a callback
 * set the list of ciphers at a client/server config


jsha:
Replaced by #79

PR #75, opened 2021-03-20
Simplify ffi_panic_boundary

jsha:
Having a separate macro for each possible return type was clunky. Now,
we have a trait that can be implemented for each return type we care
about, and a single `ff_ipanic_boundary` macro that uses that trait to
return the appropriate type for any situation.

Move the panic-handling code into its own file.

By the way, I intend to apply the same treatment to simplify try_ref_from_ptr,
but I want to do it after #73 lands. One thing at a time.

### DIFF
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -7,10 +7,6 @@ use std::ptr::{null, null_mut};
 // rustls_result, we return a dedicated error code: `Panic`. For functions
 // that don't return rustls_result, we return a default value: false, 0, or
 // null. This trait provides that logic.
-// Note: It's tempting to do a blanket `impl<T> PanicOrDefault for T: Default`.
-// However, that would conflict with the impls for `*mut T` and `*const T`.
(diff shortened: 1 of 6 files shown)


src/panic.rs
```diff
@@ -0,0 +1,69 @@
...
+impl PanicOrDefault for rustls_result {
+    fn value() -> Self {
+        rustls_result::Panic
+    }
+}
+
+impl PanicOrDefault for u16 {
+    fn value() -> Self {
+        Default::default()
+    }
```
tgeoghegan, line 27: nit: To reduce repetition, you could hoist this implementation of `value()` up into a default implementation on `trait PanicOrDefault`, and then the `impl`s for `u16`, `usize`, `bool` and `()` are reduced to opting into `PanicOrDefault`.

tgeoghegan (approved):

tgeoghegan (approved):
`Defaultable` is even better than a default impl on `PanicOrDefault`! Re-approved.

PR #76, opened 2021-03-23
Remove second argument of try_ref_*

jsha:
And add `try_slice!` / `try_mut_slice!`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,18 +62,6 @@ macro_rules! try_mut_slice {
     };
 }
 
-#[macro_export]
-macro_rules! ffi_panic_boundary_generic {
-    ( $retval:expr, $($tt:tt)* ) => {
-        match ::std::panic::catch_unwind(|| {
-            $($tt)*
-        }) {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -40,6 +40,40 @@ pub(crate) trait CastPtr {
...
+        if $ptr.is_null() {
+            return crate::panic::NullParameterOrDefault::value();
+        } else {
+            unsafe { slice::from_raw_parts_mut($ptr, $count as usize) }
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! ffi_panic_boundary_generic {
```
tgeoghegan, line 66: I don't think this macro is used anywhere. Shouldn't everything just use `panic::ffi_panic_boundary`?

tgeoghegan (changes requested):
Nice, I'm glad to see the default values removed from every usage of `try_{ref,mut}_from_ptr!`

tgeoghegan (approved):

Issue #77, opened 2021-03-23
Bazel


sayrer:
I started working on getting this project to compile with Bazel and BuildBuddy. This effort will result in reproducible builds and cached build results (meaning that something already done on a branch won't require a recompile).

I'll need to use `cargo-raze` to pull in the Cargo deps. But, this way, the project will have a build that pulls in a fixed Rust and C/C++ toolchain. (Bazel downloads it all, and doesn't depend on the local machine)

https://github.com/grafica/crustls/tree/buildbuddy

sayrer:
First try: https://app.buildbuddy.io/invocation/ee9c8180-46fd-4834-9636-e66cc3a48df5

Failed because none of the Cargo libs are downloaded for Bazel (yet).

sayrer:
BuildBuddy is buggy in building the `cargo-raze` tool itself, but things are working fine so far on my local Ubuntu x64 system. I have Bazel automatically producing all of the dependencies from Cargo, so the next step should be to get the Rust files compiling.

sayrer:
This vendors all of the Cargo stuff for Bazel, and produces a working BUILD.bazel file at the top level for them.

https://github.com/grafica/crustls/tree/de3a56683dc436bfc118395fbc86232489aa61a9

I'll see if I can get the crustls Rust files compiling now.

sayrer:
Almost works now, except the Bazel build of Ring doesn't put some data files in the Bazel sandbox:

```
INFO: From Compiling Rust rlib ring v0.16.20 (86 files):
error: couldn't read cargo/vendor/ring-0.16.20/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der: No such file or directory (os error 2)
   --> cargo/vendor/ring-0.16.20/src/ec/curve25519/ed25519/signing.rs:266:12
    |
266 |     bytes: include_bytes!("ed25519_pkcs8_v2_template.der"),
```

The file is there, Bazel just hasn't tracked it and included it in its sandbox:

```
$ ls cargo/vendor/ring-0.16.20/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der
cargo/vendor/ring-0.16.20/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der
```

sayrer:
Looks a known roadblock:
https://github.com/google/cargo-raze/issues/41#issuecomment-638986696

sayrer:
[...]

PR #78, opened 2021-03-25
Bazel build

sayrer:
As discussed in #77, a Bazel build to ensure reproducibility. Not caching in the cloud yet. The only change to the Makefile project was to stop checking in crustls.h and generate it by default.


sayrer:
Cloud build in 53 seconds on the first try:

https://app.buildbuddy.io/invocation/0cf7aeab-d5d9-4573-9966-6997002d93cf

Cloud build (with lots of cached stuff...) in .3 seconds on the follow-up:

https://app.buildbuddy.io/invocation/0fdbe91f-aed7-40a6-9cd9-0efd9d150fcc

`rules_rust` and BuildBuddy still seem a bit buggy when a `cc_toolchain` is supplied, rather than using the system C/C++ compiler.

jsha:
This is really neat! I appreciate all the work you put in on this - and sorry I didn't chime in earlier on the issue.

It seems like Bazel builds add a lot of complexity to what is currently a fairly small project. I'm not ready to take on that complexity now, but might be in the future.

sayrer:
Oh, no sweat, I was just doing it to checkout BuildBuddy anyway.

PR #79, opened 2021-03-26
Allow enumeration and config of ciphersuites.

jsha:
Define an opaque struct for SupportedCipherSuite, and an accessor for ALL_CIPHERSUITES. Add configuration for setting the list of cipher suites in ServerConfig, and for getting the negotiated cipher suite from a ServerSession.

@icing this is a counterproposal to #68. I wanted to try and avoid the complexity of the callback approach. ALL_CIPHERSUITES is a static, so we can rely on it living for the whole life of the program, which makes iterating over it somewhat easier. I haven't yet implemented parallel methods for ClientConfig and ClientSession because I wanted to get feedback first.

### DIFF
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -6,6 +6,12 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/**
+ * The number of cipher suites supported by rustls. You may call `rustls_all_ciphersuites_get()`
+ * with indexes lower than this number.
+ */
+#define RUSTLS_ALL_CIPHERSUITES_LEN 9
+
 typedef enum rustls_result {
   RUSTLS_RESULT_OK = 7000,
   RUSTLS_RESULT_IO = 7001,
@@ -194,6 +200,11 @@ typedef struct rustls_slice_slice_bytes rustls_slice_slice_bytes;
 typedef struct rustls_slice_str rustls_slice_str;
 
 /**
+ * A cipher suite supported by rustls.
+ */
+typedef struct rustls_supported_ciphersuite rustls_supported_ciphersuite;
+
+/**
  * User-provided input to a custom certificate verifier callback. See
  * rustls_client_config_builder_dangerous_set_certificate_verifier().
  */
@@ -373,6 +384,20 @@ typedef const struct rustls_certified_key *(*rustls_client_hello_callback)(rustl
  * and NUL terminated. Returns the number of bytes written before the NUL.
  */
 size_t rustls_version(char *buf, size_t len);
+
+/**
+ * Return a 16-bit unsigned integer corresponding to this cipher suite's assignment from
+ * <https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4>.
+ * The bytes from the assignment are interpreted in network order.
+ */
+uint16_t rustls_supported_ciphersuite_get_suite(const struct rustls_supported_ciphersuite *supported_ciphersuite);
+
+/**
+ * Get a pointer to a member of rustls' list of supported cipher suites. This will return non-NULL
+ * for 0 <= i < RUSTLS_ALL_CIPHERSUITES_LEN. The returned pointer is valid for the lifetime of
+ * the program and may be used directly when building a ClientConfig or ServerConfig.
+ */
+const struct rustls_supported_ciphersuite *rustls_all_ciphersuites_get(size_t i);
 
 /**
  * Build a `rustls_certified_key` from a certificate chain and a private key.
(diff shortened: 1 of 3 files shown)


src/crustls.h
```diff
@@ -6,6 +6,12 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/**
+ * The number of cipher suites supported by rustls. You may call `rustls_all_ciphersuites_get()`
+ * with indexes lower than this number.
+ */
+#define RUSTLS_ALL_CIPHERSUITES_LEN 9
```
icing, line 13: If you deploy a new version of crustls/rustls on a system as shared library, the application will not be recompiled and use the previous number of ciphers.

src/crustls.h
```diff
@@ -374,6 +385,20 @@ typedef const struct rustls_certified_key *(*rustls_client_hello_callback)(rustl
...
+ * The bytes from the assignment are interpreted in network order.
+ */
+uint16_t rustls_supported_ciphersuite_get_suite(const struct rustls_supported_ciphersuite *supported_ciphersuite);
+
+/**
+ * Get a pointer to a member of rustls' list of supported cipher suites. This will return non-NULL
+ * for 0 <= i < RUSTLS_ALL_CIPHERSUITES_LEN. The returned pointer is valid for the lifetime of
+ * the program and may be used directly when building a ClientConfig or ServerConfig.
+ */
+const struct rustls_supported_ciphersuite *rustls_all_ciphersuites_get(size_t i);
```
icing, line 400: For applications using a previous higher number from a `#define RUSTLS_ALL_CIPHERSUITES_LEN`, this call will return NULL unexpectedly on a new crustls version with less supported ciphers.

icing: It would be better to have something like this:

```
rustls_supported_ciphersuite *suite;
int i = 0;
while ((cipher = rustls_supported_ciphersuite_get(i++)) {
   /* remember cipher */
}
```

icing (commented):
I think the #define of a number needs improvement. Otherwise the functions should work.

If rustls does not expose named lookups for all ciphers in its API, the application will keep maps from u16 to the IANA names anyway. It that sense the conversions from pointers to u16 seem a bit unyieldy, but are not a problem per se.

src/crustls.h
jsha: Ooh, this is a good pattern. Thanks for the suggestion! And good point about ABI compatibility.

jsha:
> If rustls does not expose named lookups for all ciphers in its API, the application will keep maps from u16 to the IANA names anyway. It that sense the conversions from pointers to u16 seem a bit unyieldy, but are not a problem per se.

To be clear: your branch #68 does not expose named lookups either, right? I'm trying to understand if this is something that's missing from #79 vs #68, or if it's an additional comment.

icing:
No, I have that in mod_tls. I wanted to say: either it is in rustls, or the application does it. I do not think that crustls should offer this.

icing (commented):
Adapted `mod_tls` code and test suite works with this. 

`mod_tls` internal handling remains uint16 based however. static definitions cannot use rust*, so uint16 remains the identifier of choice for sorting/filtering tls ciphers.

Minor nit: the interface makes no `typedef`, but carries around the `struct rustls_supported_ciphersuite`. Can we get rid of the `struct` part?

jsha:
I don't understand the nit. In crustls.h we have a typedef:

```c
typedef struct rustls_supported_ciphersuite rustls_supported_ciphersuite;
```

It's true that the declaration for the methods uses the `struct` keyword, e.g.

```c
enum rustls_result rustls_server_config_builder_set_ciphersuites(struct rustls_server_config_builder *builder,
                                                                 const struct rustls_supported_ciphersuite *const *ciphersuites,
                                                                 size_t len);
```

But that is consistent with all our declarations:

```c
enum rustls_result rustls_server_config_builder_set_certified_keys(struct rustls_server_config_builder *builder,
                                                    const struct rustls_certified_key *const *certified_keys,
                                                    size_t certified_keys_len);
```

icing:
Just tried again and it sure works with just the typedef'ed name. Not sure what did go wrong here locally, beside my brain becoming confused. 

Sorry for the noise.

tgeoghegan (approved):

icing:
Do we proceed on this? For my planning it would be nice to know if this part becomes stable or need to be rewritten on some future design.

jsha:
Yep, let's proceed. I just need to update to address Tim's comments.

src/crustls.h
```diff
@@ -386,14 +386,18 @@ size_t rustls_version(char *buf, size_t len);
  */
 uint16_t rustls_supported_ciphersuite_get_suite(const struct rustls_supported_ciphersuite *supported_ciphersuite);
 
+/**
+ * Return the length of rustls' list of supported cipher suites.
+ */
+uintptr_t rustls_all_ciphersuites_len(void);
```
tgeoghegan, line 392: Is `uintptr_t` what `cbindgen` generates for `usize`? Interesting. I think I might have used `size_t` or `off_t` if I were writing this C function by hand, but `uintptr_t` is close enough that it's not worth fighting with `cbindgen` over it.

tgeoghegan (approved):
LGTM!

src/crustls.h
jsha: That's interesting. In most other cases, `cbindgen` generates `size_t`. There are two cases I see where it doesn't:

```
void rustls_client_session_get_alpn_protocol(const struct rustls_client_session *session,
                                             const uint8_t **protocol_out,
                                             uintptr_t *protocol_out_len);
```

(perhaps because it's a pointer?)

    uintptr_t rustls_all_ciphersuites_len(void);

(perhaps because it's a return value?)

Issue #11, opened 2020-12-28
Please consider using cargo-c


lu-zero:
It has the logic to build header, library (including the dynamic variant) and pkg-config file.

https://crates.io/crates/cargo-c

jsha:
Thanks for the suggestion! I'm still getting good at FFI and packaging a C library, so I want to do some of this stuff manually for now to make sure I understand how it all works. But I will keep this in mind, particularly when I need to generate a pkg-config file.

lu-zero:
Let me know if you need help and I'll send you a pull request :)

PR #82, opened 2021-04-20
Initial cargo-c conversion

lu-zero:
- [x] Remove the now-unneeded entries in Cargo.toml
- [x] Update the documentation 
- [ ] Port the tests to use `cargo ctest`

Addresses #11

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,3 @@ libc = "0.2"
 rustls-native-certs = "0.5.0"
 sct = "0.6.0"
 rustls-pemfile = "0.2.0"
-
-[dev_dependencies]
-cbindgen = "*"
-
-[lib]
-name = "crustls"
-crate-type = ["staticlib"]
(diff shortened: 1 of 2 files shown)


djc:
Would be nice to be a bit more explicit in your PR about the benefits: what does cargo-c provide over the previous setup?

lu-zero:
The top 4 are:
- Correct shared library at least for *BSD, Linux, Windows and Mac platforms
- Pkgconfig files containing the correct link line for static linking.
- The ability to merge this crate back to the main rustls and use the normal crate to install the C-API without having to have Cargo.toml options that impact the normal crate usage, see how we do in rav1e and other projects.
- C-API tests integrated with the normal test harness using `cargo ctest` (yet to be done in this PR)

lu-zero:
I can complete the set if there is consensus on using it.

jsha:
I don't plan to integrate cargo-c right now, but thank you for showing how it could work.

lu-zero:
Feel free to poke me when/if you'd like to add it and I'll freshen the patch. I can spin off the pkg-config generation code so you can use it for #220 in case it could help you.

Issue #81, opened 2021-04-15
application session data


icing:
We install C callbacks into `server_config` which will later be invoked from `server_session` instances created for individual connections. These callbacks need access to connection related information for reasons of proper logging or use of correct memory pools for allocations.

The `server_session` should carry a `void *` application data add during its creation. Passing this to callback installed in the base `server_config` would allow an application easy and fast access to the connection context.

In the current design, this is only possible to achieve when the complete `server_config` is created again and again for each connection. Which seems not desirable. Alternatively, having the application maintain global maps of connection <->  server_session pairing seems overly complicated.

djc:
This should be covered by https://github.com/ctz/rustls/issues/89, right? I'm working towards that.

icing:
I guess, when the app sees the full client hello before setting up the rustls server session, this would no longer be needed, unless we find some later events we need to handle in a callback.

However, this also means that rustls server config is not really used as it seems to be intended, e.g. as a global , unmutable set of TLS configs from which sesions are created for each connection. Instead, each connection would first create its own config instance and from that a session. Because there are connection specific things (from the app point of view) which can only be modified in a config, as far as I understood.

djc:
In my understanding, the goal of ctz/rustls#89 is to create a new type, `Acceptor`, that doesn't a `ServerConfig`. It can be fed with bytes coming in from the second, and you can call `client_hello() -> Option<ClientHello>` on it. Once `client_hello()` returns `Some`, there will be a way to create a `ServerSession` from the `Acceptor` and an `&Arc<ServerConfig>`. Thus, you can still have a small set of immutable `ServerConfig`s if desirable.

icing:
Understood. But if you install a client cert verifier, , it is unable to log the success/failure in relation to the connection it happens on. In pseudo code:

```
setup_server(server s) {
  client_verifier cv = my_client_verifier(s);
  config = rustls_server_config(vc);
 ...
}

setup_connection(conn c) {
  server s = server_selected_by_conn( c );
  session = ServerSession(config_for_server( s ));
  ...
}
```

The `client_verifier::verify_client_cert()` is later invoked by the `ServerSession`, but it has no way to find the connection `c`, or? without some global map.

jsha:
[...]

PR #83, opened 2021-04-21
Prototype per-session userdata.

jsha:
This introduces a `userdata: *mut c_void` parameter for
rustls_server_session_new, and changes `rustls_server_session` to be a
struct containing a Box<ServerSession> and a userdata pointer.

In rustls_server_session_process_new_packets, we set the thread-local
USERDATA to the session's value, call
ServerSession::process_new_packets(), then clear the thread-local.

This updates SessionStoreBroker to get its userdata parameter from the
session's userdata, rather than a config-wide userdata set at construction.

For this demo, disable catch_unwind because it introduces some issues
with UnwindSafe. We would have to resolve that before landing something
like this. Also, realistically we would want to take steps to ensure
userdata was reliably set and cleared before entry and exit from any
relevant rustls functions.

This compiles, but is completely untested so far. It's meant to give an
approximate shape of a solution for further discussion.

Related to #81.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -13,9 +13,8 @@ use rustls::{
 use crate::error::{self, map_error, result_to_tlserror, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_slice_slice_bytes, rustls_str};
 use crate::session::{
-    rustls_session_store_get_callback, rustls_session_store_put_callback,
-    rustls_session_store_userdata, SessionStoreBroker, SessionStoreGetCallback,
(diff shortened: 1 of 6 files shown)


djc:
Looks more or less like what I imagined it to be, except in my mind we would have a guard type that takes care of inserting into the `RefCell` and would clear it on `Drop`. This would make it so that you don't have to remember about dropping it.

Also I don't think this takes care of nested calls, although it's still a little unclear to me if that's actually needed.

jsha:
Closing in favor of #86.

PR #84, opened 2021-04-25
Add ALPN support for clients

jsha:


src/client.rs
```diff
@@ -297,6 +298,36 @@ pub extern "C" fn rustls_client_config_builder_load_roots_from_file(
     }
 }
 
+/// Set the ALPN protocol list to the given protocols. `protocols` must point
+/// to a buffer of `rustls_slice_bytes` (built by the caller) with `len`
+/// elements. Each element of the buffer must point to a slice of bytes that
+/// contains a single ALPN protocol from
+/// https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids.
```
djc, line 305: rustls doesn't enforce any restriction like that, so maybe this should be removed/worded differently?

src/client.rs
```diff
@@ -297,6 +298,36 @@ pub extern "C" fn rustls_client_config_builder_load_roots_from_file(
...
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let config: &mut ClientConfig = try_mut_from_ptr!(builder);
+        let protocols: &[rustls_slice_bytes] = try_slice!(protocols, len);
+
+        let mut vv: Vec<Vec<u8>> = Vec::new();
+        for p in protocols {
+            let v: &[u8] = try_slice!(p.data, p.len);
+            vv.push(v.to_vec());
+        }
```
djc, line 325: Pending on the exact semantics of `try_slice!()`, maybe rewrite this as an iterator chain, that collects into `vv`? Would allow the code to allocate just once with the exact number of elements necessary. (Alternatively, use `Vec::with_capacity()` to allocate the right size.)

src/crustls.h
```diff
@@ -497,6 +497,22 @@ enum rustls_result rustls_client_config_builder_load_native_roots(struct rustls_
 enum rustls_result rustls_client_config_builder_load_roots_from_file(struct rustls_client_config_builder *config,
                                                                      const char *filename);
 
+/**
+ * Set the ALPN protocol list to the given protocols. `protocols` must point
+ * to a buffer of `rustls_slice_bytes` (built by the caller) with `len`
+ * elements. Each element of the buffer must point to a slice of bytes that
```
djc, line 503: As above.

src/client.rs
jsha: I tried this but ran into some lifetime issues with [slice::from_raw_parts](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html). I believe it was inferring a too-short lifetime for resulting slice. I could possible annotate it at the cost of some complexity. `Vec::with_capacity` also sounds like a nice approach.

src/client.rs
jsha: The restriction being "contains ... from <url>"? How about:

Each element of the buffer must point to a slice of bytes containing a single ALPN protocol. Standard ALPN protocols are registered at <url>.

src/client.rs
jsha: I took another stab at this and realized it wasn't a lifetime issue. The problem is: when any of the `p.data` items is NULL, we want to early-return from the function with `NullParameter`. We can't do that from inside the closure of an iterator chain. We could make the closure return `Option<Vec<u8>>`, and then turn any Nones into an early return, but this seems like it would be more complicated.

src/client.rs
```diff
@@ -399,6 +431,53 @@ pub extern "C" fn rustls_client_session_is_handshaking(
...
+            None => 0
+        }
+    }
+}
+
+/// Get the ALPN protocol that was negotiated, if any. Store a pointer to a
+/// borrowed buffer of bytes, and that buffer's len, in the output parameters.
+/// The borrow lives as long as the session.
+/// If the session is still handshaking, or no ALPN protocol was negotiated,
+/// store NULL and 0 in the output parameters.
```
tgeoghegan, line 455: Nit: 
```suggestion
/// Get the ALPN protocol that was negotiated, if any. Stores a pointer to a
/// borrowed buffer of bytes, and that buffer's len, in the output parameters.
/// The borrow lives as long as the session.
/// If the session is still handshaking, or no ALPN protocol was negotiated,
/// stores NULL and 0 in the output parameters.
```
Seems like the more appropriate verb tense given other doccomments in the crate.

tgeoghegan (approved):
LGTM, with a small number of possibly doccomment improvements.

src/client.rs
djc: The standard solution for this is to return a `Result` type from the `map()` closure. `FromIterator` (the trait backing `collect()`) knows how to collect into a `Result`, so in this case you could collect into `Result<Vec<&[u8]>, NullParameter>` or something like that.

jsha: Ah, thanks for the tip! The collect into Result is quite nice. Here's what I wound up with:

```rust
        let result = protocols.iter().map(|p| {
            if p.data.is_null() {
                Err(())
            } else {
                unsafe { Ok(slice::from_raw_parts(p.data, p.len).to_vec()) }
            }
        }).collect();
        match result {
            Ok(vv) => {
                config.set_protocols(&vv);
                rustls_result::Ok
            }
            Err(()) => rustls_result::NullParameter,
        }
```

14 lines, vs 7 for the procedural implementation (which I also find slightly more straightforward). If there's a way I can improve on that, lemme know!

djc: Ha, challenge accepted. Here's my shot (untested):

```rust
protocols.iter().map(|p| match p.data.is_null() {
    false => Ok(unsafe { slice::from_raw_parts(p.data, p.len).to_vec() }),
    true => Err(rustls_result::NullParameter),
}).collect().map(|vv| {
    config.set_protocols(&vv);
    rustls_result::Ok
}).unwrap_or_else(|e| e)
```

Also 7 lines, but I think I prefer the procedural version in this case.

jsha: Ooh, nice code golf! That second `.map()` after collect is particularly clever.

PR #85, opened 2021-04-26
Add OCSP Stapling provisioning

icing:
Since OCSP response data is part of a rustls `CertifiedKey`, it needs to be added (when available) during the resolution of the server session certificate. This PR adds a callback to the API that can be installed (or given as NULL) when setting the list of CertifiedKeys on a SessionConfig.

Rationale: a callback is necessary, since OCSP response data has a max lifetime, commonly of a few days. Since server restarts are independent of this, the timeout of a known response might happen at any time. Also, when starting a server, no (valid) response might be readily available.

This makes OCSP response provisioning more dynamic than the rustls API seems to foresee. Hence a callback interface. The provided buffer makes sure that copying is cheap and buffer lifetimes are clear. Since a resolved certified key needs to be cloned anyway and the OCSP data is "taken" by the session, this should add little to no overhead.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -9,8 +9,10 @@ use rustls::{Certificate, PrivateKey};
 use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 
 use crate::error::rustls_result;
-use crate::{ffi_panic_boundary, try_ref_from_ptr, CastPtr};
+use crate::rslice::rustls_slice_bytes;
+use crate::{arc_with_incref_from_raw, ffi_panic_boundary, try_ref_from_ptr, try_slice, CastPtr};
 use rustls_result::NullParameter;
+use std::ops::Deref;
 
 /// The complete chain of certificates to send during a TLS handshake,
 /// plus a private key that matches the end-entity (leaf) certificate.
@@ -104,6 +106,43 @@ pub extern "C" fn rustls_certified_key_build(
         };
         let certified_key = Arc::into_raw(Arc::new(*certified_key)) as *const _;
         *certified_key_out = certified_key;
+        return rustls_result::Ok
+    }
+}
+
+/// Create a copy of the rustls_certified_key with the given OCSP response data
+/// as DER encoded bytes. The OCSP response may be given as NULL to clear any
+/// possibly present OCSP data from the cloned key.
+/// The cloned key is independent from its original and needs to be freed
+/// by the application.
+#[no_mangle]
+pub extern "C" fn rustls_certified_key_clone_with_ocsp(
+    key: *const rustls_certified_key,
+    ocsp_response: *const rustls_slice_bytes,
+    cloned_key_out: *mut *const rustls_certified_key,
+) -> rustls_result {
+    ffi_panic_boundary! {
+        let cloned_key_out: &mut *const rustls_certified_key = unsafe {
+            match cloned_key_out.as_mut() {
+                Some(c) => c,
+                None => return NullParameter,
+            }
+        };
+        let certified_key: Arc<CertifiedKey> = unsafe {
+            match (key as *const CertifiedKey).as_ref() {
+                Some(c) => arc_with_incref_from_raw(c),
+                None => return NullParameter,
+            }
+        };
+        let mut new_key = certified_key.deref().clone();
+        if !ocsp_response.is_null() {
+            let ocsp_slice = unsafe{ &*ocsp_response };
+            new_key.ocsp = Some(Vec::from(try_slice!(ocsp_slice.data, ocsp_slice.len)));
+        }
+        else {
+            new_key.ocsp = None;
+        }
+        *cloned_key_out = Arc::into_raw(Arc::new(new_key)) as *const _;
         return rustls_result::Ok
     }
 }
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -138,8 +138,7 @@ pub extern "C" fn rustls_certified_key_clone_with_ocsp(
         if !ocsp_response.is_null() {
             let ocsp_slice = unsafe{ &*ocsp_response };
             new_key.ocsp = Some(Vec::from(try_slice!(ocsp_slice.data, ocsp_slice.len)));
-        }
-        else {
+        } else {
             new_key.ocsp = None;
         }
         *cloned_key_out = Arc::into_raw(Arc::new(new_key)) as *const _;
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -199,12 +199,6 @@ typedef struct rustls_slice_str rustls_slice_str;
 typedef struct rustls_supported_ciphersuite rustls_supported_ciphersuite;
 
 /**
- * User-provided input to a custom certificate verifier callback. See
- * rustls_client_config_builder_dangerous_set_certificate_verifier().
- */
-typedef void *rustls_verify_server_cert_user_data;
-
-/**
  * A read-only view on a Rust byte slice.
  *
  * This is used to pass data from crustls to callback functions provided
@@ -221,6 +215,12 @@ typedef struct rustls_slice_bytes {
   const uint8_t *data;
   size_t len;
 } rustls_slice_bytes;
+
+/**
+ * User-provided input to a custom certificate verifier callback. See
+ * rustls_client_config_builder_dangerous_set_certificate_verifier().
+ */
+typedef void *rustls_verify_server_cert_user_data;
 
 /**
  * A read-only view on a Rust `&str`. The contents are guaranteed to be valid
@@ -421,6 +421,17 @@ enum rustls_result rustls_certified_key_build(const uint8_t *cert_chain,
                                               const uint8_t *private_key,
                                               size_t private_key_len,
                                               const struct rustls_certified_key **certified_key_out);
+
+/**
+ * Create a copy of the rustls_certified_key with the given OCSP response data
+ * as DER encoded bytes. The OCSP response may be given as NULL to clear any
+ * possibly present OCSP data from the cloned key.
+ * The cloned key is independent from its original and needs to be freed
+ * by the application.
+ */
+enum rustls_result rustls_certified_key_clone_with_ocsp(const struct rustls_certified_key *key,
+                                                        const struct rustls_slice_bytes *ocsp_response,
+                                                        const struct rustls_certified_key **cloned_key_out);
 
 /**
  * "Free" a certified_key previously returned from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
  - (empty)
 
+## 0.4.0 - 2021-04-29
+
+### Added
+
+ - ALPN support for clients (#84)
+ - Enumeration of ciphersuites (#79)
+
 ## 0.4.0 - 2021-03-18
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crustls"
-version = "0.3.0"
+version = "0.5.0"
 authors = ["Jacob Hoffman-Andrews <github@hoffman-andrews.com>"]
 description = "C-to-rustls bindings"
 edition = "2018"


djc:
> This makes OCSP response provisioning more dynamic than the rustls API seems to foresee.

Can you explain this in more detail? I think the relevant rustls APIs have changed a bit on main, as well.

icing:
> > This makes OCSP response provisioning more dynamic than the rustls API seems to foresee.
> 
> Can you explain this in more detail? I think the relevant rustls APIs have changed a bit on main, as well.

Just looking at rustls/main and seeing `ActiveCertifiedKey`: ok, that untangles the fixed CertifiedKey from the more dynamic OCSP data, nice. 

I did not see an interface yet to add the OCSP response data in `main` yet. Did I miss this or is it yet to be defined?

djc:
Hmm, I think I was confused about some other change we made, probably on the client side. Anyway, if you feel the rustls API is lacking, would be good to file an issue against it laying out the case for what use cases it should address (and potentially, how).

src/crustls.h
```diff
@@ -304,6 +304,21 @@ typedef enum rustls_result (*rustls_session_store_get_callback)(rustls_session_s
...
+ */
+typedef void *rustls_ocsp_userdata;
+
+/**
+ * Prototype of a callback that retrieves OCSP response data (DER format)
+ * for the given `certified_key`. The OCSP data has to be copied into the
+ * provided `buf`, if it's length is sufficient. The number of copied bytes
+ * need to be returned in `out_n`.
+ * If `buf` is not of sufficient size, or if not OCSP data is available,
+ * `out_n` must be set to 0.
```
tgeoghegan, line 318:
```suggestion
 * Prototype of a callback that retrieves OCSP response data (DER format)
 * for the given `certified_key`. The OCSP data has to be copied into the
 * provided `buf`, if its length is sufficient. The number of copied bytes
 * need to be returned in `out_n`.
 * If `buf` is not of sufficient size, or if no OCSP data is available,
 * `out_n` must be set to 0.
```
What happens if the client sets `out_n` to 0? Does the function that the callback was passed to return an error?

src/server.rs
```diff
@@ -569,12 +575,14 @@ pub extern "C" fn rustls_server_session_get_negotiated_ciphersuite(
...
     choices: Vec<Arc<CertifiedKey>>,
+    ocsp: OcspProvider,
 }
 
 impl ResolvesServerCertFromChoices {
-    pub fn new(choices: &[Arc<CertifiedKey>]) -> Self {
+    pub fn new(choices: &[Arc<CertifiedKey>], ocsp: OcspProvider) -> Self {
         ResolvesServerCertFromChoices {
             choices: Vec::from(choices),
+            ocsp: ocsp,
```
tgeoghegan, line 585:
```suggestion
            ocsp,
```
Since the variable name and the struct field match, you can use this form. I think `cargo clippy` will complain if you don't.

src/crustls.h
```diff
@@ -304,6 +304,21 @@ typedef enum rustls_result (*rustls_session_store_get_callback)(rustls_session_s
  */
 typedef enum rustls_result (*rustls_session_store_put_callback)(rustls_session_store_userdata userdata, const struct rustls_slice_bytes *key, const struct rustls_slice_bytes *val);
 
+/**
+ * Any context information the callback will receive when invoked.
+ */
+typedef void *rustls_ocsp_userdata;
```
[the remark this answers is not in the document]
jsha: Yeah, I think coalescing these into `rustls_userdata` or `rustls_callback_userdata` is a good idea.

jsha:
> a callback is necessary, since OCSP response data has a max lifetime, commonly of a few days. Since server restarts are independent of this, the timeout of a known response might happen at any time. Also, when starting a server, no (valid) response might be readily available.

This sounds like your intent is that this callback might make a last-minute fetch to the OCSP URL in the middle of a handshake. But I thought you were integrating this with mod_md, which does scheduled fetches of OCSP data, and should never trigger a last-minute fetch like that. Also, doing a last-minute fetch would be incompatible with the async I/O model, since the callback has to be synchronous.

If I understand correctly, it seems like the issue is: you build a set of `rustls_certified_key` rarely (on reloads), and then reuse those for each connection. But the OCSP data changes more frequently than the set of `rustls_certified_key` does.

What if we implement `rustls_certified_key_set_ocsp`, which could be used at any time to set the `ocsp` field of an existing certified key? Then, when mod_md does its scheduled fetch of OCSP data, it would find the appropriate certified_key in the list and modify it to have the updated response.

Looking at the mod_tls code, I think this would mean, instead of setting certified keys on the "archetypal" server config builder (https://github.com/abetterinternet/mod_tls/blob/ade999d1ab21775e4b453b47efcdac1e49179096/src/tls_core.c#L119-L121), you would set them "post-handshake" server config builder (https://github.com/abetterinternet/mod_tls/blob/ade999d1ab21775e4b453b47efcdac1e49179096/src/tls_core.c#L684-L686). That way you wouldn't have to remake the "archetypal" server config builder on every OCSP update.

Right now a `*const rustls_certified_key` points to an `Arc<CertifiedKey>`. To allow setting the OCSP data post-construction, we'd need to turn that into an `Arc<Mutex<CertifiedKey>>`, but I think that's doable. One potential issue: we'd be introducing a call that might block (if some other thread was simultaneously trying to modify the same CertifiedKey). That might cause issues if we block a thread in an async I/O system. But we can document that, and so long as the calling code avoids modifying the same CertifiedKey from two places at once I think it should be fine.

icing:
> > a callback is necessary, since OCSP response data has a max lifetime, commonly of a few days. Since server restarts are independent of this, the timeout of a known response might happen at any time. Also, when starting a server, no (valid) response might be readily available.
> 
> This sounds like your intent is that this callback might make a last-minute fetch to the OCSP URL in the middle of a handshake. But I thought you were integrating this with mod_md, which does scheduled fetches of OCSP data, and should never trigger a last-minute fetch like that. Also, doing a last-minute fetch would be incompatible with the async I/O model, since the callback has to be synchronous.

I am indeed integrating with mod_md, and that might or might not have an OCSP response available, and the data might change on any new connection. The way this data is obtained and refreshed is opaque to the SSL module asking for it. mod_md will not perform a retrieval when data is missing from inside the lookup. This will always be done in another thread.

> If I understand correctly, it seems like the issue is: you build a set of `rustls_certified_key` rarely (on reloads), and then reuse those for each connection. But the OCSP data changes more frequently than the set of `rustls_certified_key` does.
> 
> What if we implement `rustls_certified_key_set_ocsp`, which could be used at any time to set the `ocsp` field of an existing certified key? Then, when mod_md does its scheduled fetch of OCSP data, it would find the appropriate certified_key in the list and modify it to have the updated response.

The design in Apache is based on the design pattern that OpenSSL has, namely a callback to provide the OCSP response data. This callback becomes part of the Apache server core where modules like mod_md can register to provide such data.

There is not event like `ocsp has been updated` that could be used to set the data into a certified key instance. And I think mixing the static certificate/key data with some mutable data is undesired. Note that `ResolvesServerCert` needs to clone the certified key anyway and also that (at least in 0.19) the OCSP data from the certified key is `taken` by the session.

I was a bit confused at first reading this code, but it seems to change in `main` of rustls now.

> Looking at the mod_tls code, I think this would mean, instead of setting certified keys on the "archetypal" server config builder (https://github.com/abetterinternet/mod_tls/blob/ade999d1ab21775e4b453b47efcdac1e49179096/src/tls_core.c#L119-L121), you would set them "post-handshake" server config builder (https://github.com/abetterinternet/mod_tls/blob/ade999d1ab21775e4b453b47efcdac1e49179096/src/tls_core.c#L684-L686). That way you wouldn't have to remake the "archetypal" server config builder on every OCSP update.

There is no "OCSP update" from the point of view of mod_tls. There is only the "current OCSP data" which might be the same as before or different. But the "archetypal" server config will work much better when the session user data for the callback is implemented in PR #86.

> Right now a `*const rustls_certified_key` points to an `Arc<CertifiedKey>`. To allow setting the OCSP data post-construction, we'd need to turn that into an `Arc<Mutex<CertifiedKey>>`, but I think that's doable. One potential issue: we'd be introducing a call that might block (if some other thread was simultaneously trying to modify the same CertifiedKey). That might cause issues if we block a thread in an async I/O system. But we can document that, and so long as the calling code avoids modifying the same CertifiedKey from two places at once I think it should be fine.

I think there is not point in maintaining global, mutable instances for this. The OCSP response data in rustls is owned by the `ServerSession` and thus it needs a copy anyway (my understanding of the rustls code in 0.19). mod_md has the data and mod_tls can copy it from there. No gain to keep another copy at the global certified keys.

jsha:
> And I think mixing the static certificate/key data with some mutable data has no benefits.

Yes, now that you explain the above, I agree.

> Note that ResolvesServerCert needs to clone the certified key anyway and also that (at least in 0.19) the OCSP data from the certified key is taken by the session.

Since we need to clone this anyhow, what do you think about having Apache call its own callback as part of creating the "post-handshake" server session? It would then have an OCSP response in hand, and could make a `rustls_certificate_key` from the previously-loaded cert and key, plus the OCSP response, and make that part of the config for the server session.

The reason I'm resistant to the callback approach is that in general callbacks add complexity, and any time we reach for a callback, I want to first see if we can do it in a non-callback way. Another reason is that callbacks are how we model user-implementable traits from rustls, and it's nice to maintain that correspondence - that any time you see a callback in crustls, it generally corresponds to a trait. And if there's some feature that can only be implemented as a callback, that suggests it should maybe also be a trait in rustls (as you suggested at the top of this PR).

icing:
> > Note that ResolvesServerCert needs to clone the certified key anyway and also that (at least in 0.19) the OCSP data from the certified key is taken by the session.
> 
> Since we need to clone this anyhow, what do you think about having Apache call its own callback as part of creating the "post-handshake" server session? It would then have an OCSP response in hand, and could make a `rustls_certificate_key` from the previously-loaded cert and key, plus the OCSP response, and make that part of the config for the server session.
> 

That would require other changes as well, I think:
 * ResolvesServerCert gives a certified key that has OCSP data or not
 * crustls resolver chooses the first compatible certificate from a list
 * this is currently not doable from outside crustls
 * we do not want to pull the OCSP data for all possible certified key, only the selected one
 * when the session is created, the handshake has not happened yet (current rustls), so this all is not known. 

If you propose that rustls should have a ProvidesOCSPResponse trait, I fully agree. However, I do not see that as achievable in the timeline of my project, given the experience and growing list of other useful changes in rustls.

I understand your resistance to invent 'things' in crustls that have no corresponding concept in rustls. To move all this ahead then, I will change this to a suboptimal implementation. This will fetch OCSP for all possible certs, creates new certified keys (clone) for the session, so OCSP is "just there". 

In the hope of ripping this out again, once the "real" rustls is ready and the corresponding crustls has been made. But that will then be outside of this project, most likely.

djc:
From a rustls perspective, I'm still a little confused about the underlying issues here. On current rustls main, `ResolvesServerCert::resolve()` returns an `Option<Arc<sign::CertifiedKey>>`. `CertifiedKey` then contains the `ocsp` data. It seems to me this gives implementers of `ResolvesServerCert` a lot of freedom in how/when `CertifiedKey`s are constructed and/or cached.

(Maybe we should take the time to update crustls to current rustls soon?)

icing:
> From a rustls perspective, I'm still a little confused about the underlying issues here. On current rustls main, `ResolvesServerCert::resolve()` returns an `Option<Arc<sign::CertifiedKey>>`. `CertifiedKey` then contains the `ocsp` data. It seems to me this gives implementers of `ResolvesServerCert` a lot of freedom in how/when `CertifiedKey`s are constructed and/or cached.

We have currently the key selection for a `ClientHello` hidden in crustls:

```
impl ResolvesServerCert for ResolvesServerCertFromChoices {
    fn resolve(&self, client_hello: ClientHello) -> Option<CertifiedKey> {
        for key in self.choices.iter() {
            if key.key.choose_scheme(client_hello.sigschemes()).is_some() {
                return Some(key.as_ref().clone());
            }
        }
        None
    }
}
```

so the application provides a list of keys and remains ignorant on how the "best" one is selected. I added a `OcspProvider` there to add the OCSP data on the selected key via a callback to the application. (I think the application should be ignorant on this, as the selection depends on understanding of TLS protocol details that should not be coded into the application.)

With this being rejected, there are 2 ways forward that I see:

1. the application provides all keys with current OCSP data when creating the `ServerConfig` and that will be another reason to have the config re-created per connection.
2. To expose a "is-this-key-compatible-with-this-client-hello" function in the crustls API, so that the application can select a workable key and get the OCSP data just for this.


> (Maybe we should take the time to update crustls to current rustls soon?)

All for it, to shorten the cycles to use and improve.

djc:
> To expose a "is-this-key-compatible-with-this-client-hello" function in the crustls API, so that the application can select a workable key and get the OCSP data just for this.

This seems like a reasonable path forward.

icing:
Mr @jsha, if we add something like:

```
pub extern "C" fn rustls_certified_key_select_for_client_hello(
    certified_keys: *const *const rustls_certified_key,
    certified_keys_len: size_t,
    hello: *const rustls_client_hello
) -> *const rustls_certified_key {

```
we would need to get the "real" `ClientHello` from our `rusts_client_hello`. There are several ways to achieve that:

1. add an opaque `const void*` to the `rusts_client_hello` that is the `ClientHello` instance.
2. add an opaque `const void*` as additional arg to the `rustls_client_hello_callback`
3. use a thread local ***ducks***
4. ...

Feedback appreciated! Willingness to implement here...

jsha:
How about this?

```
/// Return the first rustls_certified_key whose signature scheme is present in the
/// provided slice of signature schemes. Returns NULL if there is no match.
/// Since this potentially iterates over the entire input list, the input list should
/// be as short as possible. For instance, there's no reason to include two
/// rustls_certified_keys that have the same signature scheme.
pub extern "C" fn rustls_certified_key_find_by_signature_schemes(
    certified_keys: *const *const rustls_certified_key,
    certified_keys_len: size_t,
    signature_schemes: *const rustls_slice_u16,
) -> *const rustls_certified_key {
```

I prefer this to passing a whole `*rustls_client_hello`, because it's more clear what it's operating on. That is, we're not doing any matching of ALPN or SNI in this function (because that would be inefficient - for SNI matching we'd want to use a hash map).

src/cipher.rs
```diff
@@ -108,6 +110,45 @@ pub extern "C" fn rustls_certified_key_build(
     }
 }
 
+/// Create a copy of the rustls_certified_key with the given OCSP response data
+/// as DER encoded bytes. The OCSP response may be given as NULL to clear any
+/// possibly present OCSP data from the cloned key.
+/// The cloned key is independant from its original and needs to be freed
```
jsha, line 116:
```suggestion
/// The cloned key is independent from its original and needs to be freed
```

src/cipher.rs
```diff
@@ -108,6 +110,45 @@ pub extern "C" fn rustls_certified_key_build(
...
+                None => return NullParameter,
+            }
+        };
+        let certified_key: Arc<CertifiedKey> = unsafe {
+            match (key as *const CertifiedKey).as_ref() {
+                Some(c) => arc_with_incref_from_raw(c),
+                None => return NullParameter,
+            }
+        };
+        let mut nkey = certified_key.deref().clone();
```
jsha, line 137:
```suggestion
        let mut new_key = certified_key.deref().clone();
```

Let's use a slightly longer name for clarity - at first I thought the `n` in `nkey` was for length.

src/cipher.rs
```diff
@@ -108,6 +110,45 @@ pub extern "C" fn rustls_certified_key_build(
...
+        if !ocsp_response.is_null() {
+            unsafe {
+                nkey.ocsp = Some(Vec::from(slice::from_raw_parts((*ocsp_response).data, (*ocsp_response).len)));
```
jsha, line 140: It's possible `ocsp_response` could be non-null, but `(*ocsp_response).data` could be null. So we should use `try_slice!(*ocsp_response.data, *ocsp_response.len)`.

src/cipher.rs
```diff
@@ -108,6 +110,45 @@ pub extern "C" fn rustls_certified_key_build(
...
+            }
+        }
+        else {
+            nkey.ocsp = None;
+        }
+        let ncertified_key = Arc::into_raw(Arc::new(nkey)) as *const _;
+        *cloned_key_out = ncertified_key;
```
jsha, line 147: Let's make these one line:

```suggestion
        *cloned_key_out = Arc::into_raw(Arc::new(nkey)) as *const _;
```

icing:
We can do `rustls_certified_key_find_by_signature_schemes()`, but I'd rather not.

Knowledge of how the internals of TLS work (in this and future versions) should not be part of the application. If, for example, it becomes desirable for TLS stacks to also inspect ClientHello extensions to chose the best certificate, then crustls would need to do a breaking change in its API.

The application is interested in the service of "chose the first suited one among these certificates" from the TLS stack. The TLS stack should then make a design that allows it to implement that with future protocol (tweaks) in mind.

This flies a bit in the face of the "minimal design" approach where you give a function only those parameters it really needs. Which makes it better testable, but also requires breaking changes on each evolving detail. This is not a problem in an application internal API, but results in a deployment/upgrade problem on public APIs.

In a TLS API that should remain stable over years, I would always pass the `ServerSession` around to everyone involved in that session. So that the context is clear and accessible. Of course, that means functions have more context and are not as "functional" any more, but the gained agility without breaking changes has benefits.

For consideration.

jsha (changes requested):
The current revision of this PR (adding `rustls_certified_key_clone_with_ocsp`) looks good to me with minor changes. Let's get that revised, land this, and then add on `rustls_certified_key_select_for_client_hello` as a subsequent PR.

@icing Your comment about not exposing the detail of signature schemes makes sense to me. My main worry is that people will treat this function as "select based on SNI" (since SNI is part of the ClientHello). At any rate, I'm satisfied with the approach you propose, passing the full `rustls_client_hello` to this function - we can make sure to document well the fact that it does not do anything with SNI or ALPN.

src/cipher.rs
icing: I am very independent with my spelling.

src/cipher.rs
icing: Needed one intermediate var so that the try_slice was not complained about.

icing:
PR updated.

> The current revision of this PR (adding `rustls_certified_key_clone_with_ocsp`) looks good to me with minor changes. Let's get that revised, land this, and then add on `rustls_certified_key_select_for_client_hello` as a subsequent PR.
> 

Yes. Lets do that on top of the thread local PR by you. Then I can migrate hopefully the `SessionConfig`s back into the initial server setup.

jsha (approved):

PR #86, opened 2021-04-27
Make userdata a per-session config

### LINKED ISSUE #81: application session data

jsha:
Use thread-local storage to make the userdata available to callbacks. This relies on the fact that rustls always does work on the thread it was called from (https://github.com/abetterinternet/mod_tls/issues/2#issuecomment-800495123).

The per-session userdata pointer is set with `rustls_*_session_set_userdata`. In process_new_packets, we push the current session's userdata pointer onto a thread-local stack, and build a guard object that will later pop it off. In our traits that call out to callbacks, we read the most recent value from the stack, and pass that as userdata. The stack makes it possible for callbacks to invoke process_new_packets themselves (for instance, if a crustls server is working with a client in a callback).

First commit is the boring renaming stuff (thanks for the suggestion @djc), subsequent commits are more interesting. @icing can you give this a try and see if it meets your needs?

Follow-up to #83
Fixes #81

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,7 +1,7 @@
 use libc::{c_char, c_void, size_t};
-use std::{convert::TryInto, ptr::null_mut};
 use std::io::{BufReader, Cursor, Read, Write};
 use std::slice;
+use std::{convert::TryInto, ptr::null_mut};
 use std::{ffi::CStr, sync::Arc};
 use std::{ffi::OsStr, fs::File};
 use webpki::DNSNameRef;
@@ -13,13 +13,13 @@ use rustls::{
 use crate::error::{self, map_error, result_to_tlserror, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_slice_slice_bytes, rustls_str};
 use crate::session::{
-    rustls_session_store_get_callback, rustls_session_store_put_callback,
-    SessionStoreBroker, SessionStoreGetCallback, SessionStorePutCallback,
+    rustls_session_store_get_callback, rustls_session_store_put_callback, SessionStoreBroker,
+    SessionStoreGetCallback, SessionStorePutCallback,
 };
 use crate::{
     arc_with_incref_from_raw, ffi_panic_boundary, is_close_notify, rslice::NulByte,
-    try_mut_from_ptr, try_mut_slice, try_ref_from_ptr, try_slice, CastPtr,
-    userdata_push, userdata_get,
+    try_mut_from_ptr, try_mut_slice, try_ref_from_ptr, try_slice, userdata_get, userdata_push,
+    CastPtr,
 };
 use rustls_result::NullParameter;
 
@@ -386,8 +386,8 @@ pub extern "C" fn rustls_client_session_set_userdata(
     session: *mut rustls_client_session,
     userdata: *mut c_void,
 ) {
-  let session: &mut Sess = try_mut_from_ptr!(session);
-  session.userdata = userdata;
+    let session: &mut Sess = try_mut_from_ptr!(session);
+    session.userdata = userdata;
 }
 
 #[no_mangle]
--- a/src/client.rs
+++ b/src/client.rs
@@ -204,10 +204,11 @@ impl rustls::ServerCertVerifier for Verifier {
 
 /// Set a custom server certificate verifier.
 ///
-/// The userdata pointer must stay valid until (a) all sessions created with this
-/// config have been freed, and (b) the config itself has been freed.
 /// The callback must not capture any of the pointers in its
 /// rustls_verify_server_cert_params.
+/// If `userdata` has been set with rustls_server_session_set_userdata, it
+/// will be passed to the callback. Otherwise the userdata param passed to
+/// the callback will be NULL.
 ///
 /// The callback must be safe to call on any thread at any time, including
 /// multiple concurrent calls. So, for instance, if the callback mutates
@@ -587,9 +588,9 @@ pub extern "C" fn rustls_client_session_write_tls(
 /// keys and values are highly sensitive data, containing enough information
 /// to break the security of the sessions involved.
 ///
-/// `userdata` must live as long as the config object and any sessions
-/// or other config created from that config object.
-///
+/// If `userdata` has been set with rustls_server_session_set_userdata, it
+/// will be passed to the callbacks. Otherwise the userdata param passed to
+/// the callbacks will be NULL.
 #[no_mangle]
 pub extern "C" fn rustls_client_config_builder_set_persistence(
     builder: *mut rustls_client_config_builder,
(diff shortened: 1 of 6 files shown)


icing:
I like in this approach that the application needs not be aware of the implementation details. The API can be kept, even if the implementation of this feature should for some reason change.

src/client.rs
```diff
@@ -59,8 +59,13 @@ pub struct rustls_client_session {
     _private: [u8; 0],
 }
 
+pub struct Sess {
```
tgeoghegan, line 62: This type name is a little terse, I don't think it needs to be exported from the crate, and seems virtually identical to `server::Sess` so WDYT about having this in `session.rs`:
```
struct Session<S: rustls::Session> {
    session: S,
    userdata: *mut c_void,
}
```
[`rustls::ClientSession`](https://docs.rs/rustls/0.19.1/rustls/struct.ClientSession.html#impl-Session) and [`rustls::ServerSession`](https://docs.rs/rustls/0.19.1/rustls/struct.ServerSession.html#impl-Session) both implement [`rustls::Session`](https://docs.rs/rustls/0.19.1/rustls/trait.Session.html)

tgeoghegan (approved):
Looks right to me; it's up to you if you want to use my (perhaps overcomplicated) suggestion of `crustls::Session<S: rustls::Session>`. And you asked me to remind you that you "should probably ask for documentation of which function calls can possibly lead to traits being invoked."

src/client.rs
jsha: Unfortunately then we wouldn't be able to call the methods on each that are specific to `Client` or `Server`, like `ServerSession::get_sni_hostname`.

Making this pub(crate) instead of pub is a good idea though.

icing:
If I read this correct, in case of a `panic!` the unwind of the stack would `drop` a guard in place and that would pop the stack as needed?

src/lib.rs
```diff
@@ -35,39 +34,210 @@ thread_local! {
...
+    /// rustls_result::Panic.
+    fn try_drop(mut self) -> Result<(), UserdataError> {
+        self.try_pop()
+    }
+
+    fn try_pop(&mut self) -> Result<(), UserdataError> {
+        let expected_data = match self.data {
+            Some(d) => d,
+            None => return Err(UserdataError::AlreadyPopped),
+        };
```
djc, line 66: Maybe use `ok_or()` here?

src/lib.rs
```diff
@@ -35,39 +34,210 @@ thread_local! {
...
+        eprintln!("popping");
         USERDATA
             .try_with(|userdata| {
-                if let Ok(mut v) = userdata.try_borrow_mut() {
-                    v.pop();
-                }
+                userdata.try_borrow_mut().map_or_else(
+                    |_| Err(UserdataError::AlreadyBorrowed),
+                    |mut v| match v.pop() {
```
djc, line 72: And here... If you combine it with `?`, you can reduce the rightward drift a fair bit.

src/server.rs
```diff
@@ -373,10 +373,17 @@ pub extern "C" fn rustls_server_session_process_new_packets(
 ) -> rustls_result {
     ffi_panic_boundary! {
         let session: &mut Sess = try_mut_from_ptr!(session);
-        let _guard = userdata_push(session.userdata);
-        match session.session.process_new_packets() {
+        let guard = match userdata_push(session.userdata) {
```
djc, line 376: The `and_then()` combinator would make this a bit less repetitive, I think.

src/session.rs
```diff
@@ -89,6 +89,10 @@ impl SessionStoreBroker {
 
     fn retrieve(&self, key: &[u8], remove: bool) -> Option<Vec<u8>> {
         let key: rustls_slice_bytes = key.into();
+        let userdata = match userdata_get() {
```
djc, line 92: You could use `.ok()?` here instead of the full match statement.

djc (commented):
Some style nits.

Also: wouldn't you normally spell "user data" as two words? In which case `user_data` and `UserData` would be less confusing than the (perhaps more C-like) `userdata`.

src/lib.rs
```diff
@@ -17,6 +18,228 @@ mod session;
...
+    fn try_drop(mut self) -> Result<(), UserdataError> {
+        self.try_pop()
+    }
+
+    fn try_pop(&mut self) -> Result<(), UserdataError> {
+        let expected_data = match self.data {
+            Some(d) => d,
+            None => return Err(UserdataError::AlreadyPopped),
+        };
+        eprintln!("popping");
```
tgeoghegan, line 67: I don't think libraries should clutter client's stderr with debug output by default. Rich logging options are more API surface than `crustls` needs but could this be restricted to debug builds?

src/lib.rs
```diff
@@ -17,6 +18,228 @@ mod session;
...
+                                Ok(())
+                            } else {
+                                Err(UserdataError::WrongData)
+                            }
+                        }
+                        None => Err(UserdataError::EmptyStack),
+                    },
+                )
+            })
+            .unwrap_or_else(|_: AccessError| Err(UserdataError::AccessError))
```
tgeoghegan, line 85: nit/curiosity: is the type annotation on the discarded closure argument necessary?

src/lib.rs
```diff
@@ -17,6 +18,228 @@ mod session;
...
+    }
+}
+
+impl Drop for UserdataGuard {
+    fn drop(&mut self) {
+        self.try_pop().ok();
+    }
+}
+
+#[derive(Debug)]
```
tgeoghegan, line 95:
```suggestion
#[derive(Clone, Debug)]
```
Just because I think everything should be `Clone` if possible.

src/lib.rs
jsha: Oops, leftover debugging output. Thanks for catching.

src/lib.rs
jsha: Not necessary; I added it as typesafe documentation, showing that the returned type is `thread::AccessError` so it is correct to return `UserdataError::AccessError`.

src/server.rs
jsha: I couldn't quite figure out an elegant way to do this with `and_then()` but if you'd like to show an example of the direction you're thinking in, I'm happy to try it. Thanks for all the tips on combinators, they are nice improvements.

src/lib.rs
```diff
@@ -17,6 +18,222 @@ mod session;
...
+
+    #[test]
+    fn userdata_multi_threads() {
+        let hello = "hello";
+        let hello_ptr: *mut c_void = hello as *const _ as _;
+        let guard = userdata_push(hello_ptr).unwrap();
+        assert_eq!(userdata_get().unwrap(), hello_ptr);
+
+        let thread1 = thread::spawn(|| {
+            let yo = "hello";
```
tgeoghegan, line 214:
```suggestion
            let yo = "yo";
```

tgeoghegan (approved):

PR #88, opened 2021-04-29
Upgrade rustls to 0.20.0

djc:
Open issues before this compiles:

* [x] `ClientConfig` now needs `root_store`, `ct_logs` and `cipher_suite` at initialization time, but rustls itself does not have `ClientConfigBuilder` yet.
* [x] webpki errors now have an `op` field, do we just drop that or combinatorially explode our errors?

To do after this compiles:

* [ ] Do something useful with the `IoState` return value from `Connection::process_new_packets()`
* [x] Rename crustls types to match renames in rustls (notably, `Session` -> `Connection`)
* [x] Review if I/O error handling is still correct


jsha:
Awesome, thanks for starting work on this! I'll try to review it soon.

One thing I want to figure out is branch structure and releases. I'd like to only release versions of crustls that depend on released versions of rustls. And I'd like to have a branch always ready to release patches on top of the current released version. I'm thinking:

 - `main`: based on a released rustls, can be built into a release at any time.
 - `next`: based on a specific rustls git revision, which we update frequently. Once a new rustls is released, we finish the work in `next` to depend on that released version, then merge it to main and cut a new `crustls` release.
 
 Thoughts?

djc:
Hmm, I think the branching model would be easier to understand if you just mirror rustls branches. So crustls main mirrors rustls main (but maybe not quite always up to date), and then have a 0.19.x branch for rustls 0.19.x.

jsha:
I pushed some updates getting us slightly closer to the goal, mainly with renames etc.

> webpki errors now have an op field, do we just drop that or combinatorially explode our errors?

I think we should combinatorially explode for now. What I'd really like is to look through which pairs of (webpki::Error, op) can actually occur, and only combinatorially explode those, leaving anything else that shows up as a catch-all "unknown" error.

jsha:
> I think we should combinatorially explode for now. 

I actually wound up deciding the other way on this, and flattening out WebPkiOp.

jsha:
This is getting fairly close. For the server config, I opted to preserve more or less the same API we currently have, rather than trying to imitate the new Builder hierarchy in rustls. We may want to do that later on, but it will be a fairly intrusive change.

For the client config, it's almost but not quite possible to do that, mainly before ClientConfig.verifier is not public, so the only way we can set it is by passing a verifier to the right builder object. #758

djc:
@jsha I think you're back soon, in which case, welcome back!

I noticed that you added a `src/config.rs` module in this PR but did not commit it. Rustls 0.20 was just released yesterday, so it's time to wrap up this PR. I started working on it but then ran into this issue, which is unfortunate. I'll see if I can continue with the client side first.

jsha:
Thanks for the welcome back, @djc! I looked in my working directory for an uncommitted `src/config.rs` and found none. I suspect that I added the `mod config;` line at a time when I expected to create `src/config.rs`, and later changed my mind and failed to delete the `mod config`. It seems like the remaining compile errors are just some things that remain to be fixed, as opposed to things that are supposed to be in `config.rs` but are missing.

I pushed a small commit removing `mod config`.

djc:
@jsha I think your previous approach to the server config builder no longer works. I think we should probably try to mimick the Rust API as much as possible, do you think we should just allocate a new `Box` at each state transition or is there something smarter we could do? (For example, have an enum with all the potential client states, box that up, then transition it through the states -- but since the Rust methods take ownership, I'm not sure if we can make that work without a bunch of unsafe stuff...)

jsha:
Agreed the existing configbuilder approach no longer works, and agreed on generally wanting to mimic the Rust API as closely as possible. One problem is that the type names and function names involved are extremely verbose, and in C we don't get to elide them with method chaining as we do in Rust.

What about defining a single struct that the caller can pass with three fields for the cipher suites, kx groups, and protocol versions? Then on the Rust side we could instantiate the builder and proceed through its typestates without ever exposing them to C code. I think this would require some finagling to achieve the "linker friendly" property of not including unused suites, but I think we can do it. We could provide a `rustls_safe_default_cipher_suites()` (etc) function that would return the defaults, so they would only be referenced if the caller called that function.

src/server.rs
```diff
@@ -65,15 +91,69 @@ impl CastPtr for rustls_server_config {
...
     ffi_panic_boundary! {
         let builder = rustls::ServerConfig::builder().with_safe_defaults();
-        let config = builder.with_no_client_auth().with_cert_resolver(Arc::new(rustls::server::ResolvesServerCertUsingSni::new()));
-        let b = Box::new(config);
-        Box::into_raw(b) as *mut _
+        BoxCastPtr::to_mut_ptr(builder)
+    }
+}
+
+/// Create a rustls_server_config_builder. Caller owns the memory and must
```
jsha, line 105: "Create a rustls_server_config_builder_wants_verifier. ..."

src/server.rs
```diff
@@ -65,15 +91,69 @@ impl CastPtr for rustls_server_config {
...
+/// when negotiating a TLS session.
+///
+/// `tls_version` is the version of the protocol, as defined in rfc8446,
+/// ch. 4.2.1 and end of ch. 5.1. Some values are defined in
+/// `rustls_tls_version` for convenience.
+///
+/// `versions` will only be used during the call and the application retains
+/// ownership. `len` is the number of consecutive `ui16` pointed to by `versions`.
+#[no_mangle]
+pub extern "C" fn rustls_client_config_builder_new(
```
jsha, line 120: rustls_*server*_config_builder_new

jsha (changes requested):
Nice work. I particularly like the BoxCastPtr abstraction. That should introduce a lot more safety by reducing the places where we have an `as` cast to or from raw pointers. It also helps think a lot more clearly about the semantics of the types we share with C: types that impl BoxCastPtr are at some point owned by C code; those that don't (once we adopt this more broadly) are only passed to C code as references.

djc:
No, this has my latest state already. Thanks for picking it up!

src/server.rs
```diff
@@ -513,14 +532,19 @@ impl ResolvesServerCert for ClientHelloResolver {
             .map(|s| s.get_u16())
             .collect();
         // Unwrap the Option. None becomes an empty slice.
-        let alpn: &[&[u8]] = client_hello.alpn().unwrap_or(&[]);
-        let alpn = rustls_slice_slice_bytes { inner: alpn };
+        let alpn = match client_hello.alpn() {
+            Some(iter) => iter.collect(),
+            None => vec![],
+        };
```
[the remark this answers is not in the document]
jsha: Tried reverting these lines to the previous version and now I see the problem: `client_hello.alpn()` now returns an Iterator, so we have to collect it. 👍🏻

jsha:
Have a great vacation!

jsha (approved):

PR #93, opened 2021-05-11
Icing/cert select

icing:
This is my attempt at adding an opaque internal to `rustls_client_hello` so that the hello can be used to call crustls (from the callback), such as the new `rustls_server_session_select_certified_key `.

I have tried several ways to map `ClientHello` directly to an opaque pointer, but was unable to overcome rust's checks.


jsha:
Thanks! I made some tweaks in #94 and merged that instead, so closing this.

PR #94, opened 2021-05-12
Add support for selecting a certified key from a client hello

jsha:
Builds on @icing's #93.

I got rid of the `internals` field in favor of turning the `signature_schemes` field back into a `Vec`.


icing:
=)

You also got rid of the future possibility to use the `ClientHello` instance in decision making. But I assume it can be smuggled back in when needed by extending to `rustls_client_hello` an a non-ABI breaking way.

PR #95, opened 2021-05-13
Change read_tls and write_tls to use a callback

### LINKED ISSUE #92: Change rustls_*_session_write_tls to take callback
Right now, `rustls_*_session_write_tls` takes a buffer, and moves bytes from the session's internal buffer into the user provided one, for the user to send on to the network. As described in https://github.com/curl/curl/issues/7007, this poses a problem: now the user code has to keep track of the bytes it received, in case they can't all be written to the network right away.

It would be simpler (and more efficient) to borrow bytes from rustls' internal buffer, write as many as we can, and tell rustls how many we consumed. In fact, [that's how rustls works in the Rust world](https://docs.rs/rustls/0.19.1/rustls/trait.Session.html#tymethod.write_tls). `write_tls` takes a `&dyn mut Write`. The Write trait receives a borrowed buffer and returns the number of bytes it accepted.

The closest thing to traits in C is callbacks, and we have some well-established patterns for using callbacks in crustls to implement traits. We should do this with `rustls_*_session_write_tls`: take a callback and turn it into an `&dyn mut Write` implementation.

This would be a breaking API change. I don't think it's worth keeping around a function that offers the old copy-to-buffer functionality because it's easy to make mistakes with and hard to use correctly.

jsha:
This makes the user code a lot simpler. In particular it allows rustls to expose borrowed memory via `Read` or `Write`, and the user can consume (or fill) as much of that memory as appropriate. This is particularly useful for `Write`, because it means if the user code receives a buf of 1000 bytes but is only able to write 700 bytes right now, it doesn't need to find someplace to stash those remaining 300 bytes: they can just stay in rustls' internal buffers.

Fixes #92.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -14,7 +14,7 @@ use rustls::{
 use webpki::DNSNameRef;
 
 use crate::connection;
-use crate::error::{self, map_error, result_to_tlserror, rustls_io_error, rustls_result};
+use crate::error::{self, map_error, result_to_tlserror, rustls_io_result, rustls_result};
 use crate::io::{rustls_read_callback, rustls_write_callback, ReadCallback, WriteCallback};
 use crate::rslice::{rustls_slice_bytes, rustls_slice_slice_bytes, rustls_str};
 use crate::session::{
@@ -653,7 +653,7 @@ pub extern "C" fn rustls_client_session_read_tls(
     callback: rustls_read_callback,
     userdata: *mut c_void,
     out_n: *mut size_t,
-) -> rustls_io_error {
+) -> rustls_io_result {
     ffi_panic_boundary! {
         let session: &mut Sess = try_mut_from_ptr!(session);
         let out_n: &mut size_t = try_mut_from_ptr!(out_n);
@@ -679,7 +679,7 @@ pub extern "C" fn rustls_client_session_write_tls(
     callback: rustls_write_callback,
     userdata: *mut c_void,
     out_n: *mut size_t,
-) -> rustls_io_error {
+) -> rustls_io_result {
     ffi_panic_boundary! {
         let session: &mut Sess = try_mut_from_ptr!(session);
         let out_n: &mut size_t = try_mut_from_ptr!(out_n);
--- a/src/client.rs
+++ b/src/client.rs
@@ -641,6 +641,9 @@ pub extern "C" fn rustls_client_session_read(
 /// I/O is performed by `callback`, which you provide. Rustls will invoke your
 /// callback with a suitable buffer to store the read bytes into. You don't have
 /// to fill it up, just fill with as many bytes as you get in one syscall.
+/// The `userdata` parameter is passed through directly to `callback`. Note that
+/// this is distinct from the `userdata` parameter set with
+/// `rustls_client_session_set_userdata`.
 /// Returns 0 for success, or an errno value on error. Passes through return values
 /// from callback. See rustls_read_callback for more details.
 /// https://docs.rs/rustls/0.19.0/rustls/trait.Session.html#tymethod.read_tls
@@ -648,6 +651,7 @@ pub extern "C" fn rustls_client_session_read(
 pub extern "C" fn rustls_client_session_read_tls(
     session: *mut rustls_client_session,
     callback: rustls_read_callback,
+    userdata: *mut c_void,
     out_n: *mut size_t,
 ) -> rustls_io_error {
     ffi_panic_boundary! {
@@ -655,7 +659,7 @@ pub extern "C" fn rustls_client_session_read_tls(
         let out_n: &mut size_t = try_mut_from_ptr!(out_n);
         let callback: ReadCallback = try_callback!(callback);
 
-        connection::read_tls(&mut session.session, callback, session.userdata, out_n)
+        connection::read_tls(&mut session.session, callback, userdata, out_n)
     }
 }
 
@@ -663,6 +667,9 @@ pub extern "C" fn rustls_client_session_read_tls(
 /// `callback`, which you provide. Rustls will invoke your callback with a
 /// suitable buffer containing TLS bytes to send. You don't have to write them
 /// all, just as many as you can in one syscall.
+/// The `userdata` parameter is passed through directly to `callback`. Note that
+/// this is distinct from the `userdata` parameter set with
+/// `rustls_client_session_set_userdata`.
 /// Returns 0 for success, or an errno value on error. Passes through return values
 /// from callback. See rustls_write_callback for more details.
 /// https://docs.rs/rustls/0.19.0/rustls/trait.Session.html#tymethod.write_tls
@@ -670,6 +677,7 @@ pub extern "C" fn rustls_client_session_read_tls(
 pub extern "C" fn rustls_client_session_write_tls(
     session: *mut rustls_client_session,
     callback: rustls_write_callback,
+    userdata: *mut c_void,
     out_n: *mut size_t,
 ) -> rustls_io_error {
     ffi_panic_boundary! {
@@ -677,7 +685,7 @@ pub extern "C" fn rustls_client_session_write_tls(
         let out_n: &mut size_t = try_mut_from_ptr!(out_n);
         let callback: WriteCallback = try_callback!(callback);
 
-        connection::write_tls(&mut session.session, callback, session.userdata, out_n)
+        connection::write_tls(&mut session.session, callback, userdata, out_n)
     }
 }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -617,6 +617,12 @@ pub extern "C" fn rustls_client_session_read(
     }
 }
 
+/// Read some TLS bytes from the network into internal buffers. The actual network
+/// I/O is performed by `callback`, which you provide. Rustls will invoke your
+/// callback with a suitable buffer to store the read bytes into. You don't have
+/// to fill it up, just fill with as many bytes as you get in one syscall.
+/// Returns 0 for success, or an errno value on error. Passes through return values
+/// from callback. See rustls_read_callback for more details.
 /// https://docs.rs/rustls/0.19.0/rustls/trait.Session.html#tymethod.read_tls
 #[no_mangle]
 pub extern "C" fn rustls_client_session_read_tls(
@@ -633,6 +639,12 @@ pub extern "C" fn rustls_client_session_read_tls(
     }
 }
 
+/// Write some TLS bytes to the network. The actual network I/O is performed by
+/// `callback`, which you provide. Rustls will invoke your callback with a
+/// suitable buffer containing TLS bytes to send. You don't have to write them
+/// all, just as many as you can in one syscall.
+/// Returns 0 for success, or an errno value on error. Passes through return values
+/// from callback. See rustls_write_callback for more details.
 /// https://docs.rs/rustls/0.19.0/rustls/trait.Session.html#tymethod.write_tls
 #[no_mangle]
 pub extern "C" fn rustls_client_session_write_tls(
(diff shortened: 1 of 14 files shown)


jsha:
one open question: should I also change the plaintext `_read()` and `_write()` methods to use callbacks?

jsha:
Also: this uses the session's userdata set with `rustls_*_session_set_userdata`. Previously, it was optional to set the userdata since you might not really need it. But now, you can't really use a session without setting userdata. Should we make it a mandatory part of the session constructors?

Alternately, should we have a separate userdata parameter that's passed to `_read_tls` / `_write_tls` and passed directly to the callback?

jsha:
Alright, ready for review (although I just realized I need to update the tests).

jsha:
Also particularly interested in reviews from @icing and @djc.

jsha:
> should I also change the plaintext _read() and _write() methods to use callbacks?

Answering my own question: No, because Reader::read and Writer::write don't _take_ a `Read` or a `Write`, they are methods of a `Read` or a `Write` implementation.

src/io.rs
```diff
@@ -0,0 +1,95 @@
...
+pub(crate) struct CallbackReader {
+    pub callback: ReadCallback,
+    pub userdata: *mut c_void,
+}
+
+impl Read for CallbackReader {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let mut out_n: usize = 0;
+        let cb = self.callback;
+        let result = unsafe { cb(self.userdata, buf.as_mut_ptr(), buf.len(), &mut out_n) };
```
tgeoghegan, line 42: I'm curious -- why assign `self.callback` to `cb` here? Why not just call `self.callback(self.userdata, ...)` in the `unsafe{}`?

src/io.rs
```diff
@@ -0,0 +1,95 @@
...
+        let cb = self.callback;
+        let result = unsafe { cb(self.userdata, buf.as_ptr(), buf.len(), &mut out_n) };
+        match result.0 {
+            0 => Ok(out_n),
+            e => Err(Error::from_raw_os_error(e)),
+        }
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        Ok(())
```
tgeoghegan, line 93: Might we want to instruct the client to flush whatever buffers they may be using? I think that might require a separate flush callback which is kind of a hassle but I worry about cases where `[c]rustls` assume that some bytes have gone out over the network because `std::io::Write::flush()` succeeded but they're actually still buffered somewhere.

src/main.c
```diff
@@ -87,20 +87,16 @@ const char *ws_strerror (int err)
...
-    m = must_use_send ? send(fd, buf, n, 0) : write(fd, buf, n);
+    m = write(fd, buf, n);
     if(m < 0) {
-      must_use_send ? perror("writing to socket") : perror("writing to stdout");
+      perror("writing to stdout");
       return 1;
     }
     if(m == 0) {
-      fprintf(stderr, "early EOF when writing to %s\n", must_use_send ? "socket" : "stdout");
+      fprintf(stderr, "early EOF when writing to %s\n", "stdout");
```
tgeoghegan, line 99:
```suggestion
      fprintf(stderr, "early EOF when writing to stdout\n");
```

src/main.c
```diff
@@ -280,6 +233,40 @@ copy_plaintext_to_stdout(struct rustls_client_session *client_session)
   return CRUSTLS_DEMO_ERROR;
 }
 
+struct demo_conn {
+  int fd;
+  const char* verify_arg;
```
tgeoghegan, line 238:
```suggestion
  const char *verify_arg;
```
For consistency.

src/main.c
```diff
@@ -280,6 +233,40 @@ copy_plaintext_to_stdout(struct rustls_client_session *client_session)
...
+
+int read_cb(void *userdata, uint8_t *buf, uintptr_t len, uintptr_t *out_n)
+{
+  ssize_t n = 0;
+  struct demo_conn *conn = (struct demo_conn*)userdata;
+  n = read(conn->fd, buf, len);
+  if(n < 0) {
+    return errno;
+  }
+  *out_n = n;
```
tgeoghegan, line 249: nit
```suggestion
  if (out_n != NULL) {
    *out_n = n;
  }
```

src/main.c
```diff
@@ -504,7 +481,7 @@ verify(void *userdata, const rustls_verify_server_cert_params *params) {
     }
   }
   fprintf(stderr, "ocsp response len: %ld\n", params->ocsp_response.len);
-  if(0 != strcmp((const char *)userdata, "verify_arg")) {
+  if(0 != strcmp(((struct demo_conn*)userdata)->verify_arg, "verify_arg")) {
```
tgeoghegan, line 484: nit: three opening parens in a row suggests you could move the cast to `struct demo_conn *` out of the `if` condition for clarity.

src/io.rs
```diff
@@ -0,0 +1,95 @@
...
+/// be triggered by subsequent calls to one of the `_read_tls` methods.
+/// `userdata` is set to the value provided to `rustls_*_session_set_userdata`. In most
+/// cases that should be a struct that contains, at a minimum, a file descriptor.
+/// The buf and out_n pointers are borrowed and should not be retained across calls.
+pub type rustls_read_callback = Option<
+    unsafe extern "C" fn(
+        userdata: *mut c_void,
+        buf: *mut u8,
+        n: size_t,
+        out_n: *mut usize,
```
tgeoghegan, line 22: Can `out_n` be `*mut size_t`, since it and `n` both represent a count of bytes? I also think the C representation is unfortunate, since the `usize` ends up as `uintptr_t` and to me, `uintptr_t` is a sign that you're doing something suspicious.

src/connection.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use libc::{size_t, EIO};
+use rustls::Session;
+
+use crate::error::rustls_io_error;
+use crate::io::{CallbackReader, CallbackWriter, ReadCallback, WriteCallback};
+
+// Call Session::read_tls, providing an &mut dyn Write implemented with a C callback.
+pub(crate) fn read_tls(
+    session: &mut dyn Session,
```
tgeoghegan, line 11: Could this be `read_tls<S: Session>(session: &mut S, ...)` to avoid dynamic dispatch? It seems like this and `write_tls` will be on the hot path so maybe it's worth specializing for `ClientSession` and `ServerSession`? Perhaps the time spent on dynamic dispatch is irrelevant since you're waiting on IO all the time anyway.

src/client.rs
```diff
@@ -636,64 +637,55 @@ pub extern "C" fn rustls_client_session_read(
...
-/// the ClientSession. This may make packets available to
-/// `rustls_client_session_process_new_packets`, which in turn may make more
-/// bytes available to `rustls_client_session_read`.
-/// On success, store the number of bytes actually read in *out_n (this may
-/// be less than `count`). This function returns success and stores 0 in
-/// *out_n when the input count is 0.
+/// Read some TLS bytes from the network into internal buffers. The actual network
+/// I/O is performed by `callback`, which you provide. Rustls will invoke your
+/// callback with a suitable buffer to store the read bytes into. You don't have
+/// to fill it up, just fill with as many bytes as you get in one syscall.
```
tgeoghegan, line 643:
```suggestion
/// to fill it up, just fill with as many bytes as are available.
```

src/client.rs
```diff
@@ -636,64 +637,55 @@ pub extern "C" fn rustls_client_session_read(
...
 }
 
-/// Read up to `count` TLS bytes from `buf` (usually read from a socket) into
-/// the ClientSession. This may make packets available to
-/// `rustls_client_session_process_new_packets`, which in turn may make more
-/// bytes available to `rustls_client_session_read`.
-/// On success, store the number of bytes actually read in *out_n (this may
-/// be less than `count`). This function returns success and stores 0 in
-/// *out_n when the input count is 0.
+/// Read some TLS bytes from the network into internal buffers. The actual network
```
tgeoghegan, line 640: nit: I think you could avoid explicitly mentioning the network because you're not necessarily reading or writing to/from a network.

src/crustls.h
```diff
@@ -281,6 +281,40 @@ typedef struct rustls_verify_server_cert_params {
 
 typedef enum rustls_result (*rustls_verify_server_cert_callback)(rustls_verify_server_cert_user_data userdata, const struct rustls_verify_server_cert_params *params);
 
+/**
+ * A return value for a function that may return either success (0) or an errno
+ * value.
+ */
+typedef int rustls_io_error;
```
tgeoghegan, line 288: nit: `rustls_io_result` for consistency with `enum rustls_result`? Also what range of errnos will the nonzero values be drawn from on Windows?

tgeoghegan (approved):
I left some nits around doccomments, the types for buffer sizes and some type names, but nothing that needs to block this change. LGTM!

src/io.rs
jsha: `self.callback(self.userdata, ...)` is ambiguous between method call and field access. So I'd have to write it as `(self.callback)(self.userdata, ...)`, which I like less.

src/io.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use libc::{c_void, size_t};
+
+use crate::error::rustls_io_error;
+
+/// A callback for rustls_server_session_read_tls or rustls_client_session_read_tls.
+/// An implementation of this callback should attempt to read up to n bytes from the
+/// network, storing them in `buf`. If any bytes were stored, the implementation should
+/// set out_n to the number of bytes stored and return 0. If there was an error,
+/// (including EAGAIN or EWOULDBLOCK), the implementation should return `errno`.
```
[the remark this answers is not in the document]
jsha: `crustls` cares about `errno` semantics because it will marshal an `errno` value into a `Result<_, io::Error>`, and `io::Error` knows how to pass through errnos. But you make a good point - I could probably embed arbitrary data inside an `io::Error` in its wrapped error. Are you thinking "arbitrary platform int" as the return type?

src/connection.rs
jsha: yeah, I think dynamic dispatch will wind up not being a big deal. And in #99 dynamic dispatch becomes even more a part of the library in a way that I think would be somewhat painful to remove.

src/client.rs
jsha: I think "network" is nicely evocative here, and folks who are reading/writing from other buffers instead will get the gist.

src/io.rs
jsha: As a practical matter, rustls' `write_tls` doesn't call `.flush()` on the `Write` its passed. So I'd rather not introduce the second callback, since it will never be called. Note that if the user wants the bytes flushed, they can call their OS' flush immediately after `write_tls()`.

src/io.rs
tgeoghegan: Oh yes, I've run into that and it's quite annoying. Your version is indeed better.

src/io.rs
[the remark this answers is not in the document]
tgeoghegan: I think it has to be. Passing the error returned from the client callback through [from_raw_os_error](https://doc.rust-lang.org/std/io/struct.Error.html#method.from_raw_os_error) and then back into an int with `std::io::Error::raw_os_error` won't work reliably unless you can guarantee that the callback returned an errno that `std::io::Error` thinks is valid. I think most values will end up as `io::ErrorKind::Other`, but you can get more perniciously incorrect errors back if the range of errors the callback returns are not errnos but their values overlap with the range of valid errnos. I suspect the best you can do is preserve the integer value returned by the callback unmodified, and trust the client of `crustls` to decide if that value is an errno or if it is drawn from some other range of error codes.

PR #99, opened 2021-05-20
Consolidate two session types into one connection type

### LINKED ISSUE #37: Factor out shared Session code
In rustls, `ClientSession` and `ServerSession` both implement the `Session` trait, which contains a number of shared methods. The C bindings for those methods are duplicated between client.rs and server.rs. We explored methods of de-duplicating them in #30, and concluded that:

 - It's worthwhile.
 - The ffi_panic_boundary_foo and the try_ref_from_ptr macros have to be in the outer function that calls the shared function.
 - Using a macro to generate code for each type of peer won't work because of token pasting issues.
 
 We should define a new shared `session.rs`. Each method, e.g. `rustls_client_session_read`, will call a corresponding `rustls_session_read(&mut impl Session, ...)`. For some methods, like `rustls_session_wants_read`, the body will be a single line. For these cases it's probably still worthwhile defining a shared version, just for consistency across methods of `Session`.

jsha:
This reduces duplication in the API and in the implementation, and aligns naming with what's in rustls master (connection instead of session).

This looks like a ton of new code in `connection.rs`, but the method bodies are all copied from `server.rs` and `client.rs`, with minor tweaks for the new setup. The individual commits are pretty reasonable chunks to review on their own if you'd like.

Build on top of #95.
Fixes #37

/cc @icing @djc - reviews welcome!

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -28,26 +28,28 @@ pub(crate) enum Inner {
     Server(ServerSession),
 }
 
+impl Conn {
+    fn as_ref(&self) -> &dyn Session {
+        match &self.conn {
+            Inner::Client(c) => c,
+            Inner::Server(c) => c,
+        }
+    }
+
+    fn as_mut(&mut self) -> &mut dyn Session {
+        match &mut self.conn {
+            Inner::Client(c) => c,
+            Inner::Server(c) => c,
+        }
+    }
+}
+
 pub struct rustls_connection {
     _private: [u8; 0],
 }
 
 impl CastPtr for rustls_connection {
     type RustType = Conn;
-}
-
-fn dyn_conn(conn: &Conn) -> &dyn Session {
-    match &conn.conn {
-        Inner::Client(c) => c,
-        Inner::Server(c) => c,
-    }
-}
-
-fn dyn_conn_mut(conn: &mut Conn) -> &mut dyn Session {
-    match &mut conn.conn {
-        Inner::Client(c) => c,
-        Inner::Server(c) => c,
-    }
 }
 
 /// Set the userdata pointer associated with this connection. This will be passed
@@ -85,7 +87,7 @@ pub extern "C" fn rustls_connection_read_tls(
         let callback: ReadCallback = try_callback!(callback);
 
         let mut reader = CallbackReader { callback, userdata };
-        let n_read: usize = match dyn_conn_mut(conn).read_tls(&mut reader) {
+        let n_read: usize = match conn.as_mut().read_tls(&mut reader) {
             Ok(n) => n,
             Err(e) => return rustls_io_error(e.raw_os_error().unwrap_or(EIO)),
         };
@@ -118,7 +120,7 @@ pub extern "C" fn rustls_connection_write_tls(
         let callback: WriteCallback = try_callback!(callback);
 
         let mut writer = CallbackWriter { callback, userdata };
-        let n_written: usize = match dyn_conn_mut(conn).write_tls(&mut writer) {
+        let n_written: usize = match conn.as_mut().write_tls(&mut writer) {
             Ok(n) => n,
             Err(e) => return rustls_io_error(e.raw_os_error().unwrap_or(EIO)),
         };
@@ -138,7 +140,7 @@ pub extern "C" fn rustls_connection_process_new_packets(
             Ok(g) => g,
             Err(_) => return rustls_result::Panic,
         };
-        let result = match dyn_conn_mut(conn).process_new_packets() {
+        let result = match conn.as_mut().process_new_packets() {
             Ok(()) => rustls_result::Ok,
             Err(e) => map_error(e),
         };
@@ -153,7 +155,7 @@ pub extern "C" fn rustls_connection_process_new_packets(
 pub extern "C" fn rustls_connection_wants_read(conn: *const rustls_connection) -> bool {
     ffi_panic_boundary! {
         let conn: &Conn = try_ref_from_ptr!(conn);
-        dyn_conn(conn).wants_read()
+        conn.as_ref().wants_read()
     }
 }
 
@@ -161,7 +163,7 @@ pub extern "C" fn rustls_connection_wants_read(conn: *const rustls_connection) -
 pub extern "C" fn rustls_connection_wants_write(conn: *const rustls_connection) -> bool {
     ffi_panic_boundary! {
         let conn: &Conn = try_ref_from_ptr!(conn);
-        dyn_conn(conn).wants_write()
+        conn.as_ref().wants_write()
     }
 }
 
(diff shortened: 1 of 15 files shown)


src/connection.rs
```diff
@@ -28,6 +28,22 @@ pub(crate) enum Inner {
     Server(ServerSession),
 }
 
+impl Conn {
+    fn as_ref(&self) -> &dyn Session {
```
djc, line 32: Maybe these should become `AsRef` and `AsMut` implementations?

djc:
I like the enum + `as_ref()`/`as_mut()` approach. I think the design looks good, I only skimmed all the changes that moved code around.

src/connection.rs
jsha: I tried this and got lifetime problems I couldn't figure out the source of:

```rust
impl AsRef<dyn Session> for Conn {
    fn as_ref(&self) -> &dyn Session {
        match &self.conn {
            Inner::Client(c) => c,
            Inner::Server(c) => c,
        }
    }
}

impl AsMut<dyn Session> for Conn {
    fn as_mut(&mut self) -> &mut dyn Session {
        match &mut self.conn {
            Inner::Client(c) => c,
            Inner::Server(c) => c,
        }
    }
}
```

```
error: `impl` item signature doesn't match `trait` item signature
   --> src/connection.rs:32:5
    |
32  |     fn as_ref(&self) -> &dyn Session {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ found `fn(&Conn) -> &dyn Session`
    | 
   ::: /home/jsha/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/convert/mod.rs:158:5
    |
158 |     fn as_ref(&self) -> &T;
    |     ----------------------- expected `fn(&Conn) -> &(dyn Session + 'static)`
    |
    = note: expected `fn(&Conn) -> &(dyn Session + 'static)`
               found `fn(&Conn) -> &dyn Session`
    = help: the lifetime requirements from the `impl` do not correspond to the requirements in the `trait`
    = help: verify the lifetime relationships in the `trait` and `impl` between the `self` argument, the other inputs and its output

error: `impl` item signature doesn't match `trait` item signature
   --> src/connection.rs:41:5
    |
41  |     fn as_mut(&mut self) -> &mut dyn Session {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ found `fn(&mut Conn) -> &mut dyn Session`
    | 
   ::: /home/jsha/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/convert/mod.rs:199:5
    |
199 |     fn as_mut(&mut self) -> &mut T;
    |     ------------------------------- expected `fn(&mut Conn) -> &mut (dyn Session + 'static)`
    |
    = note: expected `fn(&mut Conn) -> &mut (dyn Session + 'static)`
               found `fn(&mut Conn) -> &mut dyn Session`
    = help: the lifetime requirements from the `impl` do not correspond to the requirements in the `trait`
    = help: verify the lifetime relationships in the `trait` and `impl` between the `self` argument, the other inputs and its output
```

If you have any idea what's wrong here I'd be happy to switch over to `AsRef` / `AsMut`.

djc: Did you try adding the `+ 'static`?

src/connection.rs
```diff
@@ -1,41 +1,360 @@
...
+    is_close_notify,
+};
+use crate::{try_mut_from_ptr, try_slice, userdata_push, CastPtr};
+use rustls_result::NullParameter;
 
-// Call Session::read_tls, providing an &mut dyn Write implemented with a C callback.
-pub(crate) fn read_tls(
-    session: &mut dyn Session,
-    callback: ReadCallback,
+pub(crate) struct Conn {
```
tgeoghegan, line 21:
```suggestion
pub(crate) struct Connection {
```

src/connection.rs
```diff
@@ -1,41 +1,360 @@
...
+    pub(crate) conn: Inner,
+    pub(crate) userdata: *mut c_void,
+}
+
+pub(crate) enum Inner {
```
tgeoghegan, line 26: nit: I think it'd be nicer if the `conn` member were more completely abstracted away from clients. At a glance, there is one spot where a `struct Conn[ection]` needs to be handled as a concrete `ServerSession` (`rustls_server_session_get_sni_hostname`) and otherwise clients are expected to stick to methods in `trait Session`. I think the `enum Inner` type and the `conn` member on `struct Conn[ection]` should be private to `connection.rs`, and you can enable `rustls_server_session_get_sni_hostname` by providing something like a `TryFrom` that attempts to convert `struct Conn` into `ServerSession`. However the whole thing is private to the crate so I don't think it's a big deal.

src/connection.rs
```diff
@@ -1,41 +1,360 @@
...
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn rustls_connection_is_handshaking(conn: *const rustls_connection) -> bool {
+    ffi_panic_boundary! {
+        let conn: &Conn = try_ref_from_ptr!(conn);
+        conn.as_ref().is_handshaking()
+    }
+}
```
tgeoghegan, line 178: nit
```suggestion
}

```

src/connection.rs
```diff
@@ -1,41 +1,360 @@
...
+            },
+            None => null()
+        }
+    }
+}
+
+/// Get the ALPN protocol that was negotiated, if any. Stores a pointer to a
+/// borrowed buffer of bytes, and that buffer's len, in the output parameters.
+/// The borrow lives as long as the session.
+/// If the session is still handshaking, or no ALPN protocol was negotiated,
```
tgeoghegan, line 226:
```suggestion
/// The borrow lives as long as the connection.
/// If the connection is still handshaking, or no ALPN protocol was negotiated,
```

src/connection.rs
```diff
@@ -1,41 +1,360 @@
...
+    conn: *const rustls_connection,
+) -> *const rustls_supported_ciphersuite {
+    ffi_panic_boundary! {
+        let conn: &Conn = try_ref_from_ptr!(conn);
+        match conn.as_ref().get_negotiated_ciphersuite() {
+            Some(cs) => cs as *const SupportedCipherSuite as *const _,
+            None => null(),
+        }
+    }
+}
```
tgeoghegan, line 283:
```suggestion
}

```

tgeoghegan (approved):
LGTM with some nits.

src/connection.rs
jsha: Ah this is a good idea. Can't do `TryFrom` since we're working with refs, not owned values. But I think `AsRef<ServerSession>` and so on can work.

PR #103, opened 2021-05-31
Store peer certificates on Connection instance

### LINKED ISSUE #101: rustls_connection_get_peer_certificate() broken
If I read this correctly now, the rustls v0.19.x returns `session.get_peer_certificates()` as copy. This will be changed in the next release to a reference.

The crustls rustls_connection_get_peer_certificate() returns pointers to these copies in 0.6.0, any use is use after free.

Either these need to be `Arc`ed or the next rustls is needed, I guess?

icing:
This fixes a use-after-free (#101) that would happen in any code using rustls_connection_get_peer_certificate.

Also, add rustls_certificate_get_der.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -31,6 +31,26 @@ pub struct rustls_certificate {
 impl CastPtr for rustls_certificate {
     type RustType = Certificate;
 }
+
+/// Get the DER data of the certificate itself.
+/// The data is owned by the certificate and has the same lifetime.
+#[no_mangle]
+pub extern "C" fn rustls_certificate_get_der(
+    cert: *const rustls_certificate,
+    out_der_data: *mut *const u8,
+    out_der_len: *mut size_t,
+) -> rustls_result {
+    ffi_panic_boundary! {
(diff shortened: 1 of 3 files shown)


jsha:
It seems like the only useful thing to do with a `rustls_certificate` is to get a reference to its bytes (with rustls_certificate_get_der). What do you think of making `rustls_connection_get_certificate` give a reference to the bytes rather than a reference to the `Certificate` object, saving a step?

icing:
I was hoping that rustls would be more forward in supporting access to x.509 fields and extensions. If this is not on the horizon, I can fold these two methods into one.

This would mean I need openssl to make client certificates work. And even then, the support will be incomplete, as I seem unable to navigate the trust chain of peer certificates in rustls. Something necessary for OCSP verification. CLRs need to be processed by openssl as well.

At this point, I am leaning to removing client certificate support in `mod_tls`. If it needs to use openssl for the juicy parts, there is no point in re-writing that.

jsha (approved):

PR #106, opened 2021-06-04
Switch to send/recv for sockets.

jsha:
Previously we did that only for Windows, but it works on other platforms
so we'll be consistent and use the same thing.

Also, fix a missing flags argument for send().

Thanks for the bug report @gvanem! Can you review?

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -240,7 +240,7 @@ int read_cb(void *userdata, uint8_t *buf, uintptr_t len, uintptr_t *out_n)
 {
   ssize_t n = 0;
   struct demo_conn *conn = (struct demo_conn*)userdata;
-  n = recv(conn->fd, buf, len,0 );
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -242,7 +240,7 @@ int read_cb(void *userdata, uint8_t *buf, uintptr_t len, uintptr_t *out_n)
 {
   ssize_t n = 0;
   struct demo_conn *conn = (struct demo_conn*)userdata;
-  n = read(conn->fd, buf, len);
+  n = recv(conn->fd, buf, len,0 );
```
tgeoghegan, line 243:
```suggestion
  n = recv(conn->fd, buf, len, 0);
```

tgeoghegan (approved):
LGTM, with one whitespace change

PR #107, opened 2021-06-08
Add logging support

jsha:
This uses name mangling to ensure we get our own isolated copy of the `log` crate, so we are free to set our own logger. The technique is described at https://github.com/jsha/rust-private-loggers.

That logger will be set each time rustls_connection_set_log_callback is called. Every time after the first, we will fail to register a log (but will ignore the error because we already registered one, and our log instances are all identical).

Our log implementation uses the same thread-local userdata values we use for callbacks that implement customizable traits in rustls. The thread-local storage now includes a log_callback parameter as well as a userdata parameter.

Still a draft because:

 - I'd like to think about exporting the [`Level`](https://docs.rs/log/0.4.14/log/enum.Level.html) enum from `log` so users can dispatch logs to different places by level.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -118,8 +118,11 @@ pub extern "C" fn rustls_connection_set_userdata(
     conn.userdata = userdata;
 }
 
+/// Set the logging callback for this connection. The log callback will be invoked
+/// with the userdata parameter previously set by rustls_connection_set_userdata, or
+/// NULL if no userdata was set.
 #[no_mangle]
-pub extern "C" fn rustls_client_connection_set_log_callback(
+pub extern "C" fn rustls_connection_set_log_callback(
     conn: *mut rustls_connection,
     cb: rustls_log_callback,
 ) {
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -6,6 +6,7 @@ use rustls::{Certificate, ClientSession, ServerSession, Session, SupportedCipher
 
 use crate::io::{CallbackReader, CallbackWriter, ReadCallback, WriteCallback};
 use crate::is_close_notify;
+use crate::log::{ensure_log_registered, rustls_log_callback};
 use crate::{
     cipher::{rustls_certificate, rustls_supported_ciphersuite},
     error::{map_error, rustls_io_result, rustls_result},
@@ -19,6 +20,7 @@ use rustls_result::NullParameter;
 pub(crate) struct Connection {
     conn: Inner,
     userdata: *mut c_void,
+    log_callback: rustls_log_callback,
     peer_certs: Option<Vec<Certificate>>,
 }
 
@@ -32,6 +34,7 @@ impl Connection {
         Connection {
             conn: Inner::Client(s),
             userdata: null_mut(),
+            log_callback: None,
             peer_certs: None,
         }
     }
@@ -40,6 +43,7 @@ impl Connection {
         Connection {
             conn: Inner::Server(s),
             userdata: null_mut(),
+            log_callback: None,
             peer_certs: None,
         }
     }
@@ -114,6 +118,16 @@ pub extern "C" fn rustls_connection_set_userdata(
     conn.userdata = userdata;
 }
 
+#[no_mangle]
+pub extern "C" fn rustls_client_connection_set_log_callback(
+    conn: *mut rustls_connection,
+    cb: rustls_log_callback,
+) {
+    let conn: &mut Connection = try_mut_from_ptr!(conn);
+    ensure_log_registered();
+    conn.log_callback = cb;
+}
+
 /// Read some TLS bytes from the network into internal buffers. The actual network
 /// I/O is performed by `callback`, which you provide. Rustls will invoke your
 /// callback with a suitable buffer to store the read bytes into. You don't have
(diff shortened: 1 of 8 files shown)


src/log.rs
```diff
@@ -0,0 +1,38 @@
...
+
+struct Logger {}
+
+impl log::Log for Logger {
+    fn enabled(&self, _metadata: &log::Metadata<'_>) -> bool {
+        true
+    }
+    fn log(&self, record: &log::Record<'_>) {
+        if let Ok((Some(cb), userdata)) = log_callback_get() {
+            let message = format!("{} {} {}", record.target(), record.level(), record.args());
```
icing, line 15: Passing level separately as a defined int/enum is definitely useful to the application.

src/lib.rs
```diff
@@ -32,7 +34,12 @@ use crate::panic::PanicOrDefault;
 // Rust code, we model these thread locals as a stack, so we can always
 // restore the previous version.
 thread_local! {
-    pub static USERDATA: RefCell<Vec<*mut c_void>> = RefCell::new(Vec::new());
+    pub static USERDATA: RefCell<Vec<Userdata>> = RefCell::new(Vec::new());
+}
+
+pub struct Userdata {
```
icing, line 40: One might consider simply storing the rustls_connection here. It seems to contain the application context which is exactly what we want to push/pop with this.

If my Rust were better, I'd also knew if a `Arc<dyn AppContextTrait>` could be used here.

src/connection.rs
```diff
@@ -114,6 +118,16 @@ pub extern "C" fn rustls_connection_set_userdata(
     conn.userdata = userdata;
 }
 
+#[no_mangle]
+pub extern "C" fn rustls_client_connection_set_log_callback(
```
icing, line 122: Should this be rather `rustls_connection_set_log_callback`?

icing (commented):
I like this very much and would want to use this in mod_tls. Since logging in Apache is highly configurable, it would be very useful to expose the level.

src/crustls.h
```diff
@@ -330,6 +330,12 @@ typedef enum rustls_result (*rustls_session_store_get_callback)(rustls_session_s
  */
 typedef enum rustls_result (*rustls_session_store_put_callback)(rustls_session_store_userdata userdata, const struct rustls_slice_bytes *key, const struct rustls_slice_bytes *val);
 
+typedef struct rustls_log_params {
+  struct rustls_str message;
+} rustls_log_params;
+
+typedef void (*rustls_log_callback)(void *userdata, const struct rustls_log_params *params);
```
tgeoghegan, line 337: Wrapping `rustls_str` in another struct seems like overkill to me. Is the idea here that it would be easier to add members to `rustls_log_params` than it would be to add arguments to `rustls_log_callback`?

src/lib.rs
jsha: I considered this, but this would require us to violate the borrowing rules for `Connection` in ways we don't already. Specifically in the case where we are in a call that takes a `*mut rustls_connection` and turns it into a `&mut Connection`, we would then have to put a `*mut Connection` in `Userdata` and use unsafe to dereference it. I think this winds up being sound due to the threading properties of rustls (everything happens on the thread you called on), but it still seems like unnecessary risk.

Of course we are taking the same risk (having multiple mutable pointers) with `userdata: *mut c_void`, but that is unavoidable risk, if we want a general-purpose callback.

src/crustls.h
jsha: Yeah, the idea is to add other arguments, like log level. This follows a pattern for our other callbacks where they take a pointer to a struct rather than a list of params. One could argue is that that pattern is useful for large lists of params, but with small lists (like level and message), it's better to just pass the params directly. The flip side is that it's more consistent for our callbacks to always take a struct.

jsha:
One other open question: the `log` crate has an optimization such that if you're not using, e.g. `trace` logs, you pay near-zero cost. For instance you don't need to allocate a string to format the message. This PR can't support that optimization. If you don't register a callback, we won't turn on logging, and everything is cheap. If you register a callback at any point in your program, you'll pay the overhead for logging all messages, including `trace` level messages, in all parts of your program. The overhead is probably not particularly high but it's unavoidable.

So what I'm asking is: How important is it to design the API so that, for instance, you can get `warn` and `error` logs without paying the price for `trace` logs? This would probably involve some extra complication.

I guess one answer is: The way we would do this is have a `rustls_connection_set_log_level`, which controls the log level for that connection specifically. This would involve adding an extra field to `Userdata`. But we could do it after the fact as a non-API-breaking change, if benchmarking showed it to be valuable.

Also there are these compile time filters to entirely turn off logs below a certain level so they don't even get compiled into the binary: https://docs.rs/log/0.4.14/log/#compile-time-filters. Should we expose these?

tgeoghegan:
I vote for
>But we could do it after the fact as a non-API-breaking change, if benchmarking showed it to be valuable.
If it can be changed in the future without breaking clients then doing the simplest thing possible now seems best.

djc:
> But we could do it after the fact as a non-API-breaking change, if benchmarking showed it to be valuable.

+1 from me as well.

tgeoghegan (approved):

PR #108, opened 2021-06-09
additions to client config setups

icing:
Working on proxy connections in mod_tls, I added some functions similar to what we have in server configs.

 * added the following functions:
  - rustls_client_config_builder_from_config() to clone an existing config and make modifications
  - rustls_client_config_builder_free() to free an aborted config building
  - rustls_client_config_builder_use_roots() to let the config use a already prepped root_cert_store
 * removed the local definition of `rustls_root_cert_store`, as we have that defined in cipher.rs

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -13,6 +13,7 @@ use webpki::DNSNameRef;
 
 use crate::cipher::rustls_root_cert_store;
 use crate::connection::{rustls_connection, Connection};
+use crate::enums::rustls_tls_version_from_u16;
 use crate::error::{self, result_to_tlserror, rustls_result};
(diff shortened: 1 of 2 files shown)


jsha:
It looked like you had merged in the vectored changes (understandably, when you've got a couple of necessary diffs stacked up). In order to merge, I reset to just the diffs that were in this change, and then pushed some small tweaks from my review on top: https://github.com/abetterinternet/crustls/pull/108/commits/d06a97ed866b3b32376e700c9e30df76bfc2e34a. Once tests pass I'll squash merge.

jsha (approved):

PR #110, opened 2021-06-14
Remove rustls-native-roots dependency

jsha:
There are some issues integrating with trust stores: https://github.com/ctz/rustls-native-certs/issues/16, and also some issues with regards to what gets run in forked processes: https://github.com/kornelski/rust-security-framework/issues/136.


tgeoghegan (changes requested):
This breaks the tests on macOS, because `/etc/ssl/certs/ca-certificates.crt` doesn't exist there. I don't have it on my Fedora 34 install either; instead I have `/etc/ssl/certs/ca-bundle.crt`. And I doubt that bundle exists on Windows. `curl` hosts the Mozilla CA bundle at https://curl.se/ca/cacert.pem, so perhaps we could at least make CI tests behave reliably by making the makefile fetch the cert bundle before running `crustls-demo`.

tgeoghegan:
Or an alternative that wouldn't depend on `httpbin.org`'s certain chain being trusted by the CA bundle currently hosted by curl.se would be to fetch the cert chain for `httpbin.org` and then pass that to `crustls-demo` as a trust store, since the point of `crustls`'s tests is to exercise making connections over HTTPS rather than certificate trust evaluation. Continuing on that line of thought, the most robust way forward might be to have the tests spin up a local HTTPS server with a self-signed certificate, which would remove all external dependencies and even make it possible to run `crustls` tests without reaching the internet.

jsha:
Merged latest main, which includes a switch to talking to a local server and using a locally trusted root.

tgeoghegan (approved):

PR #111, opened 2021-06-15
Add test server and use with test client

jsha:
This allows us to test the server side of the code as well as the client, and also to make our tests more self-contained.

This splits out a lot of functions and some structs into common.[ch], and moves all the C test code into tests/. To have a somewhat more realistic test, the server keeps connections alive after a request, and so the client has to judge the end of a response by a Content-Length header. That in turn required the introduction of a growable buffer to hold the headers, and some basic header parsing. So far only the client parses Content-Length, and neither the client nor the server fully supports connection reuse, but the scaffolding is there.

The server uses a local root, and the client is configured to trust that root.

This gets rid of the dependency on system roots, unblocking #110.

The client.c diff is collapsed; don't forget to hit "Load diff."


jsha:
Hm, getting some weird connection refused errors. I have some code to try and wait for the server to be listening before running the client, and it all works nicely locally. Not sure why it's not working in GitHub Actions. Will try and figure that out later this week.

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+
+  result = rustls_connection_process_new_packets(rconn);
+  if(result != RUSTLS_RESULT_OK) {
+    print_error("in process_new_packets", result);
+    return CRUSTLS_DEMO_ERROR;
+  }
+
+  result = copy_plaintext_to_buffer(conn);
+  if(result != CRUSTLS_DEMO_CLOSE_NOTIFY) {
+    fprintf(stderr, "do_read returning %d\n", result);
```
tgeoghegan, line 114: Should this be a `print_error()` call?

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+    return result;
+  }
+
+  char buf[2048];
```
tgeoghegan, line 118: Interesting that you insist on declaring and initializing `err` and `result` at the top of the function like we're back in 1989 but allocate 2k on the stack ~20 lines into the function!

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+  n = read(conn->fd, buf, sizeof(buf));
+  if(n != 0 && errno != EWOULDBLOCK) {
+    fprintf(stderr, "read returned %ld after receiving close_notify: %s\n", n, strerror(errno));
+    return CRUSTLS_DEMO_ERROR;
+  }
+  return CRUSTLS_DEMO_CLOSE_NOTIFY;
+}
+
+void
+send_response(struct conndata_t *conn) {
```
tgeoghegan, line 130:
```suggestion
send_response(struct conndata_t *conn)
{
```

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+  size_t n;
+  rustls_connection_write(rconn, (const uint8_t *)response, strlen(response), &n);
+  if (n != strlen(response)) {
+    fprintf(stderr, "failed to write all response bytes. wrote %ld\n", n);
+    abort();
+  }
+}
+
+void *
+handle_conn(void *userdata) {
```
tgeoghegan, line 142:
```suggestion
handle_conn(void *userdata)
{
```

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+
+void
+send_response(struct conndata_t *conn) {
+  struct rustls_connection *rconn = conn->rconn;
+  const char* response = "HTTP/1.1 200 OK\r\nContent-Length: 6\r\n\r\nhello\n";
+  size_t n;
+  rustls_connection_write(rconn, (const uint8_t *)response, strlen(response), &n);
+  if (n != strlen(response)) {
+    fprintf(stderr, "failed to write all response bytes. wrote %ld\n", n);
+    abort();
```
tgeoghegan, line 137: It doesn't really matter since this is a test program, but `abort()` seems a little harsh since it'd be pretty easy to return an error to the single callsite in `handle_conn`, which already returns an error.

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+    if(result == -1) {
+      perror("select");
+      goto cleanup;
+    }
+    if(result == 0) {
+      fprintf(stderr, "no fds from select, sleeping\n");
+      ts.tv_sec = 0;
+      ts.tv_nsec = 1000000000;
+      nanosleep(&ts, NULL);
+    }
```
tgeoghegan, line 176: Why not pass a timeout to `select(2)` instead of sleeping?

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+  return certified_key;
+}
+
+int
+main(int argc, const char **argv)
+{
+  int ret = 1;
+  int result = 1;
+  const char *certfile = argv[1];
+  const char *keyfile = argv[2];
```
tgeoghegan, line 260: It's not safe to access elements of `argv` before you check `if(argc <=2)`, below.

tests/server.c
```diff
@@ -0,0 +1,359 @@
...
+    conndata_t *conndata;
+    conndata = calloc(1, sizeof(conndata_t));
+    conndata->fd = clientfd;
+    conndata->rconn = rconn;
+    ret = pthread_create(&thrd, NULL, handle_conn, conndata);
+    if (ret != 0) {
+      fprintf(stderr, "error from pthread_create: %d\n", ret);
+      goto cleanup;
+    }
+    pthread_join(thrd, NULL);
```
tgeoghegan, line 340: What's the point of handling connections on threads if you're going to block the loop that does accept on `pthread_join(3)` anyway? I think it's perfectly reasonable for this server to only handle one connection at a time, but you could just block on a call to `handle_conn` in the loop and make your life a little easier.

tgeoghegan (changes requested):
I mostly have style nits, and I tried to be guided by the implicit style of the existing C in crustls rather than my own preferences. It's debatable how rigorous we need to be about style or correctness in test/demo code but then again it's what users of `crustls` are most likely to examine as they try to figure out how to use the library so perhaps we should be picky about its quality.

tests/common.c
```diff
@@ -0,0 +1,297 @@
...
+    return CRUSTLS_DEMO_ERROR;
+  }
+#else
+  int flags;
+  flags = fcntl(sockfd, F_GETFL, 0);
+  if(flags < 0) {
+    perror("getting socket flags");
+    return CRUSTLS_DEMO_ERROR;
+  }
+  flags = fcntl(sockfd, F_SETFL, flags | O_NONBLOCK);
```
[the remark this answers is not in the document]
jsha: Yeah, in my brief noodling on StackOverflow and elsewhere this seems to be the way it's done. We're pretty much assuming exclusive control of the fd by one thread at a time so I think it's okay.

tests/common.c
```diff
@@ -0,0 +1,297 @@
...
+}
+
+char *
+bytevec_writeable(struct bytevec *vec)
+{
+  return vec->data + vec->len;
+}
+
+void
+bytevec_consume(struct bytevec *vec, size_t n) {
```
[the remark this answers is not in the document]
jsha: Yeah, exactly. I figure it's too late to check here, and checking here means adding an error return value, which adds pain.

tests/server.c
jsha: No, because that's only for rustls errors. This is an error internal to the demo.

tests/common.h
```diff
@@ -0,0 +1,93 @@
...
+};
+
+/* A growable vector of bytes. */
+struct bytevec {
+  char *data;
+  size_t len;
+  size_t capacity;
+};
+
+struct conndata_t {
```
tgeoghegan, line 20: nit
```suggestion
struct conndata {
```
to match struct naming convention in the rest of the project.

tests/common.c
```diff
@@ -0,0 +1,307 @@
...
+    newsize = vec->len + n;
+    if(newsize < vec->capacity * 2) {
+      newsize = vec->capacity * 2;
+    }
+    newdata = realloc(vec->data, newsize);
+    if(newdata == NULL) {
+      fprintf(stderr, "out of memory trying to get %ld bytes\n", newsize);
+      return CRUSTLS_DEMO_ERROR;
+    }
+    vec->data = (char *)newdata;
```
tgeoghegan, line 171: nit: Is this cast necessary? C will let you assign the `void *` returned from `calloc(3)` to `char *` without a warning. The cast is only needed if you also want to compile this as C++, where the compilers are pickier, but then you'd also have to conditionally put in `extern "C" {}` around everything.

tests/common.c
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * Return Value
+ *  pointer to the memory address of the match or NULL.
+ */
+void *
+memmem(const void *haystack, size_t haystacklen, const void *needle,
+       size_t needlelen)
+{
+  const char *bf = (const char *)haystack;
+  const char *pt = (const char *)needle;
```
tgeoghegan, line 236: nit
```suggestion
  const char *bf = haystack;
  const char *pt = needle;
```
Same as elsewhere: the casts are only needed if compiling as C++.

tests/server.c
```diff
@@ -0,0 +1,368 @@
...
+  result = copy_plaintext_to_buffer(conn);
+  if(result != CRUSTLS_DEMO_CLOSE_NOTIFY) {
+    fprintf(stderr, "do_read returning %d\n", result);
+    return result;
+  }
+
+  /* If we got a close_notify, verify that the sender then
+   * closed the TCP connection. */
+  n = read(conn->fd, buf, sizeof(buf));
+  if(n != 0 && errno != EWOULDBLOCK) {
```
tgeoghegan, line 125: This doesn't seem right: if `n = 1` (i.e. `read(2)` succeeds), then `errno` will have some garbage value unrelated to the `read(2)` call. Should this be `if(n > 0 || (n < 0 && errno != EWOULDBLOCK))`?

tests/server.c
```diff
@@ -0,0 +1,368 @@
...
+    nonblock(clientfd);
+
+    result = rustls_server_connection_new(server_config, &rconn);
+    if(result != RUSTLS_RESULT_OK) {
+      print_error("making session", result);
+      goto cleanup;
+    }
+
+    struct conndata_t *conndata;
+    conndata = calloc(1, sizeof(struct conndata_t));
```
tgeoghegan, line 350: I think you leak `conndata`, since it does not get `free(3)`d in `handle_conn`, though I would vote for freeing it in this scope after `handle_conn` returns to make it more obvious that `calloc(3)` is paired with a `free(3)`. For similar reasons you might consider closing `clientfd` in this function instead of in `handle_conn`.

Another option would be to allocate `conndata` on the stack, which is safe since you don't need to share the value with another thread and this function will not return until `handle_conn` does.

tests/server.c
```diff
@@ -0,0 +1,368 @@
...
+  rustls_server_config_builder_set_certified_keys(
+    config_builder, &certified_key, 1);
+  server_config = rustls_server_config_builder_build(config_builder);
+
+#ifdef _WIN32
+  WSADATA wsa;
+  WSAStartup(MAKEWORD(1, 1), &wsa);
+#endif
+
+  int sockfd = socket(AF_INET, SOCK_STREAM, 0);
```
tgeoghegan, line 301: `sockfd` is never closed, but it's probably fine to leak it since it'll get closed no matter what when the program exits.

tests/client.c
```diff
@@ -333,6 +218,9 @@ send_request_and_read_response(struct demo_conn *conn,
   fd_set read_fds;
   fd_set write_fds;
   size_t n = 0;
+  const char *body, *content_length_str, *content_length_end;
```
tgeoghegan, line 221: nit
```suggestion
  const char *body; 
  const char *content_length_str;
  const char *content_length_end;
```
...would be my preference but feel free to ignore me.

tests/client.c
```diff
@@ -492,34 +435,41 @@ main(int argc, const char **argv)
...
-  const char *path = argv[2];
+  const char *port = argv[2];
+  const char *path = argv[3];
 
   struct rustls_client_config_builder *config_builder =
     rustls_client_config_builder_new();
   const struct rustls_client_config *client_config = NULL;
-  const struct rustls_slice_bytes alpn_http11 = { (const uint8_t *)"http/1.1", 8 };
+  const struct rustls_slice_bytes alpn_http11 = { (const uint8_t *)"http/1.1",
+                                                  8 };
```
tgeoghegan, line 453: nit: If you're OK with requiring C99, designated initializers are very good for readability
```suggestion
  const struct rustls_slice_bytes alpn_http11 = { 
      .data = (const uint8_t *)"http/1.1",
      .len = 8,
  };
```

tgeoghegan (changes requested):

tests/client.c
jsha: I decided not to use the designated initializers for now, but to set fields by hand.

tests/server.c
jsha: I opted to put the free in `handle_conn` since I still hope to make that execute concurrently eventually.

tests/server.c
jsha: Fixed; lemme know what you think.

tgeoghegan (approved):
🚢

PR #112, opened 2021-06-16
CallbackWriter optimisation for vectored use

icing:
* adding an implementation of Write::write_vectored() to avoid
    the default trait implementation that copies all slices into  one.
 * added a vectored callback type to crustls and a new rustls_connection_write_tls_vectored.


src/io.rs
```diff
@@ -97,4 +97,15 @@ impl Write for CallbackWriter {
...
     }
+
+    fn write_vectored(&mut self, bufs: &[IoSlice<'_>]) -> Result<usize> {
+        let mut out_n: usize = 0;
+        for buf in bufs.iter() {
+            out_n += match self.write(buf) {
+                Ok(n) => n,
+                e => return e,
+            }
+        }
```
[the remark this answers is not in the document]
icing: A valid point. With the callback interface as it now stand in `crustls`, this implementation is too naive.

jsha:
[The docs for Write::write_vectored say](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored):

> The default implementation calls write with either the first nonempty buffer provided, or an empty one if none exists.

So if, for instance, rustls always calls write_vectored with 3 IoSlices, currently it would have to call write_vectored 3 times, which would result in crustls firing its callback 3 times. That would typically mean 3 syscalls.

This PR as it currently stands still makes those 3 callbacks and 3 syscalls, and only saves in some internal call overhead.

We would do better by _adding_ some copying: If the slice of IoSlices has len > 1, we should copy them all into a temporary buffer, then send that buffer in a single callback.

We could do better still by providing a way to register a different callback for performing vectored I/O. But @djc has mentioned that write_vectored support in rustls may go away, so probably better not to expose it in the crustls API.

icing:
> [The docs for Write::write_vectored say](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored):
> 
> > The default implementation calls write with either the first nonempty buffer provided, or an empty one if none exists.
> 
> So if, for instance, rustls always calls write_vectored with 3 IoSlices, currently it would have to call write_vectored 3 times, which would result in crustls firing its callback 3 times. That would typically mean 3 syscalls.

In a very naive web server, this would happen. But you do not want to use such a thing in production.
 
> This PR as it currently stands still makes those 3 callbacks and 3 syscalls, and only saves in some internal call overhead.
> 
> We would do better by _adding_ some copying: If the slice of IoSlices has len > 1, we should copy them all into a temporary buffer, then send that buffer in a single callback.
> 
> We could do better still by providing a way to register a different callback for performing vectored I/O. But @djc has mentioned that write_vectored support in rustls may go away, so probably better not to expose it in the crustls API.

You ignore the fact that this change causes a ~10% improved throughput in my benchmark. So something improved, while in your argument it should have gotten worse.

I will experiment with a vectored callback function today. I think that is the way to go here.

src/io.rs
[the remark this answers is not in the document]
icing: Fixed in updated PR.

icing:
* Fixed the `write_vectored` implementation to return on the first incomplete write of a buffer.
 * added `rustls_connection_write_tls_vectored` with a new callback type the receives all `IoSlice`s as `rustls_slice_bytes`

src/io.rs
```diff
@@ -97,4 +99,85 @@ impl Write for CallbackWriter {
...
+
+    fn flush(&mut self) -> Result<()> {
+        Ok(())
+    }
+
+    fn write_vectored(&mut self, bufs: &[IoSlice<'_>]) -> Result<usize> {
+        let mut out_n: usize = 0;
+        let cb = self.callback;
+        let mut slices: Vec<rustls_slice_bytes> = Vec::new();
+        for buf in bufs.iter().filter(|b| !b.is_empty()) {
```
jsha: Do we want to provide a guarantee, separate from rustls, that we never pass zero-length slices to the callback? If so, why. It seems like rustls probably doesn't create such slices under normal circumstances. If we need additional guarantees for C code we should make sure to document them.

src/io.rs
```diff
@@ -97,4 +99,85 @@ impl Write for CallbackWriter {
...
+        let mut out_n: usize = 0;
+        for buf in bufs.iter().filter(|b| !b.is_empty()) {
+            out_n += match self.write(buf) {
+                Ok(n) => {
+                    if n != buf.len() {
+                        return Ok(out_n + n);
+                    }
+                    n
+                }
+                e => return e,
```
jsha: I think this is not right. If we've already written 100 bytes, and then we get EWOULDBLOCK, we need to return Ok(100), not Error.

jsha (changes requested):
> In a very naive web server, this would happen. But you do not want to use such a thing in production.

> You ignore the fact that this change causes a ~10% improved throughput in my benchmark. So something improved, while in your argument it should have gotten worse.

I'm taking it as implied from your first statement that mod_tls does not make a syscall every time the write callback is invoked, but instead buffers up to some fixed size and then sends a syscall. Is that right? If so, I'm not totally sure where the 10% speedup would have come from. Under my theory, I don't think it should _necessarily_ have gotten worse. Knowing now that mod_tls is doing internal buffering, it should have been about the same.

One possibility occurs to me: On `main`, if there are 3 chunks available, mod_tls would have to call rustls_connection_write 3 times, and that would invoke the callback 3 times. With your initial optimization for the non-vectored callback, it becomes a single call to rustls_connection_write, which makes 3 calls to the callback. That removes 2 times that we have to cross the C->Rust boundary. That shouldn't really be significant, but maybe the initial test that showed a 10% improvement ran in debug mode?

Do you have interest in performance-testing a variant where `write_vectored` concatenates into a single buffer before invoking the callback? I pushed a commit adding such a concatenation (17d23e86a4dc537b38bc44dfe9224cceec1e298a). I'm not 100% convinced it's the right path but I think it's worth testing.

jsha:
By the way, [the IoSlice docs say](https://doc.rust-lang.org/nightly/std/io/struct.IoSlice.html):

> It is semantically a wrapper around an &[u8], but is guaranteed to be ABI compatible with the iovec type on Unix platforms and WSABUF on Windows.

I think we are losing that ABI compatibility by wrapping in our rustls_byte_slice type. I have to admit I'm a bit lost here. Is it worth some changes to keep this ABI compatibility so users can pass the args straight through with less wrangling? It sounds like Apache doesn't need it because it has another I/O buffering layer, but perhaps other users might?

icing:
> By the way, [the IoSlice docs say](https://doc.rust-lang.org/nightly/std/io/struct.IoSlice.html):
> 
> > It is semantically a wrapper around an &[u8], but is guaranteed to be ABI compatible with the iovec type on Unix platforms and WSABUF on Windows.
> 
> I think we are losing that ABI compatibility by wrapping in our rustls_byte_slice type. I have to admit I'm a bit lost here. Is it worth some changes to keep this ABI compatibility so users can pass the args straight through with less wrangling? It sounds like Apache doesn't need it because it has another I/O buffering layer, but perhaps other users might?

If this is compatible, I am all for putting `Iovec` into the callback signature.

icing:
> One possibility occurs to me: On `main`, if there are 3 chunks available, mod_tls would have to call rustls_connection_write 3 times, and that would invoke the callback 3 times. With your initial optimization for the non-vectored callback, it becomes a single call to rustls_connection_write, which makes 3 calls to the callback. That removes 2 times that we have to cross the C->Rust boundary. That shouldn't really be significant, but maybe the initial test that showed a 10% improvement ran in debug mode?
> 

This is what I originally thought and why I benchmarked calltimes against rust. With my `release` stumbling resolved, call times showed to be comparable.

The positive aspect of the vectored callback is that the network layer in Apache can work more efficiently. With 1 push of n TLS records to the network, that layer  has more information to choose the correct strategy. It can send or buffer, it can wait for completion or return early and send it all async when the thread handling the request is long done.

The best strategy here is implemented in the `mpm` modules of the server (multi-processing modules) and there are varying implementations an admin may choose depending on use cases and/or OS platform. A module such as `mod_tls` or `mod_ssl` has no idea what is happening there. It can only help by providing data in the most efficient way.

> Do you have interest in performance-testing a variant where `write_vectored` concatenates into a single buffer before invoking the callback? I pushed a commit adding such a concatenation ([17d23e8](https://github.com/abetterinternet/crustls/commit/17d23e86a4dc537b38bc44dfe9224cceec1e298a)). I'm not 100% convinced it's the right path but I think it's worth testing.

We can test this, if you want. Maybe add a 3rd write_tls method, then I can easily test all 3 from the same crustls branch.

djc:
> I do not understand. If they hold `n` TLS records, how do they plan to write them out efficiently?

In a future release (not the next one), rustls probably will take caller-allocated memory and write consecutives TLS records into that. I'm not convinced this would be slower than writing into separate `IoSlice`s, but I don't have much experience with vectored I/O.

icing:
> > I do not understand. If they hold `n` TLS records, how do they plan to write them out efficiently?
> 
> In a future release (not the next one), rustls probably will take caller-allocated memory and write consecutives TLS records into that. Would that be slower than writing into separate `IoSlice`s? I don't have much experience with vectored I/O.

It depends how it is done.

Atm, `rustls` encrypts when plain data is written into a session. Since it can not immediately write this out, it has to keep the encrypted TLS records somewhere. In such a design a vectored write_tls is best, IMO.

If `rustls` is changed so that a `Write` implementation is set in the Session, it can always write out TLS data. Say, every TLS record it completes, it writes out immediately. It would still have to collect the record in the first place, of course. So this does not save anything on copying. Instead the application gets called for every TLS record and *has to copy* the data or write it immediately to the network. Same as now, only more frequently.

Or `rustls` makes a design where it does not even keep the complete TLS record but writes immediately chunks of if (1-n bytes as they come out of the cipher) to an application provided destination. This would be called *very* frequently and needs a fast implementation. However it could place the TLS record exactly into outgoing buffers and flush them as needed. Say, you have a fixed 64KB buffer, append the byte chunks and flush it when it is full.

OpenSSL follows this later approach: the application installs Binary IO handlers (BIOs) for in- and output. But note that such a design needs more than read/write. You have to take care of blocking IO and flushing of buffers.

jsha:
I pushed some changes to make it clear you can cast directly to a `struct iovec *`. I tried to get cbindgen to generate the header with a reference directly to `struct iovec` from `sys/uio.h`, like:

```
typedef rustls_io_result (*rustls_write_vectored_callback)(void *userdata, const struct iovec *iov, size_t count, size_t *out_n);
```

But I couldn't get it to work. So I did the second-best thing and defined an opaque `rustls_iovec` struct whose sole purpose is to be cast into a `struct iovec *`. I don't love it but it's progress.

If you're interested in splitting out just the write_tls_vectored stuff while we continue to noodle around with whether write_tls should implement the `write_vectored` method, I'd be happy to merge that and save you juggling some branches.

jsha:
I wound up pushing a revert of the vectored impl for the plain write_tls case, because I'd like to land the write_tls_vectored case and do a release (ahead of the upcoming rustls release), then we can fiddle around with which implementation is fastest for the write_tls case.

src/io.rs
```diff
@@ -98,3 +100,82 @@ impl Write for CallbackWriter {
...
+
+/// A callback for rustls_connection_write_tls_vectored.
+/// An implementation of this callback should attempt to write the bytes in
+/// the given `count` rustls_slice_bytes to the network. If any bytes were written,
+/// the implementation should set out_n to the number of bytes written and return 0.
+/// If there was an error, the implementation should return a nonzero rustls_io_result,
+/// which will be passed through to the caller. On POSIX systems, returning `errno` is convenient.
+/// On other systems, any appropriate error code works.
+/// (including EAGAIN or EWOULDBLOCK), the implementation should return `errno`.
+/// It's best to make one write attempt to the network per call. Additional write will
```
tgeoghegan, line 119:
```suggestion
/// It's best to make one write attempt to the network per call. Additional writes will
```

src/io.rs
```diff
@@ -98,3 +100,82 @@ impl Write for CallbackWriter {
...
+}
+
+/// A callback for rustls_connection_write_tls_vectored.
+/// An implementation of this callback should attempt to write the bytes in
+/// the given `count` rustls_slice_bytes to the network. If any bytes were written,
+/// the implementation should set out_n to the number of bytes written and return 0.
+/// If there was an error, the implementation should return a nonzero rustls_io_result,
+/// which will be passed through to the caller. On POSIX systems, returning `errno` is convenient.
+/// On other systems, any appropriate error code works.
+/// (including EAGAIN or EWOULDBLOCK), the implementation should return `errno`.
```
tgeoghegan, line 118: This line seems like a copy/paste error?

src/io.rs
```diff
@@ -98,3 +100,82 @@ impl Write for CallbackWriter {
...
+/// An alias for `struct iovec` from uio.h. You should cast `const struct rustls_iovec *` to
+/// `const struct iovec *`.
+#[cfg(unix)]
+pub struct rustls_iovec {
+    _private: [u8; 0],
+}
+
+/// A callback for rustls_connection_write_tls_vectored.
+/// An implementation of this callback should attempt to write the bytes in
+/// the given `count` rustls_slice_bytes to the network. If any bytes were written,
```
tgeoghegan, line 113: This is confusing: there's a reference to `rustls_slice_bytes` but that type doesn't occur in the function signature. I assume `count` is the number of buffers in the `struct iovec`?

tests/common.c
```diff
@@ -131,6 +132,22 @@ write_cb(void *userdata, const unsigned char *buf, size_t len, size_t *out_n)
...
 }
 
+#ifndef _WIN32
+rustls_io_result write_vectored_cb(
+    void *userdata, const struct rustls_iovec *iov, size_t count, size_t *out_n)
+{
+  ssize_t n = 0;
+  struct conndata *conn = (struct conndata *)userdata;
+
+  n = writev(conn->fd, (const struct iovec*)iov, count);
```
tgeoghegan, line 142:
```suggestion
  n = writev(conn->fd, (const struct iovec *)iov, count);
```

src/io.rs
```diff
@@ -98,3 +100,82 @@ impl Write for CallbackWriter {
...
+    fn flush(&mut self) -> Result<()> {
+        Ok(())
+    }
+
+    fn write_vectored(&mut self, bufs: &[IoSlice<'_>]) -> Result<usize> {
+        let mut out_n: usize = 0;
+        let cb = self.callback;
+        let slices: Vec<rustls_slice_bytes> = bufs
+            .iter()
+            .filter(|b| !b.is_empty())
```
tgeoghegan, line 163: Why filter out the empty buffers? Will `writev(2)` (which the client is heavily encouraged to use in their callback) care if one of the `struct iovec`s is empty?

tgeoghegan (approved):
LGTM, modulo some doccomment improvements.

I saw the discussion about how using `rustls_byte_slice` would break the ABI compatibility with `struct iovec`, but the current implementation has the downside of shutting out Windows clients, even though `std::io::IoSlice` has gone to the trouble of being compatible with Windows `WSABUF`. I'll admit I'd never heard of `WSABUF` before reading that rust doc, but it looks like it goes with [`wsasend`](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasend) which is analogous to `writev(2)`. So could `rustls_connection_write_tls_vectored` also transparently support vectored IO on Windows, with doccomment hints to cast to `WSABUF` and use `wsasend()`?

jsha:
Yep, we should be able to have ABI compatibility with WSABUF as well, but I don't have a solid Windows setup at the moment and didn't want to commit to supporting that (and document it) without having a chance to test it. I anticipate that as a good followup (if write_vectored support in rustls sticks around).

gvanem:
When I build now, I get:
```
   Compiling rustls v0.19.1=======>      ] 19/24: sct, webpki, ring
error[E0432]: unresolved imports `crate::io::rustls_write_vectored_callback`, `crate::io::VectoredCallbackWriter`, `crat
e::io::VectoredWriteCallback`
 --> src\connection.rs:8:5
  |
8 |     rustls_write_vectored_callback, CallbackReader, CallbackWriter, ReadCallback,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |     |
  |     no `rustls_write_vectored_callback` in `io`
  |     help: a similar name exists in the module: `rustls_write_callback`
9 |     VectoredCallbackWriter, VectoredWriteCallback, WriteCallback,
  |     ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^ no `VectoredWriteCallback` in `io`
  |     |
  |     no `VectoredCallbackWriter` in `io`
```
What!?

tgeoghegan:
> When I build now, I get:
> 
> ```
>    Compiling rustls v0.19.1=======>      ] 19/24: sct, webpki, ring
> error[E0432]: unresolved imports `crate::io::rustls_write_vectored_callback`, `crate::io::VectoredCallbackWriter`, `crat
> e::io::VectoredWriteCallback`
>  --> src\connection.rs:8:5
>   |
> 8 |     rustls_write_vectored_callback, CallbackReader, CallbackWriter, ReadCallback,
>   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   |     |
>   |     no `rustls_write_vectored_callback` in `io`
>   |     help: a similar name exists in the module: `rustls_write_callback`
> 9 |     VectoredCallbackWriter, VectoredWriteCallback, WriteCallback,
>   |     ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^ no `VectoredWriteCallback` in `io`
>   |     |
>   |     no `VectoredCallbackWriter` in `io`
> ```
> 
> What!?

I filed #119 to track this -- sorry for the disruption.

PR #114, opened 2021-06-22
Switch to optimized builds by default.

jsha:
This avoids having people accidentally do debug builds, which are extremely slow in Rust.


icing (approved):
Yes, I think that will do nicely. Thanks!

PR #116, opened 2021-06-25
Rename library to rustls-ffi

jsha:
Interested in feedback on the naming here. The crate name will be rustls-ffi. For the generated artifacts, my original intent was to use `rustls.h` and `librustls.a` (since the "FFI" part is clear from the fact that these are not rust libs). But because we also generate an rlib (thanks to @sagebind's #41), naming the lib output `rustls` generates a conflict with the actual rustls crate.

So currently I have `rustls.h` and `librustls_ffi.a`, which is somewhat awkward. I don't love the underscore, and to users of the library from C this is not "FFI" so it's odd to put that in the name. Thoughts? @icing @djc @tgeoghegan


sagebind:
One option might be to just let the output be `librustls_ffi.a` and then copy/rename it to `librustls.a` in the Makefile, since that's the interface rustls-ffi will generally be built with.

A quick survey of [cbindgen users](https://crates.io/crates/cbindgen/reverse_dependencies) might yield some creative ideas. A fair portion of crates I looked at simply left the suffix as `_ffi.a`. 🤷

Makefile
```diff
@@ -29,27 +29,27 @@ test: all
 target:
 	mkdir -p $@
 
-src/crustls.h: src/*.rs cbindgen.toml
+src/rustls.h: src/*.rs cbindgen.toml
 	cargo check
 	cbindgen --lang C > $@
 
-target/$(PROFILE)/libcrustls.a: src/*.rs Cargo.toml
+target/$(PROFILE)/librustls.a: src/*.rs Cargo.toml
```
tgeoghegan, line 36:
```suggestion
target/$(PROFILE)/librustls_ffi.a: src/*.rs Cargo.toml
```
I think, judging by the build failures on this PR

tgeoghegan (changes requested):

djc:
I don't have a particularly informed opinion here, looking at other cbindgen users seems like a fine approach.

jsha:
> One option might be to just let the output be librustls_ffi.a and then copy/rename it to librustls.a in the Makefile, since that's the interface rustls-ffi will generally be built with.

Good plan! So I've settled on `rustls.h` and `librustls.a`, with code in the Makefile to do the necessary naming to avoid conflicts. I also added some symlinks from the old `crustls.h` and `libcrustls.a` names to make the transition to the new names a little easier.

PR #120, opened 2021-06-25
Fix msvc build

### LINKED ISSUE #119: Build broken on Windows by #112
When I build now, I get:
```
   Compiling rustls v0.19.1=======>      ] 19/24: sct, webpki, ring
error[E0432]: unresolved imports `crate::io::rustls_write_vectored_callback`, `crate::io::VectoredCallbackWriter`, `crat
e::io::VectoredWriteCallback`
 --> src\connection.rs:8:5
  |
8 |     rustls_write_vectored_callback, CallbackReader, CallbackWriter, ReadCallback,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |     |
  |     no `rustls_write_vectored_callback` in `io`
  |     help: a similar name exists in the module: `rustls_write_callback`
9 |     VectoredCallbackWriter, VectoredWriteCallback, WriteCallback,
  |     ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^ no `VectoredWriteCallback` in `io`
  |     |
  |     no `VectoredCallbackWriter` in `io`
```
What!?

_Originally posted by @gvanem in https://github.com/rustls/rustls-ffi/issues/112#issuecomment-868299417_

tgeoghegan:
PR#112 broke the Windows build because `rustls_write_vectored_callback`
is hidden behind `#[cfg(unix)]`. Rust's `std::io::IoSlice` should be
compatible with `wsasend` and `WSABUF` just as it is with `write(2)` and
`struct iovec`, so it should be safe to include this functionality in
Windows builds.

Additionally, this commit configures GitHub Actions to at least build on
Windows.

Closes #119

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,6 +16,14 @@ jobs:
           persist-credentials: false
       - run: make CC=${{ matrix.cc }} PROFILE=release test
 
+  test-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+      - run: make -f Makefile.Windows
+
   ensure-header-updated:
     runs-on: ubuntu-latest
     steps:
--- a/Makefile.Windows
+++ b/Makefile.Windows
@@ -4,13 +4,13 @@
 #
 export CL=
 
-CRUSTLS_LIB = target/i686-pc-windows-msvc/release/crustls.lib
+CRUSTLS_LIB = target/release/crustls.lib
 
 USE_CLANG_CL ?= 1
 
 CFLAGS     = -nologo -MD -Zi -W3 -O2 -I. -Dssize_t=int -D_CRT_SECURE_NO_WARNINGS
 LDFLAGS    = -nologo -incremental:no
-CARGOFLAGS = --color never --release --target i686-pc-windows-msvc
+CARGOFLAGS = --color never --release
 
 ifeq ($(USE_CLANG_CL),1)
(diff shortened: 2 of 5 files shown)


jsha:
Looks good to me! Will send a formal GH approval when I'm back at my laptop.

tgeoghegan:
@jsha `test/ensure-header-updated (push)` is failing and I'm not sure why -- the header *should* change since I updated some doccomment.
@gvanem One change I want to highlight here is that I removed the `--target i686-pc-windows-msvc` rustc flag from `Makefile.windows`, and similarly changed the path where the Makefile looks for `libcrustls.lib`. I'm not sure why we would want to force i686 builds. Using the default `cargo` toolchain (more commonly `x86_64-pc-windows-msvc` for modern Windows users) seems better?

jsha:
The header needed updating; I went ahead and pushed that change.

tgeoghegan:
> The header needed updating; I went ahead and pushed that change.

I understand now -- I didn't realize committers had to check in the updated header manually.

jsha:
> I understand now -- I didn't realize committers had to check in the updated header manually.

Yeah. I originally had this as generated at build time, but since the header is (currently) the main source of documentation for the C side, I wanted to have an artifact in the repo that curious C coders could browse through without having to check out and build the whole thing.

jsha (approved):

jsha:
Thanks for the report @gvanem , and for the quick fix @tgeoghegan

Issue #123, opened 2021-06-28
client certificate in outgoing connections


icing:
Thinking about adding support in `rustls-ffi` for client configs. The `rustls` functions are there, the question is how to make them concise int the `ffi` part:

* ServerConfigs work on `CertifiedKey`s and `ffi` has functions to load these.
* a `CertifiedKey` has a `Vec<Certificate>` and a `SigningKey`
* ClientConfig's `set_single_client_cert()` gets a `Vec<Certificate>` and `PrivateKey`

I can see no way to downgrade a `SigningKey` to a `PrivateKey`. So, if applications load a `CertifiedKey` they seem to be unable to pass that to `set_single_client_cert()`. 

Correct?

djc:
`PrivateKey` is a thin wrapper over `pub Vec<u8>`, so it seems like we could easily add some rustls-ffi API to create one?

icing:
The question is how rustls-ffi wants to let the API user load certificate/key pairs and where it can use them.

Atm we load a `rustls_certified_key` and use that in `rustls_server_config_builder_set_certified_keys` and `rustls_client_hello_select_certified_key`.

I would like to use that also in a new `rustls_client_config_ set_single_client_cert`, but than the rust implemenation would need to convert the `SigningKey`in `rustls_certified_key` to a `PrivateKey` and I do not know how to do that.

Can you add such a method to rustls-ffi?

djc:
In general, a `SigningKey` cannot be converted into a `PrivateKey`; that conversion really doesn't make sense in terms of the type definitions. We could potentially change rustls so that it takes a `SigningKey` directly for `with_single_cert()` on both the client and server side, I'd have to see how that works out. Do you think that would help?

icing:
Yes, then we can work with `rusts_certified_key` in those places. That would be good.

djc:
Hmm, I don't think it makes sense to do that. The APIs here have very different levels of visibility. The `ConfigWantsServerCert::with_single_cert()` API is called by pretty much everyone wants to run a server, so it makes sense for it to get a high-level, "shallow" type that is easy to construct and pass in. The inside of the server cert resolver is a much more rarely customization point.

I think the core idea here is that `PrivateKey` can be easily converted to a `CertifiedKey`, but not the other way around, and rustls-ffi should work with that. I suppose rustls-ffi would have to have its own `ResolvesClientCert` implementation similar to how it has its own `ResolvesServerCert` impl.

[...]

PR #128, opened 2021-07-05
Initial take on configuring client authentication certificates

djc:
This pretty much adapts the server-side support to the client.

This is the Rust side of #123. @icing can you work with this?

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,19 +1,19 @@
-use std::ffi::OsStr;
+use std::convert::TryInto;
+use std::ffi::{CStr, OsStr};
 use std::fs::File;
 use std::io::BufReader;
 use std::slice;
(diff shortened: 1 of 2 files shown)


icing:
Looks good. Will start a branch to test it.👍🏻

djc:
Feel free to just reuse this branch.

icing:
Used this PR as patch against `main` and implemented machine certificate configuration in `mod_tls`,
so the server can authenticate against proxied backends via client certificates. 

Code and test cases in https://github.com/abetterinternet/mod_tls/tree/icing/machine-certs

Will merge this when rustls-ffi make a release with this feature. Many thanks!

icing (approved):
Works well in my mod_tls adaption and test suite.

PR #137, opened 2021-09-16
Fix typo in actual source code

djc:
I was a bit too fast in merging #136, which fixes a typo in generated header code. In my defense, the generated code lives in `src`, which seems like a choice to maybe reconsider. (Maybe we should not keep the header file in the repo at all -- I also got confused by `make clean` not doing what I expected in this regard.)


kevinburke:
Gah, you're right. Sorry about that. I'll see if I can submit a PR that will fail the build if we have more issues like this.

kevinburke:
Ah, never mind, the check already exists.

tgeoghegan (approved):

PR #138, opened 2021-09-16
README: add note about TLS certificate verification

### LINKED ISSUE #135: Document default certificate behavior
If I try to just read and write TLS using certs that are descended from certs in my system's trust store, should they work without any additional configuration on my part? Or do I need to write some code in order to load the system certs?

I've read the documentation, the curl code in vtls/rustls.c and tests/*.c and I haven't found anything related to loading the system certs - there are examples to load from a file but not load the default set.

kevinburke:
Fixes #135.


djc (approved):

djc:
LGTM, thanks!

PR #141, opened 2021-09-18
src/client.rs: use correct name for rustls_connection_free function

kevinburke:
The documentation for `new` referred to a nonexistent method.


djc (approved):

djc:
Thanks!

PR #145, opened 2021-09-19
error.rs: better documentation of rustls_io_result

### LINKED ISSUE #144: `rustls_error` equivalent for `rustls_io_result`
I can call `rustls_error` on a `rustls_result` to get a string explaining it, but I can't do the same for a `rustls_io_result`. 

I _believe_ that a `rustls_io_result` is supposed to represent values like `EAGAIN`, `EWOULDBLOCK`, `EINTR` etc, but the existing documentation is not really clear about that:

```rust
/// A return value for a function that may return either success (0) or a
/// non-zero value representing an error.
```

kevinburke:
Fixes #144.

### DIFF
--- a/src/crustls.h
+++ b/src/crustls.h
@@ -349,7 +349,9 @@ typedef void (*rustls_log_callback)(void *userdata, const struct rustls_log_para
 
 /**
  * A return value for a function that may return either success (0) or a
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Thanks!

PR #147, opened 2021-09-22
cipher: add method to get ciphersuite name

### LINKED ISSUE #143: Get a given ciphersuite as a string
As far as I can tell, it's impossible to translate between a `rustls_supported_ciphersuite` and the actual string value of that ciphersuite. For example if I am in a C program, and rustls has negotiated `TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384` for me, I (or specifically psql) would like to display that to the user but the best I can get right now is a u16, I believe, unless I manually copy all of the constants out from the Rust program.

See also rustls/rustls#822.

kevinburke:
Fixes #143.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ no_log_capture = []
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "=0.20.1", features = [ "dangerous_configuration" ] }
+rustls = { git = "https://github.com/kevinburke/rustls", branch = "test_string", features = [ "dangerous_configuration" ] }
 webpki = "0.22"
 libc = "0.2"
 sct = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ no_log_capture = []
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { git = "https://github.com/kevinburke/rustls", branch = "test_string", features = [ "dangerous_configuration" ] }
+rustls = { version = "=0.20.2", features = [ "dangerous_configuration" ] }
 webpki = "0.22"
 libc = "0.2"
 sct = "0.7"
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -100,30 +100,6 @@ pub extern "C" fn rustls_supported_ciphersuite_get_name(
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use crate::cipher::*;
-    use std::slice;
-    use std::str;
-
-    #[test]
-    fn ciphersuite_get_name() {
-        let suite = rustls_all_ciphersuites_get_entry(0);
-        let s = rustls_supported_ciphersuite_get_name(suite);
-        let want = "TLS13_AES_256_GCM_SHA384";
-        unsafe {
-            let got = str::from_utf8(slice::from_raw_parts(s.data as *const u8, s.len)).unwrap();
-            assert_eq!(want, got)
-        }
-    }
-
-    #[test]
-    fn test_all_ciphersuites_len() {
-        let len = rustls_all_ciphersuites_len();
-        assert!(len > 2);
-    }
-}
-
 /// Return the length of rustls' list of supported cipher suites.
 #[no_mangle]
 pub extern "C" fn rustls_all_ciphersuites_len() -> usize {
@@ -223,6 +199,8 @@ pub static RUSTLS_DEFAULT_CIPHER_SUITES_LEN: usize = unsafe { RUSTLS_DEFAULT_CIP
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::slice;
+    use std::str;
 
     #[test]
     fn all_cipher_suites_arrays() {
@@ -249,6 +227,23 @@ mod tests {
             let ffi_cipher_suite = try_ref_from_ptr!(ffi);
             assert_eq!(original, ffi_cipher_suite);
         }
+    }
+
+    #[test]
+    fn ciphersuite_get_name() {
+        let suite = rustls_all_ciphersuites_get_entry(0);
+        let s = rustls_supported_ciphersuite_get_name(suite);
+        let want = "TLS13_AES_256_GCM_SHA384";
+        unsafe {
+            let got = str::from_utf8(slice::from_raw_parts(s.data as *const u8, s.len)).unwrap();
+            assert_eq!(want, got)
+        }
+    }
+
+    #[test]
+    fn test_all_ciphersuites_len() {
+        let len = rustls_all_ciphersuites_len();
+        assert!(len > 2);
     }
 }
 
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -1,4 +1,5 @@
 use libc::size_t;
+use std::convert::TryFrom;
 use std::io::Cursor;
 use std::ptr::null;
 use std::slice;
@@ -13,7 +14,7 @@ use rustls::{
 use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 
 use crate::error::rustls_result;
-use crate::rslice::rustls_slice_bytes;
+use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
     ffi_panic_boundary, try_mut_from_ptr, try_ref_from_ptr, try_slice, ArcCastPtr, BoxCastPtr,
     CastConstPtr, CastPtr,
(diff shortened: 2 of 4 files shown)


src/cipher.rs
```diff
@@ -86,6 +87,30 @@ pub extern "C" fn rustls_supported_ciphersuite_get_suite(
...
+        if buf.is_null() {
+            return 0;
+        }
+        slice::from_raw_parts_mut(buf as *mut u8, len as usize)
+    };
+    let supported_ciphersuite = try_ref_from_ptr!(supported_ciphersuite);
+    let suite_name: String = format!("{:?}", supported_ciphersuite);
+    let suite_name: &[u8] = suite_name.as_bytes();
+    let len: usize = min(write_buf.len() - 1, suite_name.len());
+    write_buf[..len].copy_from_slice(&suite_name[..len]);
```
djc, line 109: Since there is an implementation of `std::io::Write` for `&mut [u8]` I think we could write format directly into the user-provided buffer with `write_buf.write_fmt(format_args!("{:?}", supported_ciphersuite))`. Do you want to try that? The documentation at https://doc.rust-lang.org/std/io/trait.Write.html#impl-Write-6 seems to indicate this will correctly take care of the situation where the cipher suite name is longer than the provided buffer.

djc (commented):
Looks pretty good!

src/cipher.rs
kevinburke: > I think we could format directly into the user-provided buffer with write_buf.write_fmt(format_args!("{:?}", supported_ciphersuite)). Do you want to try that?

Sure. I copied this implementation from `rustls_version` - should I update that one as well?

djc: Sounds good, but maybe do it in a separate commit?

src/cipher.rs
```diff
@@ -86,6 +87,30 @@ pub extern "C" fn rustls_supported_ciphersuite_get_suite(
...
+    len: size_t,
+) -> size_t {
+    let write_buf: &mut [u8] = unsafe {
+        if buf.is_null() {
+            return 0;
+        }
+        slice::from_raw_parts_mut(buf as *mut u8, len as usize)
+    };
+    let supported_ciphersuite = try_ref_from_ptr!(supported_ciphersuite);
+    let suite_name: String = format!("{:?}", supported_ciphersuite);
```
tgeoghegan, line 106: Unless I'm mistaken, this will end up invoking the the `std::fmt::Debug` implementation on [`SupportedCipherSuite`](https://docs.rs/rustls/0.19.1/rustls/struct.SupportedCipherSuite.html), which includes quite a bit more than the name. That seems a little risky, since [`Debug` implementations are generally understood not to be stable](https://doc.rust-lang.org/std/fmt/trait.Debug.html). If this function is exposed just so you can get a string to print out to logs from a C program, then it's fine, but `rustls-ffi` should make it clear that the string obtained through this function should not be parsed by code because its format could change at any time. I think mentioning that in the doccomment should suffice.

djc: Oh, I missed that! What you're looking for is the `CipherSuite`, not the `SupportedCipherSuite`. The former should be easily accessible from the latter, though.

tgeoghegan: True, but I think my point about `std::fmt::Debug` being unstable still applies.

kevinburke: I think that's OK, for the use case that I need it for, which is compatibility with the existing psql implementation - the OpenSSL one shows the cipher you use to negotiate TLS, so I'm hoping to show something similar.

tgeoghegan: Another nit: can we get the cipher suite name as a `&'static str` and avoid allocating a `String`?

djc: Not right now, but see https://github.com/rustls/rustls/pull/831.

kevinburke:
I'd like to try to write a test for this change that would let me/us actually verify it does what it says and maybe give us a hook to writing more tests in the future. Roughly, I want to:

- Malloc a buffer in Rust
- Convert it to a `* c_char` and pass to the `get_ciphersuite_name` function
- Verify that the correct data was written to the buffer.

Then do the same with a buffer that is shorter than the data I plan to write into it, and then verify the behavior is what we expect in that case.

tgeoghegan:
Elsewhere, we wrap objects handed off to C in `Box` or `Arc`, which forces them onto the heap (and then requires a call to some `*_free` function). The docs for [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html) say it's analogous to Rust `String` (with `&CStr` being analogous to `&str`) so I suspect it is also heap allocated, and your approach should be sound.

jsha:
That approach should be sound, but in general we've avoided using NUL-terminated strings in this library because they are a common source of bugs. Instead we try to give a pointer and length (and also guarantee no internal NULs, for convenience).

> Another nit: can we get the cipher suite name as a &'static str and avoid allocating a String?

I like this idea, and I think it's more than a nit. It affects the API. If we can give the caller a pointer to a &' static str, we don't have to provide a corresponding _free() method, because the str lives for the life of the program. *And* the caller doesn't need to provide a buffer. This is much more convenient for the caller.

Since we don't yet have as_str() offered by the rustls library, I think we should fake it for now with a hardcoded map of ciphersuites to string literals. Like so:

```rust
let name = match supported_ciphersuite {
  TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 => "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
  ...
};
```

The list of all such ciphersuites is at https://docs.rs/rustls/0.20.0/src/rustls/suites.rs.html#117-135. We could then return that as a rustls_str, similar to how rustls_log_level_str does it.

Glad to see you getting to know Rust better, @kevinburke ! Thanks for the contribution.

djc:
I'm kind of on the fence if it's worth going through the trouble of duplicating the cipher suite names here, since I suspect the as_str() API to land in a few days (and we can likely follow up with a 0.20.1 release soonish).

kevinburke:
I think it makes sense to wait for rustls/rustls#831 so we can get the better interface, but unfortunately at that point we're blocked on #88, so this might take a while to actually land.

djc:
I hope to get #88 to at least compile today or tomorrow (I left it to wait until Rustls 0.20 was released, which finally happened on Sunday), so hopefully it won't be too long.

src/cipher.rs
kevinburke: I've now made this change

kevinburke:
I've updated this PR to add a test - the previous implementation had an error (it was calling debug on the SupportedCipherSuite and not calling `.suite()`, as a result we were writing the struct details to the buffer in addition to the name. Fortunately the test caught it.

kevinburke:
Note the tests currently fail because I wasn't sure how to handle the case where the input buffer is shorter than the amount of bytes we want to write.

kevinburke:
After some discussion on Slack the conclusion is that this should wait until rustls 0.20.1 has been released.

jsha:
To clarify: I think this should return a `rustls_str`. To make that happen we need one of two things:

 - rustls 0.20.1 released, or
 - write a big match statement mapping from cipher suites to string literals.

I actually favor the latter approach, if you're up for it. I think it's not too complicated. You would take the list of names from https://docs.rs/rustls/0.20.0/src/rustls/suites.rs.html#117-135, e.g. TLS13_AES_256_GCM_SHA384, and transform them (e.g. with sed, or regex in your favorite editor) such that you could do:

```rust
match supported_ciphersuite.suite() {
  CipherSuite::TLS13_AES_128_GCM_SHA256 => "TLS13_AES_128_GCM_SHA256",
  ...
}
```

I would recommend doing a further transform after that to replace "TLS13" on the right hand side with "TLS", to match the [IANA-assigned names](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml):

```rust
match supported_ciphersuite.suite() {
  CipherSuite::TLS13_AES_128_GCM_SHA256 => "TLS_AES_128_GCM_SHA256",
  ...
}
```

djc:
Why do you favor the latter approach?

jsha:
> Why do you favor the latter approach?

Mainly because I wasn't sure how long rustls 0.20.1 would take and wanted to get this change unblocked. Now that 0.20.1 is out, we can go ahead and do this the right way!

kevinburke:
I'll take another pass at this now

Cargo.toml
```diff
@@ -21,7 +21,7 @@ no_log_capture = []
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "=0.20.1", features = [ "dangerous_configuration" ] }
+rustls = { git = "https://github.com/kevinburke/rustls", branch = "test_string", features = [ "dangerous_configuration" ] }
```
kevinburke, line 24: Note this needs to be reverted before merge.

jsha (approved):
This looks good, thanks! We'll merge once 0.20.2 is released.

jsha:
0.20.2 is released, this can now be updated.

jsha (approved):
Thanks!

PR #149, opened 2021-09-29
lib: expose rustls_version as a rustls_str

### LINKED ISSUE #148: CARGO_PKG_VERSION: move to build.rs
Currently we load CARGO_PKG_VERSION by calling `env!` at runtime. However, from @jsha in Slack, it sounds like we could hardcode it.

> The trick is to use build.rs, which I was reluctant to do previously. But now we have a build.rs for other reasons, so it would be fine to do that there.
> Alternatively, we could just hardcode the CARGO_PKG_VERSION and make sure to update it in two places whenever we bump the version number.

The advantage to doing this would be that we could construct the version as a static string, which would avoid the need to allocate memory at runtime.

kevinburke:
This simplifies the work of retrieving the rustls-ffi version in C
code, since you don't need to bring your own buffer and hope it is big
enough. Rust can return you a string that is compiled into the binary
and lives for the duration of the program.

To accomplish this, at build time, we write constants.rs with the
cargo version and the package version.

This also helps us achieve our goals of memory safety, by making
data available to C that is not vulnerable to use-after-free issues.

Fixes #148.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,1 +1,0 @@
-pub const RUSTLS_FFI_VERSION: &'static str = "crustls/0.7.1/rustls/0.19.0";
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,0 +1,1 @@
+pub const RUSTLS_FFI_VERSION: &'static str = "crustls/0.7.1/rustls/0.19.0";
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,6 @@ use std::thread::AccessError;
 pub mod cipher;
 pub mod client;
 pub mod connection;
-mod constants;
 pub mod enums;
 mod error;
 pub mod io;
@@ -24,6 +23,8 @@ pub use error::rustls_result;
 
 use crate::log::rustls_log_callback;
 use crate::panic::PanicOrDefault;
+
+include!(concat!(env!("OUT_DIR"), "/version.rs"));
 
 // For C callbacks, we need to offer a `void *userdata` parameter, so the
 // application can associate callbacks with particular pieces of state. We
@@ -275,9 +276,6 @@ mod tests {
     }
 }
 
-// Keep in sync with Cargo.toml.
-const RUSTLS_CRATE_VERSION: &str = "0.20.0";
-
 /// CastPtr represents the relationship between a snake case type (like rustls_client_config)
 /// and the corresponding Rust type (like ClientConfig). For each matched pair of types, there
 /// should be an `impl CastPtr for foo_bar { RustTy = FooBar }`.
@@ -495,14 +493,14 @@ macro_rules! try_callback {
 /// and NUL terminated. Returns the number of bytes written before the NUL.
 #[no_mangle]
 pub extern "C" fn rustls_version() -> rustls_str<'static> {
-    return rustls_str::from_str_unchecked(crate::constants::RUSTLS_FFI_VERSION);
+    return rustls_str::from_str_unchecked(RUSTLS_FFI_VERSION);
 }
 
 #[test]
 fn test_rustls_version() {
     // very rough check that the version number is being interpolated into the
     // variable
-    assert!(crate::constants::RUSTLS_FFI_VERSION.contains("/0."));
+    assert!(RUSTLS_FFI_VERSION.contains("/0."));
     let vsn = rustls_version();
     assert!(vsn.len > 4)
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,14 +1,16 @@
 #![crate_type = "staticlib"]
 #![allow(non_camel_case_types)]
-use libc::{c_char, c_void, size_t};
+use crate::rslice::rustls_str;
+use libc::{c_void, size_t};
 use std::cell::RefCell;
+use std::mem;
 use std::sync::Arc;
-use std::{cmp::min, thread::AccessError};
-use std::{mem, slice};
+use std::thread::AccessError;
 
 pub mod cipher;
 pub mod client;
 pub mod connection;
+mod constants;
 pub mod enums;
 mod error;
 pub mod io;
@@ -492,23 +494,15 @@ macro_rules! try_callback {
 /// provided buffer, up to a max of `len` bytes. Output is UTF-8 encoded
 /// and NUL terminated. Returns the number of bytes written before the NUL.
 #[no_mangle]
-pub extern "C" fn rustls_version(buf: *mut c_char, len: size_t) -> size_t {
-    ffi_panic_boundary! {
-        let write_buf: &mut [u8] = unsafe {
-            if buf.is_null() {
-                return 0;
-            }
-            slice::from_raw_parts_mut(buf as *mut u8, len as usize)
-        };
-        let version: String = format!(
-            "crustls/{}/rustls/{}",
-            env!("CARGO_PKG_VERSION"),
-            RUSTLS_CRATE_VERSION,
-        );
-        let version: &[u8] = version.as_bytes();
-        let len: usize = min(write_buf.len() - 1, version.len());
-        write_buf[..len].copy_from_slice(&version[..len]);
-        write_buf[len] = 0;
-        len
-    }
+pub extern "C" fn rustls_version() -> rustls_str<'static> {
+    return rustls_str::from_str_unchecked(crate::constants::RUSTLS_FFI_VERSION);
+}
+
+#[test]
+fn test_rustls_version() {
+    // very rough check that the version number is being interpolated into the
+    // variable
+    assert!(crate::constants::RUSTLS_FFI_VERSION.contains("/0."));
+    let vsn = rustls_version();
+    assert!(vsn.len > 4)
 }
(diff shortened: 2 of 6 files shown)


src/crustls.h
```diff
@@ -232,6 +232,24 @@ typedef struct rustls_slice_str rustls_slice_str;
...
 
+/**
+ * A read-only view on a Rust `&str`. The contents are guaranteed to be valid
+ * UTF-8. As an additional guarantee on top of Rust's normal UTF-8 guarantee,
+ * a `rustls_str` is guaranteed not to contain internal NUL bytes, so it is
+ * safe to interpolate into a C string or compare using strncmp. Keep in mind
+ * that it is not NUL-terminated.
+ *
+ * The memory exposed is available as specified by the function
+ * using this in its signature. For instance, when this is a parameter to a
```
kevinburke, line 243: > The memory exposed is available as specified by the function using this in its signature.

To be honest, I'm not sure how to interpret this if I'm calling this from C code.

src/lib.rs
```diff
@@ -383,25 +382,17 @@ macro_rules! try_callback {
...
-    }
+pub extern "C" fn rustls_version() -> rustls_str<'static> {
+    return rustls_str::from_str_unchecked(crate::constants::RUSTLS_FFI_VERSION);
+}
+
+#[test]
+fn test_rustls_version() {
+    // very rough check that the version number is being interpolated into the 
+    // variable
+    assert!(crate::constants::RUSTLS_FFI_VERSION.contains("/0."));
```
kevinburke, line 393: We could also e.g. write CARGO_PKG_VERSION to constants.rs and then check that FFI_VERSION contains CARGO_PKG_VERSION.

src/constants.rs
```diff
@@ -0,0 +1 @@
+pub const RUSTLS_FFI_VERSION: &'static str = "crustls/0.7.1/rustls/0.19.0";
```
kevinburke, line 1: Not sure if this file should be checked in or not.

djc:
I wonder if we should just not expose `RUSTLS_FFI_VERSION` as a complex string with two version numbers and names in there. Instead, we could define `RUSTLS_VERSION` as a `&'static str` and add a getter that returns the `rustls_str` for it, and a `RUSTLS_FFI_VERSION` as a `&static str` and add a getter that returns the `rustls_str` for it (neither of these getters would need to allocate or accept a user-defined allocation pointer). That way, we get to expose a simpler API that gives the caller access to both version numbers, and the caller can concatenate these versions as it sees fit rather than being locked into a somewhat arbitrary format defined in the FFI bindings.

jsha:
> Instead, we could define RUSTLS_VERSION as a &'static str and add a getter that returns the rustls_str for it, and a RUSTLS_FFI_VERSION as a &static str and add a getter that returns the rustls_str for it (neither of these getters would need to allocate or accept a user-defined allocation pointer). That way, we get to expose a simpler API that gives the caller access to both version numbers, and the caller can concatenate these versions as it sees fit rather than being locked into a somewhat arbitrary format defined in the FFI bindings.

This is a good point. One could take it a step further and define RUSTLS_VERSION_MAJOR, RUSTLS_VERSION_MINOR, RUSTLS_VERSION_PATH, RUSTLS_FFI_VERSION_MAJOR, etc. If we define these as ints or #defines, that would allow users to do semver range checking without trying to parse a string, *and* format a version string however they want. I think the best thing would be to offer all of these constants, and also offer a getter for the current arbitrarily-formatted version string, for the common case when software just wants to show the user what version is in use, and isn't picky about the format.

icing:
Just a reminder, if you `#define` these constants in the exported header files, they will not change when an application links the library. To log the version linked, getting a string formatted by the library is still useful.

djc:
We could define a `struct version` and add `rustls_version` and `rustls_ffi_version` functions that return it.

I still don't see much utility in the formatted string, and given constraints on backwards compatibility I'd aim for a more minimal public API.

kevinburke:
I'm trying to synthesize the comments but I'm not sure that there's a consensus. A few different ideas

a) Return some sort of `{major: 0, minor: 9, patch: 2}` struct with the right values for both rustls_version and rustls_ffi_version
b) Return static strings from rustls_version and rustls_ffi_version
c) like (b) but also add rustls_version_major(), rustls_version_patch() etc. which returns you an integer. This seems like a lot

I'm indifferent between a and b honestly, maybe a silght preference for strings, to the extent that integers make it easier to do things like "if the version is greater than X, do Y" in your code. I would hope people would stay on the latest library, and maybe strings - which you'd need to print the version as part of a `-V` flag - would be a better fit.

jsha:
> We could define a struct version and add rustls_version and rustls_ffi_version functions that return it.
> I still don't see much utility in the formatted string, and given constraints on backwards compatibility I'd aim for a more minimal public API

The struct seems nice, though it does seem like a bigger public API than "here's a string." Also: the struct would need to have at least one string field, since semver minor versions can include a prerelease string tag. And it would be nice to have a way for the version in non-release builds to express things like "this was built off commit X, from a dirty source tree, at 2020-10-02." We have that flexibility if the only interface we express is a string. If we expose a struct, adding fields to it would be an ABI-breaking change, so we'd need to make sure to get it right from the start (or be willing to make breaking changes, though of course we want to minimize those).

jsha:
As a point of reference, [libcurl has `char *curl_version`, which says](https://curl.se/libcurl/c/curl_version.html):

> We recommend using curl_version_info instead! 

[curl_version_info says](https://curl.se/libcurl/c/curl_version_info.html):

```
curl_version_info_data *curl_version_info( CURLversion age); 

typedef struct {
  CURLversion age;          /* see description below */
 
  const char *version;      /* human readable string */
  unsigned int version_num; /* numeric representation */
  const char *host;         /* human readable string */
  int features;             /* bitmask, see below */
  char *ssl_version;        /* human readable string */
  long ssl_version_num;     /* not used, always zero */
  const char *libz_version; /* human readable string */
  const char * const *protocols; /* protocols */
 
  /* when 'age' is CURLVERSION_SECOND or higher, the members below exist */
  const char *ares;         /* human readable string */
  int ares_num;             /* number */
 
  /* when 'age' is CURLVERSION_THIRD or higher, the members below exist */
  const char *libidn;       /* human readable string */
 
  /* when 'age' is CURLVERSION_FOURTH or higher (>= 7.16.1), the members
     below exist */
  int iconv_ver_num;       /* '_libiconv_version' if iconv support enabled */
 
  const char *libssh_version; /* human readable string */
 
  /* when 'age' is CURLVERSION_FIFTH or higher (>= 7.57.0), the members
     below exist */
  unsigned int brotli_ver_num; /* Numeric Brotli version
                                  (MAJOR << 24) | (MINOR << 12) | PATCH */
  const char *brotli_version; /* human readable string. */
 
  /* when 'age' is CURLVERSION_SIXTH or higher (>= 7.66.0), the members
     below exist */
  unsigned int nghttp2_ver_num; /* Numeric nghttp2 version
                                   (MAJOR << 16) | (MINOR << 8) | PATCH */
  const char *nghttp2_version; /* human readable string. */
 
  const char *quic_version;    /* human readable quic (+ HTTP/3) library +
                                  version or NULL */
 
  /* when 'age' is CURLVERSION_SEVENTH or higher (>= 7.70.0), the members
     below exist */
  const char *cainfo;          /* the built-in default CURLOPT_CAINFO, might
                                  be NULL */
  const char *capath;          /* the built-in default CURLOPT_CAPATH, might
                                  be NULL */
  /* when 'age' is CURLVERSION_EIGHTH or higher (>= 7.71.0), the members
     below exist */
  unsigned int zstd_ver_num; /* Numeric Zstd version
                                  (MAJOR << 24) | (MINOR << 12) | PATCH */
  const char *zstd_version; /* human readable string. */
  /* when 'age' is CURLVERSION_NINTH or higher (>= 7.75.0), the members
     below exist */
  const char *hyper_version; /* human readable string. */
  /* when 'age' is CURLVERSION_TENTH or higher (>= 7.77.0), the members
     below exist */
  const char *gsasl_version; /* human readable string. */
} curl_version_info_data;
```

djc:
> The struct seems nice, though it does seem like a bigger public API than "here's a string." Also: the struct would need to have at least one string field, since semver minor versions can include a prerelease string tag. And it would be nice to have a way for the version in non-release builds to express things like "this was built off commit X, from a dirty source tree, at 2020-10-02." We have that flexibility if the only interface we express is a string. If we expose a struct, adding fields to it would be an ABI-breaking change, so we'd need to make sure to get it right from the start (or be willing to make breaking changes, though of course we want to minimize those).

Fair enough.

kevinburke:
If I am reading these comments correctly it sounds like the consensus is for a string, for now? This is what's currently implemented in the patch, maybe you can take another look?

jsha (approved):
Yes, you've correctly grokked the consensus - sorry for not concluding it more clearly.

I pushed some tweaks - in particular a `build.rs` is supposed to emit files only into OUT_DIR, so I did that - and I removed the checked-in constants.rs.

I also removed the `pub` on the constant, to make clear that it's not part of the external interface (which it can't be, because it's a `&str`, which is not `extern "C"`

PR #153, opened 2021-10-16
Fix client.c and server.c.

jsha:
Some issues were introduced in v0.7.0.
If there was plaintext left in the buffers when the connection closed,
the client would call select with no FDs, and hang forever. In this
change, if we reach that state, we drain the remaining plaintext and
then write it stdout and exit.

Ensure copy_plaintext_to_buffer copies all available plaintext, even if
that requires growing the buffer multiple times.

On the server side, send larger responses, and exit the loop when
neither read nor write is wanted.

Remove unused copy_plaintext_to_stdout.

### DIFF
--- a/tests/server.c
+++ b/tests/server.c
@@ -144,7 +144,7 @@ send_response(struct conndata *conn)
   struct rustls_connection *rconn = conn->rconn;
   const char *prefix = "HTTP/1.1 200 OK\r\nContent-Length:";
   const int body_size = 10000;
-  const int response_size = strlen(prefix) + 10 + body_size;
+  const int response_size = strlen(prefix) + 15 + body_size;
   char *response = malloc(response_size);
   size_t n;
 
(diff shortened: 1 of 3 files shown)


tests/server.c
```diff
@@ -142,8 +142,17 @@ enum crustls_demo_result
 send_response(struct conndata *conn)
 {
   struct rustls_connection *rconn = conn->rconn;
-  const char *response = "HTTP/1.1 200 OK\r\nContent-Length: 6\r\n\r\nhello\n";
+  const char *prefix = "HTTP/1.1 200 OK\r\nContent-Length:";
+  const int body_size = 10000;
+  const int response_size = strlen(prefix) + 20 + body_size;
```
tgeoghegan, line 147: Why 20? It seems like you need 10 bytes for ` 10000\r\n\r\n`?

tests/server.c
```diff
@@ -142,8 +142,17 @@ enum crustls_demo_result
...
+  char *response = malloc(response_size);
```
tgeoghegan, line 148: Should check whether `malloc` fails.

PR #154, opened 2021-10-16
Switch back to a setter-based config builder object

jsha:
In #88 we introduced a typestate-like set of builders in rustls-ffi, mimicking rustls' new typestate config builders.

This is an attempt to make rustls' config builder work using the setter-based config builder interface we were using in rustls-ffi 0.7. I think it winds up being much easier to use.


src/client.rs
```diff
@@ -298,15 +322,13 @@ pub extern "C" fn rustls_client_config_builder_dangerous_set_certificate_verifie
...
     roots: *const rustls_root_cert_store,
-    builder: *mut *mut rustls_client_config_builder,
 ) -> rustls_result {
     ffi_panic_boundary! {
+        let builder = try_mut_from_ptr!(config_builder);
         let root_store: &RootCertStore = try_ref_from_ptr!(roots);
-        let prev = *BoxCastPtr::to_box(wants_verifier);
-        let config = prev.with_root_certificates(root_store.clone()).with_no_client_auth();
-        BoxCastPtr::set_mut_ptr(builder, config);
+        builder.verifier = Arc::new(rustls::client::WebPkiVerifier::new( root_store.clone(), None));
```
tgeoghegan, line 331:
```suggestion
        builder.verifier = Arc::new(rustls::client::WebPkiVerifier::new(root_store.clone(), None));
```

src/lib.rs
```diff
@@ -311,6 +314,19 @@ pub(crate) trait BoxCastPtr: CastPtr + Sized {
     }
 }
 
+// An implementation of BoxCastPtr means that when we give C code a pointer to the relevant type,
```
tgeoghegan, line 317:
```suggestion
// An implementation of ArcCastPtr means that when we give C code a pointer to the relevant type,
```

src/client.rs
```diff
@@ -31,25 +31,23 @@ use crate::{
...
-    type RustType = ConfigBuilder<ClientConfig, WantsVerifier>;
-}
-
-impl BoxCastPtr for rustls_client_config_builder_wants_verifier {}
-
-pub struct rustls_client_config_builder {
-    _private: [u8; 0],
+pub(crate) struct ClientConfigBuilder {
+    base: rustls::ConfigBuilder<ClientConfig, WantsVerifier>,
+    verifier: Arc<dyn ServerCertVerifier>,
```
tgeoghegan, line 43: nit: instead of providing `NoneVerifier` to act as the default/sentinel value indicating that the client hasn't set a `ServerCertVerifier`, why not use `Option<Arc<dyn ServerCertVerifier>>`, analogously to how you handled `cert_resolver`?

src/client.rs
```diff
@@ -98,14 +119,14 @@ pub extern "C" fn rustls_client_config_builder_new_with_safe_defaults(
 /// `rustls_tls_version` for convenience.
 ///
 /// `versions` will only be used during the call and the application retains
-/// ownership. `len` is the number of consecutive `ui16` pointed to by `versions`.
+/// ownership. `len` is the number of consecutive `u16` pointed to by `versions`.
 #[no_mangle]
-pub extern "C" fn rustls_client_config_builder_new(
+pub extern "C" fn rustls_client_config_builder_new_fancy(
```
tgeoghegan, line 124: It seems weird that some options are set using `rustls_client_config_builder_*` functions but some others can only be set in this function's arguments. If you had something like `rustls_client_config_builder_cipher_suites()` and `rustls_client_config_builder_protocol_versions()`, could this function be deleted?

src/server.rs
```diff
@@ -120,12 +111,12 @@ pub extern "C" fn rustls_server_config_builder_new_with_safe_defaults(
 /// `versions` will only be used during the call and the application retains
 /// ownership. `len` is the number of consecutive `ui16` pointed to by `versions`.
 #[no_mangle]
-pub extern "C" fn rustls_server_config_builder_new(
+pub extern "C" fn rustls_server_config_builder_new_fancy(
```
tgeoghegan, line 114: Same comment as on `rustls_client_config_builder_new_fancy`: I think you could provide functions that let clients set cipher suites and TLS protocol version on any `rustls_server_config` and get rid of this entry point.

tgeoghegan (changes requested):

src/client.rs
jsha: If I do that, then I have to be able to return an error from `rustls_client_config_builder_build` to indicate "hey you never told me how to verify server certificates." That's a possibility, and there's an argument that in the builder pattern, we should defer _all_ errors until the build step. But for now I wanted to hew somewhat to the existing 0.7.0 behavior which is "if you forget to set root certs or a custom verifier, you will get errors on every connection."

src/client.rs
jsha: I've merged main, and tested (locally) a client that only uses one ciphersuite and one TLS version; it doesn't build any smaller. I need to experiment a bit more to see if I can make this work as expected.

src/client.rs
jsha: Correct. `rustls` requires you to either set root certificates or set a custom verifier: https://docs.rs/rustls/0.20.0/rustls/struct.ConfigBuilder.html#impl-3.

src/client.rs
jsha: I haven't found time to figure out how to make the dead code elimination work as expected. I need to try compiling rustls itself with default ciphersuites and with a narrowed set of ciphersuites to see if it works upstream. Meanwhile, though, I'd like to proceed with this API, since I suspect it will be required if we ever want dead code elimination to work.

tgeoghegan (approved):

tgeoghegan:
LGTM! Kinda strange that there's a conflict in `src/crustls.h` without a corresponding conflict in either `client.rs` or `server.rs` though.

PR #158, opened 2021-10-29
all: fix reference to function

kevinburke:
The docs reference rustls_all_ciphersuites_get() which does not exist.
Fix the references.


djc (approved):

jsha:
Thanks!

PR #161, opened 2021-10-29
cipher: add separate method to retrieve default ciphersuites

### LINKED ISSUE #159: Retrieve the default set of ciphersuites
There's a new rustls_server_config_builder_new_with_safe_defaults helper. Unfortunately Postgres has a functionality that allows users to set the min and max TLS versions, so I still need to use rustls_server_config_builder_new, which looks like the only way to set those.

That function requires me to also pass in a list of cipher suites. I can get _all_ of the ciphersuites, but as far as I can tell, there's no way to get the default set of ciphersuites that are used by e.g. the with_safe_defaults function, in C code? 

Would it be possible to expose that default list? For example:

```c
/**
 * Return the length of rustls' list of default cipher suites.
 */
size_t rustls_default_ciphersuites_len(void);

/**
 * Get a pointer to a member of rustls' list of supported cipher suites. This will return non-NULL
 * for i < rustls_default_ciphersuites_len().
 * The returned pointer is valid for the lifetime of the program and may be used directly when
 * building a ClientConfig or ServerConfig.
 */
const struct rustls_supported_ciphersuite *rustls_default_ciphersuites_get_entry(size_t i);
```

kevinburke:
Fixes #159.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -6,7 +6,10 @@ use std::sync::Arc;
 
 use rustls::server::{AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient};
 use rustls::sign::CertifiedKey;
-use rustls::{Certificate, PrivateKey, RootCertStore, SupportedCipherSuite, ALL_CIPHER_SUITES};
+use rustls::{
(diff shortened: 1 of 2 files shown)


jsha (approved):
Looks good, thanks!

PR #164, opened 2021-10-29
client: use DEFAULT_CIPHER_SUITES if none are specified

### LINKED ISSUE #163: rustls_client_config_builder_new: use the default set if cipher_suites is NULL
I'm having trouble building an array of ciphersuites in C because I don't know C very well and keep being told I'm either discarding the 'const' qualifier from the pointer target type or trying to assign to a constant pointer, and so I keep getting in trouble.

Anyway all I am trying to do is use the default set of cipher suites. It would be nice if `rustls_client_config_builder_new` could check for `null` in the cipher_suites argument, and then just use the default set of ciphersuites.

kevinburke:
I'm having trouble figuring out how to pass in a set of cipher suites
using C, and since all I want is the default set, just make it easier
by providing a reasonable default for the NULL case.

Fixes #163.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -10,7 +10,7 @@ use libc::{c_char, size_t};
 use rustls::client::{ResolvesClientCert, ServerCertVerified, ServerCertVerifier};
 use rustls::{
     sign::CertifiedKey, Certificate, ClientConfig, ClientConnection, ProtocolVersion,
-    RootCertStore, SupportedCipherSuite, WantsVerifier, ALL_CIPHER_SUITES,
+    RootCertStore, SupportedCipherSuite, WantsVerifier, ALL_CIPHER_SUITES, DEFAULT_CIPHER_SUITES,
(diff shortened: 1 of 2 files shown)


djc (approved):
This seems like a nice usability tweak!

jsha:
I like this! I would like a little more an explicit option where you could pass something like:

```
rustls_client_config_builder_new_custom(
  RUSTLS_DEFAULT_CIPHER_SUITES,
  RUSTLS_DEFAULT_CIPHER_SUITES_LEN,
  tls_versions,
  tls_versions_len)
```

But I _think_ that would require rustls-ffi to make a copy of `rustls::DEFAULT_CIPHER_SUITES` so that it could present a `#[repr(C)]` view of them, which is a little awkward. I'll think on it a bit more - but if we don't come up with a nicer solution shortly we should go ahead and merge this, which is plenty nice too.

PR #165, opened 2021-11-03
client: use DEFAULT_CIPHER_SUITES if none are specified

### LINKED ISSUE #163: rustls_client_config_builder_new: use the default set if cipher_suites is NULL

kevinburke:
I'm having trouble figuring out how to pass in a set of cipher suites
using C, and since all I want is the default set, just make it easier
by providing a reasonable default for the NULL case.

Fixes #163.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -10,7 +10,7 @@ use libc::{c_char, size_t};
 use rustls::client::{ResolvesClientCert, ServerCertVerified, ServerCertVerifier};
 use rustls::{
     sign::CertifiedKey, Certificate, ClientConfig, ClientConnection, ProtocolVersion,
-    RootCertStore, SupportedCipherSuite, WantsVerifier, ALL_CIPHER_SUITES,
+    RootCertStore, SupportedCipherSuite, WantsVerifier, ALL_CIPHER_SUITES, DEFAULT_CIPHER_SUITES,
 };
 
 use crate::cipher::{rustls_certified_key, rustls_root_cert_store, rustls_supported_ciphersuite};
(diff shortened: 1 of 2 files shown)


jsha:
As I mentioned in https://github.com/rustls/rustls-ffi/issues/203#issuecomment-964350398, I'd like a more explicit way to do this:

```
rustls_client_config_builder_new_custom(
  RUSTLS_DEFAULT_CIPHER_SUITES,
  RUSTLS_DEFAULT_CIPHER_SUITES_LEN,
  tls_versions, // provided by your program
  tls_versions_len,
  &builder);
```

DEFAULT_CIPHER_SUITES should be defined like so:

```
const struct *rustls_supported_cipher_suite[] DEFAULT_CIPHER_SUITES;
const size_t DEFAULT_CIPHER_SUITES;
```

To make this happen we'll need to declare our own copy of DEFAULT_CIPHER_SUITES from rustls, since ours will need to contain pointers to the entries of rustls::DEFAULT_CIPHER_SUITES.

To ensure our copy remains up-to-date, we can have a test that checks the length is the same, and the set of items is the same.

We should document that the order of DEFAULT_CIPHER_SUITES will change between releases.

jsha:
I think this is resolved by #242

PR #167, opened 2021-11-03
all: make the links link-able

kevinburke:
This makes them hyperlinks when you open them in the rustdoc browser,
and removes the warnings when you run `cargo doc --open`.


jsha (approved):
🔗 👍🏻

PR #170, opened 2021-11-04
Remove or-patterns for compat with Rust 1.52

### LINKED ISSUE #169: Windows: error[E0658]
When using a heavily modified [Makefile.Windows](https://github.com/rustls/rustls-ffi/blob/main/Makefile.Windows),
I got this error from `rustc` (?):
```
cargo build --color never --release --target i686-pc-windows-msvc
    Updating crates.io index
  Downloaded libc v0.2.106ining bytes: 211.3 KB
  Downloaded 1 crate (554.8 KB) in 1.43s..
   Compiling cc v1.0.71
   Compiling winapi v0.3.9
   Compiling untrusted v0.7.1
   Compiling spin v0.5.2
   Compiling log v0.4.14
   Compiling libc v0.2.106
   Compiling cfg-if v1.0.0
   Compiling rustls-ffi v0.7.1 (F:\MingW32\src\inet\Crypto\Rustls)
   Compiling base64 v0.13.0               ] 0/24: rustls-ffi(build.rs), log(build.rs), cc, spin, untrusted, libc(b...
   Compiling rustls-pemfile v0.2.1        ] 3/24: rustls-ffi(build.rs), base64, log(build.rs), cc, libc(build.rs),...
   Compiling ring v0.16.20=>             ] 13/24: libc, winapi, log, cc
   Compiling webpki v0.22.0=======>      ] 19/24: ring
   Compiling sct v0.7.0
   Compiling rustls v0.20.0=======>      ] 19/24: webpki, sct, ring
error[E0658]: or-patterns syntax is experimental: rustls-ffi
  --> src\error.rs:51:17
   |
51 | /                 Error::InvalidCertificateData(_)
52 | |                 | Error::InvalidCertificateEncoding
53 | |                 | Error::InvalidCertificateSignature
54 | |                 | Error::InvalidCertificateSignatureType
55 | |                 | Error::InvalidSct(_),
   | |______________________________________^
   |
   = note: see issue #54883 <https://github.com/rust-lang/rust/issues/54883> for more information

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: could not compile `rustls-ffi`
```
What's this? I know nothing about Rust/Cargo. Perhaps my Cargo/rustc is too old?
* `rustc -v   -> rustc 1.52.1 (9bc8c42bb 2021-05-09)`
* `cargo -V  -> cargo 1.52.0 (69767412a 2021-04-21)`

I just wanted to try `rustls_ffi.lib` in libcurl. But that seems impossible now for other reasons too (besides this).

jsha:
Fixes #169

Thanks to @gvanem for reporting.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,6 +2,8 @@ name: test
 permissions:
   contents: read
 on: [push, pull_request]
+
+# Copied from https://github.com/rustls/rustls/blob/ba49f6f0b5704ecc0c7a15e5ab8819bf22edc4cb/.github/workflows/build.yml
 jobs:
   test:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,8 +2,6 @@ name: test
 permissions:
   contents: read
 on: [push, pull_request]
-
-# Copied from https://github.com/rustls/rustls/blob/ba49f6f0b5704ecc0c7a15e5ab8819bf22edc4cb/.github/workflows/build.yml
 jobs:
   test:
     runs-on: ${{ matrix.os }}
--- a/src/error.rs
+++ b/src/error.rs
@@ -47,13 +47,14 @@ impl rustls_result {
     #[no_mangle]
     pub extern "C" fn rustls_result_is_cert_error(result: rustls_result) -> bool {
         match result_to_error(&result) {
-            Either::Error(
-                Error::InvalidCertificateData(_)
-                | Error::InvalidCertificateEncoding
-                | Error::InvalidCertificateSignature
-                | Error::InvalidCertificateSignatureType
-                | Error::InvalidSct(_),
-            ) => true,
+            Either::Error(e) => match e {
+                Error::InvalidCertificateData(_) => true,
+                Error::InvalidCertificateEncoding => true,
+                Error::InvalidCertificateSignature => true,
+                Error::InvalidCertificateSignatureType => true,
+                Error::InvalidSct(_) => true,
+                _ => false,
+            },
             _ => false,
         }
     }


Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Jacob Hoffman-Andrews <github@hoffman-andrews.com>"]
 description = "C-to-rustls bindings"
 edition = "2018"
 links = "rustls_ffi"
+rust-version = "1.52.0"
```
kevinburke, line 8: A little curious that this is named "rust-version" and not "minimum-rust-version" or similar... if I was reading this I might assume it will force my compiler to 1.52.0 or only work with that version or something.

tgeoghegan (commented):
So there's kind of a paradox here. The [`rust-version` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) is only supported by `cargo` 1.56 and up, so this won't actually help with this specific problem, though it will once the whole world rolls past 1.56.

Second, this will help with people doing `cargo build` locally, but what about CI builds? I think they'll still use whatever `cargo` and `rustc` happen to be on the GitHub runners, and last time I checked, that was 1.55. I think we should have CI builds use 1.52 if we want to support the claim that this builds with that version.

[The advice in this fasterthanli.me article](https://fasterthanli.me/articles/my-ideal-rust-workflow) was to use a [`rust-toolchain.toml`](https://rust-lang.github.io/rustup/overrides.html), which AFAICT both works with older toolchains and will do the right thing in CI, because it'll force `cargo` to install and use the toolchain specified by `rust-toolchain.toml`. But I'm curious if there are other tradeoffs between `rust-version` and the toolchain fil.

jsha:
> The rust-version field is only supported by cargo 1.56 and up, so this won't actually help with this specific problem, though it will once the whole world rolls past 1.56.

Yep, indeed - I realized this shortly after pushing and removed this line.

> Second, this will help with people doing cargo build locally, but what about CI builds? 

Yes, I'm working on a change to add more rust versions to CI right now. I had thought it might require some iteration on the yaml but it seems to have worked right off the bat, so I'll merge it into this one.

jsha:
Tests were failing, so I removed the CI updates from this PR to land the smaller bit. I'll submit the test updates separately.

.github/workflows/test.yaml
```diff
@@ -2,6 +2,8 @@ name: test
 permissions:
   contents: read
 on: [push, pull_request]
+
+# Copied from https://github.com/rustls/rustls/blob/ba49f6f0b5704ecc0c7a15e5ab8819bf22edc4cb/.github/workflows/build.yml
```
tgeoghegan, line 6: I'm guessing this comment is left over from the CI changes that were punted out of this PR?

tgeoghegan (approved):
LGTM, minus the stray comment in `.github/workflows/test.yaml`

PR #171, opened 2021-11-04
lib.rs: correct rustls_version docs

### LINKED ISSUE #168: rustls_version has wrong description in header
This is a snippet from the current git main head:
~~~c
/**
 * Write the version of the crustls C bindings and rustls itself into the
 * provided buffer, up to a max of `len` bytes. Output is UTF-8 encoded
 * and NUL terminated. Returns the number of bytes written before the NUL.
 */
struct rustls_str rustls_version(void);
~~~

The description does not match the prototype!

kevinburke:
We changed the behavior of this function to return a static str, but
the docs still referred to the old behavior (write to a buffer). Fix
the documentation, and provide an example of how to use it in
tests/client.c.

Update the version to reflect the new name of the package
("rustls-ffi").

Fixes #168.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -19,7 +19,7 @@ fn main() {
     let pkg_version = env!("CARGO_PKG_VERSION");
     write!(
         &mut f,
-        r#"const RUSTLS_FFI_VERSION: &'static str = "crustls/{}/rustls/{}";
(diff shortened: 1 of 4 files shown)


jsha (approved):
Thanks!

bagder:
:+1:

PR #176, opened 2021-11-06
Switch back to `_get` names on rustls_connection

jsha:
These are the names that were in use for v0.7.2 and prior. I updated
them as part of the update to rustls 0.20.0 to be more similar to
upstream, which removed `_get` from getter methods. However, the `_get`
naming convention is reasonable enough in C, and maintaining the
similarity to upstream is not worth causing an API break for these
functions.


kevinburke:
Need to update the comment in src/cipher.rs that still mentions `rustls_connection_peer_certificate`. Otherwise LGTM.

PR #181, opened 2021-11-06
server.c: switch from inet_aton to inet_addr

### LINKED ISSUE #172: Windows: inet_aton() is missing.
Linking the `target/server.exe` fails due to a missing `inet_aton()` in Winsock2.
I suggest you switch to `inet_addr()` instead. Isn't that available everywhere?
Something like:
```diff
--- a/tests/server.c 2021-11-04 14:06:41
+++ b/tests/server.c 2021-11-05 11:36:19
@@ -315,7 +315,7 @@
   my_addr.sin_family = AF_INET;
   my_addr.sin_addr.s_addr = INADDR_ANY;
   my_addr.sin_port = htons(8443);
-  inet_aton("127.0.0.1", &my_addr.sin_addr);
+  my_addr.sin_addr.s_addr = inet_addr("127.0.0.1");

   if(bind(sockfd, (struct sockaddr *)&my_addr, sizeof(struct sockaddr_in)) ==
```

And BTW, it would be nice to move most of the Winsock *shoe-horn* macros in `test/client.c` + `test/server.c` 
into  `test/common.h`.

jsha:
Fixes build on Windows.

Fixes #172

Thanks for reporting, @gvanem!

### DIFF
--- a/tests/server.c
+++ b/tests/server.c
@@ -343,7 +343,7 @@ main(int argc, const char **argv)
   my_addr.sin_family = AF_INET;
   my_addr.sin_addr.s_addr = INADDR_ANY;
   my_addr.sin_port = htons(8443);
-  inet_aton("127.0.0.1", &my_addr.sin_addr);
+  my_addr.sin_addr.s_addr = inet_addr("127.0.0.1");
 
   if(bind(sockfd, (struct sockaddr *)&my_addr, sizeof(struct sockaddr_in)) ==
      -1) {


gvanem:
Go ahead and merge it. 
And BTW, I see that `Makefile.Windows` is used for CI. But there is no client/server tests in it. 
I could make an updated version.

jsha:
> And BTW, I see that Makefile.Windows is used for CI. But there is no client/server tests in it.
> I could make an updated version.

That would be terrific, thank you!

PR #183, opened 2021-11-06
Quiet the "wait for port" code in test.

jsha:
When the server fails to come up for some reason, we were emitting a lot
of logs, like:

./test.sh: connect: Connection refused
./test.sh: line 6: /dev/tcp/localhost/8443: Connection refused
+ return 1
++ date
+ echo 'Thu Nov  4 16:41:19 UTC 2021 - still trying to connect to localhost:8443'
Thu Nov  4 16:41:19 UTC 2021 - still trying to connect to localhost:8443
+ sleep .1
+ '[' 102 -eq 120 ']'
+ for n in '`seq 1 $max_tries`'
+ port_is_open localhost 8443
+ local host=localhost port=8443
+ exec

This turns off `-x`, which was responsible for much of the logs, and also
captures stdout and stderr when testing the port. So instead we see:

Thu Nov  4 16:41:19 UTC 2021 - still trying to connect to localhost:8443
Thu Nov  4 16:41:20 UTC 2021 - still trying to connect to localhost:8443
Thu Nov  4 16:41:20 UTC 2021 - still trying to connect to localhost:8443
Thu Nov  4 16:41:21 UTC 2021 - still trying to connect to localhost:8443

This also increases the sleep time somewhat and decreases the number of attempts.

Also, add some shellcheck fixes from #190.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
-set -xeu
+set -eu
 
 port_is_open() {
   local host="$1" port="$2"
-  if exec 6<>/dev/tcp/$host/$port; then
+  if { exec 6<>/dev/tcp/"${host}"/"${port}" ; } 2>/dev/null ; then
     exec 6>&-
     return 0
   fi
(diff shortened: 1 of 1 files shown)


kevinburke:
Suggested some further fixes in https://github.com/rustls/rustls-ffi/pull/190 from github.com/koalaman/shellcheck.

PR #186, opened 2021-11-06
Add build status badge

jsha:


kevinburke:
I think the code looks good but I don't know that I particularly like these in README's... is the idea just to give people an indication that the code is better quality because the icon is green ?

jsha:
Yeah, exactly - it indicates that the owners use CI to keep the code quality good, and that the CI is currently passing. I can see the argument that they aren't so useful these days, when _most_ projects seem to use CI (and require that it passes before merging). But the `rustls` project has one, so I think it's reasonable to copy here.

kevinburke:
ok LGTM

PR #187, opened 2021-11-06
Use length specifier for User-Agent string

jsha:
Thanks @gvanem for spotting!


kevinburke:
LGTM, that was harder to verify in the printf manpage than I suspected it would be.

PR #188, opened 2021-11-07
Pass through ioslice type directly

### LINKED ISSUE #173: Windows: write_vectored_cb()
The vectored callback is not enabled for Winsock. But trying it and trusting the comment in
`src/rustls.h` for `struct rustls_iovec`:
```
/**
 * An alias for `struct iovec` from uio.h (on Unix) or `WSABUF` on Windows. You should cast
 * `const struct rustls_iovec *` to `const struct iovec *` on Unix, or `const *LPWSABUF`
 * on Windows. See [`std::io::IoSlice`] for details on interoperability with platform
 * specific vectored IO.
 */
```
Like:
```c
  if (WSASend(conn->fd, (LPWSABUF)iov, count, &n, 0, NULL, NULL) < 0) {
```

With a `set VECTORED_IO=1`and running the `client.exe` program, it gives me a `WSAEFAULT` from Winsock2
(`Error in rustls_connection_write_tls: errno 10014`)

Reason seems to be one cannot simply cast to a `LPWSABUF`. The layouts differ.
Reading https://docs.w3cub.com/rust/std/io/struct.ioslice did convince me a cast is good.

But this code with a locally defined `struct _iovec` works:
```c
rustls_io_result write_vectored_cb(
    void *userdata, const struct rustls_iovec *iov, size_t count, size_t *out_n)
{
  struct conndata *conn = (struct conndata *)userdata;

#ifdef _WIN32
  struct _iovec {
    void  *iov_base;
    size_t iov_len;
  };
  struct _iovec *_iov = (struct _iovec *) iov;
  DWORD n = 0;
  WSABUF *wsabufs = alloca (count * sizeof(WSABUF));
  int  i;

  for (i = 0; i < count; i++)
  {
    wsabufs[i].len = _iov[i].iov_len;
    wsabufs[i].buf = _iov[i].iov_base;
  }
  if (WSASend(conn->fd, wsabufs, count, &n, 0, NULL, NULL) < 0) {
    return errno;
  }
  *out_n = n;
#else
  ssize_t n = 0;
  n = writev(conn->fd, (const struct iovec *)iov, count);
  if(n < 0) {
    return errno;
  }
  *out_n = n;
#endif
  return 0;
}
```

Can someone please explain this?

jsha:
The docs at https://doc.rust-lang.org/std/io/struct.IoSlice.html say:

> It is semantically a wrapper around a &[u8], but is guaranteed to be ABI compatible with the iovec type on Unix platforms and WSABUF on Windows.

However, we were breaking this ABI compatibility by transforming into our own internal `rustls_slice_bytes` type. That type happened to have the same layout as `struct iovec` on Unix, so things worked fine on Unix but broke on Windows.

There's no need to transform into `rustls_slice_bytes`; we can just return a pointer to the beginning of the `&[IoSlice]`, plus a len.

Fixes #173
Followup to https://github.com/rustls/rustls-ffi/pull/112#issuecomment-864320536

Thanks to @gvanem for reporting!

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -3,8 +3,6 @@ use std::io::{Error, IoSlice, Read, Result, Write};
 use libc::{c_void, size_t};
 
 use crate::error::rustls_io_result;
-use crate::rslice::rustls_slice_bytes;
-use std::ops::Deref;
 
 /// A callback for rustls_connection_read_tls.
 /// An implementation of this callback should attempt to read up to n bytes from the
@@ -153,14 +151,13 @@ impl Write for VectoredCallbackWriter {
     fn write_vectored(&mut self, bufs: &[IoSlice<'_>]) -> Result<usize> {
         let mut out_n: usize = 0;
         let cb = self.callback;
-        let slices: Vec<rustls_slice_bytes> = bufs.iter().map(|b| b.deref().into()).collect();
         let result = unsafe {
             cb(
                 self.userdata,
                 // This cast is sound because IoSlice is documented to by ABI-compatible with
                 // iovec on Unix, and with WSABUF on Windows.
-                slices.as_ptr() as *const rustls_iovec,
-                slices.len(),
+                bufs.as_ptr() as *const rustls_iovec,
+                bufs.len(),
                 &mut out_n,
             )
         };


gvanem:
Looks good.
So `tests/*.c` code could now also support Windows in `rustls_io_result write_vectored_cb()`. Another PR?

jsha:
Yep!

PR #194, opened 2021-11-08
Fix documented lifetime on connection getters

### LINKED ISSUE #178: Lifetime issues for rustls_connection's getter methods
We offer three getter methods on connection that return a pointer to borrowed data:

```
const struct rustls_certificate *rustls_connection_get_peer_certificate(struct rustls_connection *conn,
                                                                        size_t i);
void rustls_connection_get_alpn_protocol(const struct rustls_connection *conn,
                                         const uint8_t **protocol_out,
                                         size_t *protocol_out_len);
const struct rustls_supported_ciphersuite *rustls_connection_get_negotiated_ciphersuite(const struct rustls_connection *conn);
```

Each of these has documentation in rustls-ffi that says "the returned pointer lives as long as the program" or similar. However, this is not guaranteed by Rust's lifetime system.

For instance, consider [rustls_connection_get_peer_certificate](https://docs.rs/rustls/0.20.0/rustls/enum.Connection.html#method.peer_certificates). Inside the rustls-ffi implementation, we take the input `*const rustls_connection` and cast it into an `&Connection`. We can then call methods on Connection that take `&self`, for instance:

```
pub fn peer_certificates(&self) -> Option<&[Certificate]>
```

So the lifetime of the returned `&[Certificate]` is bounded by our `&Connection` - which drops when we return from `rustls_connection_get_peer_certificate`. But we're returning a pointer to data in that `&[Certificate]`.

This would be fine if C only had a `const rustls_connection *`. But in practice, C has a `rustls_connection *`, and does in fact interleave mutating calls like `_read_tls` with calls that create `&` references. As a concrete example: if rustls ever implemented renegotiation (I know it's on the "will never" list), C code could make a call to `rustls_connection_get_peer_certificates` and store the returned pointer, then `_read_tls` / `_process_new_packets` some data that reallocates the internal `Vec<Certificates>`, which would cause C's stored pointer to be invalid.

I believe as a practical matter the data pointed to by these methods is set only once per connection, so we don't run an actual risk of it changing out from underneath us. But I'd like to figure out if we can use the type system somehow to ensure that. If we can't do that, we should ask rustls to commit to the set-once-per-connection property via documentation (and possibly code to the extent it is possible).

Thoughts @djc?

jsha:
And change rustls_connection_get_peer_certificate to take a `*const rustls_connection`.

Fixes #178.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,30 +2,85 @@
 
 ## 0.8.0 (unreleased)
 
-The package name has changed from "crustls" to "rustls-ffi". If you are
-importing it as a library from other Rust code, you should import `rustls_ffi`.
+The package name has changed to "rustls-ffi" (from "crustls").
+The header file (as installed by `make DESTDIR=/path/ install`)
+is now `rustls.h` and the library is `librustls.a`. The old library and header
+names are symlinked as part of the install process, to simplify upgrading to the
+new version.
 
-The header file is now named `rustls.h` instead of `crustls.h`, and the
-generated `.a` file is named `librustls_ffi.a` instead of `libcrustls.a`.
+If you are importing this as a library from other Rust code, you should import `rustls_ffi`.
+
+## New
+ - rustls_client_config_builder_new_custom and rustls_server_config_builder_new_custom:
+   start building a config, with ciphersuites and TLS versions set at initial construction.
+ - rustls_default_ciphersuites_get_entry() and
+   rustls_default_ciphersuites_len(): get default ciphersuites as opposed to
+   all ciphersuites (these happen to be the same today but might not always be).
+
+## Changed
+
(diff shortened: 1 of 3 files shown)


djc (approved):

CHANGELOG.md
```diff
@@ -52,6 +52,10 @@ If you are importing this as a library from other Rust code, you should import `
   rustls_server_config_builder_set_client_verifier and
   rustls_server_config_builder_set_client_verifier_optional, which are setters
   rather than constructors.
+- The documented lifetime for pointers returned by rustls_connection_get_peer_certificate
+  and rustls_connection_get_alpn_protocol has been fixed - the pointers those
+  function provide are valid until the next mutating function call on that
```
kevinburke, line 57: those functions*

kevinburke (approved):
LGTM

Issue #152, opened 2021-10-16
rustls_error can cause undefined behavior when a non-error is passed


jsha:
Steps to reproduce:

```
  char buf[256];
  size_t n;
  fprintf(stderr, "watch this:\n");
  rustls_error(0, buf, sizeof(buf), &n);
```

Result:

    watch this:
    Segmentation fault (core dumped)

This is because it's undefined behavior in Rust to have an enum where the contents are out of range for the enum definition, and there's no rustls_result corresponding to 0.

To fix this, we need to do some validation any time we accept a rustls_result from C. I think that's just in rustls_error.

jsha:
Fixed in #195

PR #195, opened 2021-11-08
Check input in rustls_error to avoid UB

jsha:
The input parameter for rustls_error was `rustls_result`. However, in
Rust [it's undefined behavior for an enum to hold an invalid value](https://github.com/rust-lang/unsafe-code-guidelines/issues/69). That
meant that if C passed an invalid value to rustls_error, UB would
result.

This changes the input parameter to be a uint, and relies on a macro
from the num_enum crate to check the value of that input parameter. If
the input is invalid, we emit the error for "InvalidParameter".

Part of #152.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ libc = "0.2"
 sct = "0.7"
 rustls-pemfile = "0.2.1"
 log = "0.4.14"
+num_enum = "0.5.4"
 
 [dev_dependencies]
 cbindgen = "*"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ libc = "0.2"
 sct = "0.7"
 rustls-pemfile = "0.2.1"
 log = "0.4.14"
+num_enum = "0.5.4"
```
djc, line 17: IMO the extra dependency for this is overkill. In this case it seems fine to write a custom `TryFrom` impl based on the repr type.

djc (commented):
Seems okay.

djc:
Yeah, it's not exactly trivial. Maybe something like the enum builder declarative macro we use in rustls?

jsha:
I'm going to merge as-is, then file a separate issue for achieving this without a dependency.

jsha:
Follow-up issue: #198

PR #196, opened 2021-11-08
Plug server mem-leaks

gvanem:
Building server with MSVC and in `_DEBUG` mode, showed 2 memory leaks:
```
 tests/server.c(120) : {524} normal block at 0x0E3B5B48, 10047 bytes long.
   Data: <HTTP/1.1 200 OK > 48 54 54 50 2F 31 2E 31 20 32 30 30 20 4F 4B 0D
tests/common.c(189) : {523} normal block at 0x0E3B5318, 2048 bytes long.
   Data: <GET / HTTP/1.1  > 47 45 54 20 2F 20 48 54 54 50 2F 31 2E 31 0D 0A
```

This patch frees the `response` and the plain-text vector.

### DIFF
--- a/tests/server.c
+++ b/tests/server.c
@@ -156,14 +156,13 @@ send_response(struct conndata *conn)
   n = sprintf(response, "%s %d\r\n\r\n", prefix, body_size);
   memset(response + n, 'a', body_size);
   *(response + n + body_size + 1) = '\0';
-  fprintf(stderr, "strlen response %ld\n", strlen(response));
+  response_size = strlen(response);
+  fprintf(stderr, "strlen response %ld\n", response_size);
 
   rustls_connection_write(
-    rconn, (const uint8_t *)response, strlen(response), &n);
+    rconn, (const uint8_t *)response, response_size, &n);
   
-  response_size = strlen(response);
-  free(response);
-  
+  free(response);  
   if(n != response_size) {
     fprintf(stderr, "failed to write all response bytes. wrote %ld\n", n);
     return CRUSTLS_DEMO_ERROR;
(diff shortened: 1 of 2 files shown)


tests/server.c
```diff
@@ -144,7 +144,7 @@ send_response(struct conndata *conn)
   struct rustls_connection *rconn = conn->rconn;
   const char *prefix = "HTTP/1.1 200 OK\r\nContent-Length:";
   const int body_size = 10000;
-  const int response_size = strlen(prefix) + 15 + body_size;
+  int response_size = strlen(prefix) + 15 + body_size;
```
jsha, line 147: Why get rid of the const here? I think that is causing the test failures:

```
tests/server.c: In function ‘send_response’:
Makefile:40: recipe for target 'target/server.o' failed
tests/server.c:160:38: error: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘int’ [-Werror=format=]
   fprintf(stderr, "strlen response %ld\n", response_size);
                                    ~~^
                                    %d
tests/server.c:166:8: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if(n != response_size) {
        ^~
```

Note that we could also probably fix the tests by declaring response_size a size_t.

jsha (changes requested):
Thanks for the fix! One question about the `const` below.

tests/server.c
gvanem: > Why get rid of the const here? 

Because it's reassigned further down. And it better be `size_t response_size = ..`.
But regarding the `%ld` you're using. It better be `%zd` since that matches an `int` for 32-bit mode.

jsha (approved):

jsha:
Thanks!

PR #197, opened 2021-11-08
Windows: fix Makefile.Windows

gvanem:
Extend the Windows-build (for CI) with:
* Add rules for `target/client.exe` and `target/server.exe`.
* `src/crustls.h` is no longer generated (why should it?)
* The C-library is now named `target/release/rustls_ffi.lib`.
* Reduce warning with `-D_CRT_NONSTDC_NO_WARNINGS`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,9 +45,7 @@ jobs:
           persist-credentials: false
       - name: Setup PATH for CL.EXE
         uses: ilammy/msvc-dev-cmd@v1
-      - name: Locate Cygwin's link.exe that shadows for MSVC's link.exe (so we can delete it first).
-      - run: where link.exe
-      - run: make -f Makefile.Windows
+      - run: where link.exe & make -f Makefile.Windows
 
   ensure-header-updated:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


gvanem:
After fixing the `test.yaml` file (that now works), there are issues with an `#include <sys/uio.h>` in `tests/client.c`.
Ref: https://github.com/rustls/rustls-ffi/runs/4137631017?check_suite_focus=true

Another PR to clean-up the mess in `tests/*.[ch]` coming soon.

jsha (commented):
This looks good, but still fails in CI with:

```
link -nologo -incremental:no -debug -map -verbose -out:target/client.exe  common.obj client.obj target/release/rustls_ffi.lib advapi32.lib userenv.lib ws2_32.lib > link.tmp
link: unknown option -- n
```

According to some googling, it sounds like this is most likely because there is a copy of the MinGW linker in the path ahead of the MSVC linker. I'm not sure how best to resolve. If we can figure out the page we can move it aside.

gvanem:
>  ... there is a copy of the cygwin linker in the path ahead of the MSVC linker. 

Add a `del  /Q<Cygwin-root>\bin\link.exe 2> NUL`?

jsha:
Thanks for doing this! I messed around with GitHub Actions and got things working. I put the resulting fixes in #211. Closing this PR out in favor of that one.

gvanem:
> I'm pretty mystified about why we're getting the "unknown option" error.

The `unknown option -- n` is possibly due to a `MAKEFLAGS` env-var in some other *step*.

PR #206, opened 2021-11-09
Fix enabling of ALPN and SNI in config builder

### LINKED ISSUE #204: no ALPN protocol sent by client connection
AFAICT, the ALPN protocol set by mod_tls on proxy connection is not seen in the client hello arriving at the contacted server.

my code sets `h2` as ALPN protocol via:
```
            rr = rustls_client_config_builder_set_alpn_protocols(builder,
                (rustls_slice_bytes*)rustls_protocols->elts, (size_t)rustls_protocols->nelts);
            if (RUSTLS_RESULT_OK != rr) goto cleanup;
```

but the contacted server connection does not see it.

This (with change method names, of course) used to work in previous versions.

Anything additional info I can provide, please do not hesitate to ask.

jsha:
We were collecting the ALPN and SNI information into our internal config builder object but not setting it on the actual rustls config builder object.

Add unittests for client and server builders that check this.

Fixes #204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## 0.8.0 (unreleased)
+## 0.8.1 (unreleased)
+
+## Changed
+ - Setting of ALPN protocols for client configs was broken in the 0.8.0 release.
+   This release fixes it.
+
+## 0.8.0 (2021-11-08)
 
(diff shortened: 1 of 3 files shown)


icing:
not only ALPN, but also SNI use! @jsha is on a roll!

icing (approved):

djc (approved):

PR #211, opened 2021-11-11
Fix Makefile.Windows

jsha:
Adapted from #197 by @gvanem. I squashed the commits from that PR, and added my own on top.

We need to remove the link.exe provided by MSYS and Git to make sure they don't interfere with MSVC's link.exe.

Also, add a #define for strncasecmp, which is not available on Windows.

### DIFF
--- a/Makefile.Windows
+++ b/Makefile.Windows
@@ -23,7 +23,7 @@ CFLAGS = -nologo -MD -Zi -W3 -O2   \
          -D_CRT_SECURE_NO_WARNINGS \
          -D_CRT_NONSTDC_NO_WARNINGS
 
-LDFLAGS = -nologo -incremental:no -debug -map -verbose
+LDFLAGS = -nologo -incremental:no -debug
 
 ifeq ($(USE_CLANG_CL),1)
   CC = clang-cl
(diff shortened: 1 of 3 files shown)


Makefile.Windows
```diff
@@ -1,50 +1,66 @@
...
 clean:
-	rm -f *.obj target/.rustc_info.json $(CRUSTLS_LIB) crustls.h vc1*.pdb
+	rm -f *.obj target/.rustc_info.json $(RUSTLS_LIB) vc1*.pdb
 	rm -fR target/*
 	rmdir target
 
+define link_EXE
+  $(call green_msg, Linking $(1))
+  link $(LDFLAGS) -out:$(strip $(1)) $(2) > link.tmp
+  @cat link.tmp >> $(1:.exe=.map)
```
jsha, line 63: @gvanem I'm confused about this aspect of your original PR: why emit output to link.tmp and then concatenate it to the binary's .map? The problem with this approach is that if linking errors, the errors wind up in link.tmp rather than in the CI output (which confused me for quite a while).

gvanem: Oh. Right. That's something I've used for years since I uses the `-verbose` option to see the details. Like which libraries are *not* needed etc. It has no place in CI.

PR #216, opened 2021-11-12
Add valgrind to C tests

jsha:
Fix some memory leaks.

Add clean shutdown to the server test so it can be tested under valgrind.

Send the stdout of client tests to /dev/null so they aren't so noisy.

Add a client or server prefix to all stderr messages to distinguish which side they came from.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,9 +47,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
-    - run: apt install valgrind
-    - run: export VALGRIND="valgrind -q"
-    - run: make test
+      - run: apt install valgrind
+      - run: export VALGRIND="valgrind -q"
+      - run: make test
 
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yaml
```diff
@@ -47,7 +44,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
-      - run: apt install valgrind
+      - name: Install valgrind
+        run: sudo apt-get install -y valgrind
       - run: export VALGRIND="valgrind -q"
       - run: make test
```
kevinburke, line 50: Interesting. What's the mechanism here that makes this run with valgrind turned on? Does Rust look for the VALGRIND env var?

jsha: Nope, test.sh is supposed to check for it! But I messed up the logic in test.sh; fixed in the latest push.

kevinburke:
ok LGTM

PR #219, opened 2021-11-13
Add feature to inhibit log capture

### LINKED ISSUE #212: Add no_log_capture feature
Right now rustls-ffi makes the assumption that it's built by the Makefile, with `-C metadata=rustls-ffi`, and so it's safe to take over the global log facility for itself, because that facility is now scoped to the library rather than global (#107).

However, when we're used as an rlib (#41), that won't be the case, and we should specifically avoid trying to take over the global log facility. We should do this with a feature flag that inhibits the log capture. Code that uses rustls-ffi via rlib will be able to provide its own log handler if it wants to.

jsha:
This is necessary to work properly when built as a Rust dependency (a
rare use case).

Fixes #212.

/cc @sagebind for review. I think this is the last thing we need before uploading rustls-ffi and making it usable as a backend for curl-rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,14 @@ description = "C-to-rustls bindings"
 edition = "2018"
 links = "rustls_ffi"
 
+[features]
+# Enable this feature when building as Rust dependency. It inhibits the
+# default behavior of capturing the global logger, which only works when
+# built using the Makefile, which passes -C metadata=rustls-ffi to avoid
+# interfering with copies of the global logger brought in by other Rust
+# libraries.
+no_log_capture = []
+
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
 rustls = { version = "=0.20", features = [ "dangerous_configuration" ] }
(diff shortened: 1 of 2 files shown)


sagebind (approved):
Looks OK from my perspective.

Due to the additive nature of Cargo features, it might make more sense from a Rust perspective to invert this feature (`log_capture`) which is disabled by default, but enabled by the Makefile, but I don't see any immediate issues with doing it this way.

jsha:
It's funny, I chose the negative flag exactly because of the additive nature of Cargo features. If crate A is expecting logs to not be captured, it shouldn't be possible for crate B somewhere else in the dependency tree to add a feature that causes logs to be captured after all.

What's the way in which additivity of features advocates for using a positive feature?

sagebind:
> If crate A is expecting logs to not be captured

I guess I was thinking of it in an inverse way -- crate A doesn't really have any expectations at all about logs, but when using it from C you want the logs to be captured. I was also thinking that if other crates were to decide to use rustls-ffi, almost universally they would want to enable `no_log_capture` which would become just extra ceremony, but when being used from C you use the Makefile, which could trivially request a non-default feature be enabled automatically.

But I'm fine with either way, there's an understandable logic behind the inverse feature as well.

PR #227, opened 2021-11-14
Improve handling of errors

### LINKED ISSUE #226: Update rustls_result_is_cert_error to include InvalidCertificateData
Per https://github.com/curl/curl/issues/8004, there is a test case in curl failing because an invalid certificate returns the wrong error. That's because, in the upgrade to rustls 0.20, we didn't add InvalidCertificateData to the list of is_cert_error results.

jsha:
There were still three places where we could have undefined behavior if
C gave us an invalid rustls_result:

 - rustls_result_is_cert_error took a rustls_result as a parameter.
 - The callback for server certificate validation returned
   rustls_result.
 - The callbacks for session storage returned rustls_result.

Change all of these to use u32 and the try_from implementation.

Fix the comment for the session storage callbacks, which indicated the
wrong error strategy.

Remove result_to_error in favor of directly implementing Display on
rustls_result. This fixes an issue where CertInvalidData was turned into
Either::String rather than Either::Error, which caused
rustls_result_is_cert_error to (incorrectly) say it wasn't a cert error.
This PR also removes the Either struct and the From impls on it.

Fixes #226

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -313,7 +313,7 @@ impl Display for rustls_result {
         CorruptMessagePayload => write!(f, "received corrupt message"),
 
         PeerIncompatibleError => write!(f, "peer is incompatible"),
-        PeerMisbehavedError => write!(f, "peer is incompatible"),
+        PeerMisbehavedError => write!(f, "peer misbehaved"),
 
         General => write!(f, "general error"),
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -47,18 +47,41 @@ impl rustls_result {
     }
 
     #[no_mangle]
-    pub extern "C" fn rustls_result_is_cert_error(result: rustls_result) -> bool {
-        match result_to_error(&result) {
-            Either::Error(e) => matches!(
-                e,
-                Error::InvalidCertificateData(_)
-                    | Error::InvalidCertificateEncoding
-                    | Error::InvalidCertificateSignature
-                    | Error::InvalidCertificateSignatureType
-                    | Error::InvalidSct(_)
-            ),
-            _ => false,
-        }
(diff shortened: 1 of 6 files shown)


src/error.rs
```diff
@@ -47,18 +47,41 @@ impl rustls_result {
...
+                | CertSCTMalformed
+                | CertSCTInvalidSignature
+                | CertSCTTimestampInFuture
+                | CertSCTUnsupportedVersion
+                | CertSCTUnknownLog
+        )
+    }
+}
+
+/// For cert-related resutl_results, turn them into a rustls::Error. For other
```
kevinburke, line 69: `rustls_results`

src/error.rs
```diff
@@ -247,139 +285,96 @@ pub(crate) fn map_error(input: rustls::Error) -> rustls_result {
...
         // message. So we reproduce a simplified error string.
-        InappropriateMessage => "received unexpected message".into(),
-        InappropriateHandshakeMessage => "received unexpected handshake message".into(),
-        CorruptMessagePayload => "received corrupt message".into(),
+        InappropriateMessage => write!(f, "received unexpected message"),
+        InappropriateHandshakeMessage => write!(f, "received unexpected handshake message"),
+        CorruptMessagePayload => write!(f, "received corrupt message"),
+
+        PeerIncompatibleError => write!(f, "peer is incompatible"),
+        PeerMisbehavedError => write!(f, "peer is incompatible"),
```
kevinburke, line 316: Copy paste?

kevinburke:
Looks OK to me, should we update the changelog now vs. later?

jsha:
We should update the CHANGELOG as part of the PR - I'll push a commit with that later on.

djc:
Seems okay to me.

djc (approved):

PR #233, opened 2021-11-14
doc: replace session with connection

jsha:
There were a few holdovers from when we renamed session to connection.


kevinburke:
LGTM

PR #240, opened 2021-11-15
rslice: add Debug impl for rustls_str

### LINKED ISSUE #239: Make it easier to inspect a rustls_str
I'm writing a test in Rust that is trying to check a rustls_str, and it's trickier because it's difficult to print the contents of the str to the console. 

It might be nice to implement one of

- #[derive(Debug)] for that struct
- a method to convert a rustls_str back into a Rust str
- std::fmt::Display for that struct. 

Open to any of these that you think would be best.

kevinburke:
This makes it easier to troubleshoot the contents of one of these
strings.

Fixes #239.

### DIFF
--- a/src/rslice.rs
+++ b/src/rslice.rs
@@ -166,6 +166,10 @@ impl<'a> TryFrom<&'a str> for rustls_str<'a> {
     }
 }
 
+/// rustls_str represents a string that can be passed to C code. The string
+/// should not have any internal null bytes and is not null terminated. C code
+/// should not create rustls_str objects, they should only be created in Rust
+/// code.
 impl<'a> rustls_str<'a> {
     pub fn from_str_unchecked(s: &'static str) -> rustls_str<'static> {
         rustls_str {
@@ -176,17 +180,27 @@ impl<'a> rustls_str<'a> {
     }
 }
 
+// If the assertion about Rust code being the only creator of rustls_str objects
+// changes, you must change this Debug impl, since the assertion in it no longer
+// holds.
 impl<'a> fmt::Debug for rustls_str<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        unsafe {
-            let ptr = self.data as *const u8;
-            let raw = slice::from_raw_parts(ptr, self.len as usize);
-            let s = str::from_utf8(raw).unwrap_or("%!(ERROR)");
-            f.debug_struct("rustls_str")
-                .field("data", &s)
-                .field("len", &self.len)
-                .finish()
-        }
+        let raw = unsafe {
+            // Despite the use of "unsafe", we know that this is safe because:
+            // - self.data is a u8, and single bytes are aligned
+            // - the entire memory range is a single allocated object, because
+            // we always build rustls_str objects from a slice (though this may
+            // change if we accept rustls_str objects from C code)
+            // - all values are properly initialized because we only init
+            // rustls_str objects inside of Rust code
+            // - not larger than isize::MAX because, again, it's coming from Rust
+            slice::from_raw_parts(self.data as *const u8, self.len as usize)
+        };
+        let s = str::from_utf8(raw).unwrap_or("%!(ERROR)");
+        f.debug_struct("rustls_str")
+            .field("data", &s)
+            .field("len", &self.len)
+            .finish()
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/rslice.rs
```diff
@@ -173,6 +176,20 @@ impl<'a> rustls_str<'a> {
     }
 }
 
+impl<'a> fmt::Debug for rustls_str<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        unsafe {
+            let ptr = self.data as *const u8;
+            let raw = slice::from_raw_parts(ptr, self.len as usize + 1);
+            let cstr = CStr::from_bytes_with_nul_unchecked(raw);
```
kevinburke, line 184: This was the best I could come up with but I'm sure that there's a better way

src/rslice.rs
```diff
@@ -173,6 +176,20 @@ impl<'a> rustls_str<'a> {
     }
 }
 
+impl<'a> fmt::Debug for rustls_str<'a> {
```
kevinburke, line 179: Should I add unsafe to the function definition

src/rslice.rs
```diff
@@ -173,6 +176,20 @@ impl<'a> rustls_str<'a> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        unsafe {
+            let ptr = self.data as *const u8;
+            let raw = slice::from_raw_parts(ptr, self.len as usize + 1);
+            let cstr = CStr::from_bytes_with_nul_unchecked(raw);
+            f.debug_struct("rustls_str")
+                .field("data", &cstr)
+                .field("len", &self.len)
```
kevinburke, line 187: I omitted the "phantom" field on purpose

src/rslice.rs
kevinburke: This is not correct and Miri flags it. All of the `CStr` API's expect a null byte at the end, we don't have one, which I try to fake by making the array one larger than it actually is. 

I'll take a look at doing this another way.

kevinburke: ah, I found a better way, I think

src/rslice.rs
```diff
@@ -184,6 +201,23 @@ fn test_rustls_str() {
...
+    use std::convert::TryInto;
+
+    #[test]
+    fn test_rustls_str_debug() {
+        let s = "abcd";
+        let rs: rustls_str = s.try_into().unwrap();
+        let mut buf = String::new();
+        write!(buf, "{:?}", rs).unwrap();
+        let want = r#"rustls_str { data: "abcd", len: 4 }"#;
+        assert_eq!(buf, want)
```
jsha, line 217: I would combine these into one line:

```suggestion
        assert_eq!(format!("{:?}", rs), r#"rustls_str { data: "abcd", len: 4 }"#)
```

jsha:
> Should I add unsafe to the function definition

No, and I don't think you can, since `Debug` is not an unsafe trait. Instead we should try to constraint the `unsafe` use inside this function as much as we can.

First, let's wrap the `unsafe` block around the one place that actually needs it: `slice::from_raw_parts(...)`.

`from_raw_parts` is unsafe, so let's figure out what constraints we need to follow to use it soundly: https://doc.rust-lang.org/nightly/std/slice/fn.from_raw_parts.html#safety

> Behavior is undefined if any of the following conditions are violated:
> - data must be valid for reads for len * mem::size_of::<T>() many bytes, and it must be properly aligned. This means in particular:

We get "properly aligned" for free. This is a u8 and u8's have align(1).

>  The entire memory range of this slice must be contained within a single allocated object! Slices can never span across multiple allocated objects. See below for an example incorrectly not taking this into account.

This is satisfied for rustls_str's we produce internally, because we always get them from a slice. If we accept rustls_strs from C code\*, we should document the fact that they must point to a single allocated object. It would be pretty bizarre to not do that, but we might as well pass along the requirement, in the name of precision.

\*I checked briefly and AFAICT we don't currently do this, though I thought we did at one point.

> data must be non-null and aligned even for zero-length slices. One reason for this is that enum layout optimizations may rely on references (including slices of any length) being aligned and non-null to distinguish them from other data. You can obtain a pointer that is usable as data for zero-length slices using NonNull::dangling().

We can check for NULL pretty easily; might as well do that. Though we could also document here "we only ever handle rustls_str that were generated internally, and those are always non-NULL."

> data must point to len consecutive properly initialized values of type T.

[This one is trickier than it sounds](https://www.ralfj.de/blog/2019/07/14/uninit.html)! Similar to the above, if we only ever handle rustls_str generated internally, this is true because the original `&str` must have been initialized. If we accept `rustls_str` from C, we would have to document "the pointed-to data must be initialized" (one of those things that gets assumed in a lot of places, but good to be explicit).

> The memory referenced by the returned slice must not be mutated for the duration of lifetime 'a, except inside an UnsafeCell.
>
> The total size len * mem::size_of::<T>() of the slice must be no larger than isize::MAX. See the safety documentation of pointer::offset.

Similar to the above.

So, long story short: When we're taking an unsafe function and wrapping it in a safe function, we should explain in a comment why we think it's sound to do so. Although, having said that, I realize that we use unsafe blocks throughout rustls-ffi without documenting the reasoning every single time, but instead relying on conventions.

jsha (changes requested):

kevinburke:
I've made the suggested changes

src/rslice.rs
kevinburke: Ah, much better thanks

jsha (approved):
Thanks!

Issue #203, opened 2021-11-09
rustls_{server/client}_config_builder_new_custom() parameters


icing:
As I see it, the new `rustls_client_config_builder_new_custom` and `rustls_server_config_builder_new_custom` require speficifaction of protocols *and* ciphersuites and panic if one is NULL.

Questions: 
 * how can one configure a supported cipher without mentioning the tls protocol version?
 * how can one enable a certain minimum protocol version without know all available?
 * how can one enable a specific protocol version without being aware which ciphers belong to it?
 * what is the outcome of a cipher and tls version that do not match?

Thanks for your help.

jsha:
> how can one configure a supported cipher without mentioning the tls protocol version?
> how can one enable a specific protocol version without being aware which ciphers belong to it?

@kevinburke provided a PR to support default cipher suites by passing NULL in the appropriate position: https://github.com/rustls/rustls-ffi/pull/165. Right now that's blocked on a redesign. Instead of NULL (which isn't clear to the reader of the code what it does), I'd like to have a global constant array RUSTLS_DEFAULT_CIPHER_SUITES and RUSTLS_DEFAULT_CIPHER_SUITES_LEN, so it's clear at the call site. Though perhaps our new rustls_default_cipher_suites_get() could do the same, if we can guarantee that its returned pointers are contiguous in memory (I'd have to check).

We'll also need the same treatment for protocol versions.

> how can one enable a certain minimum protocol version without know all available?

I think for this we need to add a list of supported_tls_versions or similar, and guarantee it is in sorted order.

> what is the outcome of a cipher and tls version that do not match?

There will be an error returned from rustls_{client,server}_config_builder_new_custom. We should document that. According to https://docs.rs/rustls/0.20.0/src/rustls/builder.rs.html#211-239, the error kind will be "General." We should file an upstream ticket to get a more specific error for that case.

icing:
Thanks for clarifying. I think passing NULL for default is quite common in a C API, but I have no qualms with passing any other constant you prefer.

PR #242, opened 2021-11-19
client: add ciphersuite and version arrays

divergentdave:
This adds arrays of ciphersuites and versions as suggested in #165 and #203.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
    rustls_result (#227).
  - rustls_session_store_get_callback and rustls_session_store_put_callback now
    return uint32_t (#227).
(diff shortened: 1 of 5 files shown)


jsha (approved):
This looks great, thank you! One interesting thing I noticed:

```
extern const uint16_t RUSTLS_DEFAULT_VERSIONS[2];
```

The length of the array is part of the declaration. That feels a little redundant with `RUSTLS_DEFAULT_VERSIONS_LEN`, but I guess the latter is more convenient that `sizeof(RUSTLS_DEFAULT_VERSIONS)/sizeof(uint16_t)`. The only downside I can see is if someone comes to depend on the latter, _and_ uses a version of the header that is out of date with respect to the library. I think that's fine though.

divergentdave:
Whoops, I had gotten the version order wrong at first, and then I only fixed it in one place.

jsha (approved):

PR #243, opened 2021-11-19
Implement Acceptor API

### LINKED ISSUE #155: Implement Acceptor API
rustls 0.20.0 [introduced the Acceptor API](https://docs.rs/rustls/0.20.0/rustls/server/struct.Acceptor.html). This allows servers to read off a ClientHello _before_ constructing a Connection, which makes selection of server certificates easier. For more details see https://github.com/rustls/rustls/issues/89.

jsha:
Fixes #155.

Since this relies on an "into" pattern (for the Accepted -> Connection transform), I've implemented the safer "into" pattern from #244: a rustls_acceptor always needs to be freed, whether its "into" method was called or not. This reduces the risk of double frees. I've introduced the new result code `rustls_result::AlreadyUsed` to cover the case when an accessor was called after the "into" method was called.

I've introduced another new result code `rustls_result::AcceptorNotReady` to cover the case when `accept()` returns `Ok(None)`.

For convenience I've flattened the methods from ClientHello onto `rustls_accepted`, so you don't need to get a ClientHello pointer out of the `rustls_accepted`, you can call the accessors directly.

Since we now have an accessor returning a `rustls_str` (previously they were passed in callbacks), we needed a way to make the lifetime of a `rustls_str` longer than the function that made it, so I introduced an `rustls_str::into_static` method.


jsha:
I've rewritten this thanks to some very useful offline feedback from @djc. It now looks a lot more like the Rust API.

src/acceptor.rs
```diff
@@ -0,0 +1,535 @@
...
+
+impl rustls_accepted {
+    /// Return the server name indication (SNI) from the ClientHello.
+    /// If the SNI contains a NUL byte, return a zero-length rustls_str.
+    /// Also return a zero-length rustls_str if there was some other
+    /// usage error, like calling with a NULL pointer or with an already-used
+    /// rustls_accepted. The returned rustls_str is valid until the next
+    /// time a method is called on the `rustls_accepted`. It is not owned
+    /// by the caller and does not need to be freed.
+    #[no_mangle]
```
djc, line 188: 0.20.6 added `ClientHello::cipher_suites()`, might want to add that here?

src/acceptor.rs
```diff
@@ -0,0 +1,535 @@
...
+                *out_n = n_read;
+            }
+
+            rustls_io_result(0)
+        }
+    }
+
+    /// Parse all TLS bytes read so far. Returns:
+    ///
+    /// - RUSTLS_RESULT_NOT_READY: a full ClientHello has not yet been read.
```
djc, line 150: I think this could use some more explanation on the state of `acceptor` after calling this. Should the caller free `acceptor` after getting `RUSTLS_RESULT_OK` or another `RUSTLS_RESULT_NOT_READY` return value?

src/acceptor.rs
```diff
@@ -0,0 +1,535 @@
...
+            } else {
+                0
+            }
+        }
+    }
+
+    /// Return the i'th ALPN protocol requested by the client in the ClientHello.
+    /// If the client did not offer the ALPN extension, return a zero-length rustls_slice_bytes.
+    /// Also return a zero-length rustls_slice_bytes on a usage error, like
+    /// calling with a NULL pointer or an already-used rustls_accepted.
```
djc, line 240: I noticed that the comment for `rustls_accepted_signature_scheme()` calls out the case where `i`is past the end of the slice, whereas this doesn't (although it talks generically about "usage error"). Maybe good to make that explicit?

In general, I feel like these comments could do with more structuring, like maybe different paragraphs about arguments and invariants and return values? The blob of text makes it harder to find the relevant concerns (though maybe this is specific to the way I'm reviewing here).

djc (commented):
Sorry to take so long. I took another pass, mainly through `acceptor.rs`.

src/acceptor.rs
jsha: Good idea. I updated all the comments to have `Parameters`, `Returns`, and `Memory and lifetimes` sections. I think it makes things a lot clearer. I'll update the other comments in the library to match the format after this lands.

src/acceptor.rs
jsha: Updated. Let me know if you think it's clearer.

djc:
Clippy says there's a redundant clone in src/acceptor.rs#L455. (Should we set CI to deny clippy warnings? I usually like to do that in other projects I maintain.)

src/acceptor.rs
djc: I feel like it might be more clear if the relevant memory and lifetimes concerns were attached to the particular parameters and return value?

src/acceptor.rs
```diff
@@ -0,0 +1,712 @@
...
+    /// to fill it up, just fill with as many bytes as you get in one syscall.
+    ///
+    /// Parameters:
+    ///
+    /// acceptor: The rustls_acceptor to read bytes into.
+    /// callback: A function that will perform the actual network I/O.
+    /// userdata: An opaque parameter to be passed directly to `callback`.
+    ///   Note: this is distinct from the `userdata` parameter set with
+    ///   `rustls_connection_set_userdata`.
+    /// out_n: An output parameter. On success, this will be written with
```
djc, line 142: Nit: "this will be written with" seems like an odd turn of phrase.

src/acceptor.rs
```diff
@@ -0,0 +1,712 @@
...
+    ///   the number of bytes `callback` said that it wrote.
+    ///
+    /// Returns:
+    ///
+    /// - 0: Success. You should call rustls_acceptor_accept() next.
```
djc, line 147: Maybe also backtick-quote things like `rustls_acceptor_accept()` and `callback` and `rustls_read_callback` (below)?

src/acceptor.rs
```diff
@@ -0,0 +1,712 @@
...
+    ///   *out_accepted.
+    /// - RUSTLS_RESULT_NOT_READY: a full ClientHello has not yet been read.
+    ///   Read more TLS bytes to continue.
+    /// - Any other rustls_result: the TLS bytes read so far cannot be parsed
+    ///   as a ClientHello, and reading additional bytes won't help.
+    ///
+    /// Memory and lifetimes:
+    ///
+    /// After this method returns RUSTLS_RESULT_OK, the rustls_acceptor is
+    /// still allocated and valid. It needs to be free regardless of success
```
djc, line 209: Nit: "free" -> "freed". Also, (caution, even more nitpicky) maybe it's clearer to talk about the argument `acceptor` vs the argument of type `rustls_acceptor`?

src/acceptor.rs
```diff
@@ -0,0 +1,712 @@
...
+                Some(a) => a,
+                None => return 0,
+            };
+            let hello = accepted.client_hello();
+            let cipher_suites = hello.cipher_suites();
+            if i < cipher_suites.len() {
+                cipher_suites[i].get_u16()
+            } else {
+                0
+            }
```
djc, line 359: I'd maybe write this as a `match`? For example:

```rust
match cipher_suites.get(i) {
    Some(cs) => cs.get_u16(),
    None => 0,
}
```

djc (commented):
All the nits.

src/acceptor.rs
jsha: Rewrote this.

src/acceptor.rs
jsha: I moved some of the explanations inline with the parameters. I left a few that felt like they were too long to be described inline.

src/acceptor.rs
jsha: Fixed, thanks.

src/acceptor.rs
jsha: Ah, good catch. Fixed.

jsha:
Regarding clippy warnings, I filed a PR: https://github.com/rustls/rustls-ffi/pull/261

jsha:
Also, in light of https://github.com/rustls/rustls/issues/1045, I removed `rustls_acceptor_wants_read` in the latest push and updated the `rustls_acceptor` comment to describe the reading flow without it.

src/acceptor.rs
djc: Yes, this is pretty okay now.

src/acceptor.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// Get the server name indication (SNI) from the ClientHello.
+    ///
+    /// Parameters:
+    ///
+    /// accepted: The rustls_accepted to access.
+    ///
+    /// Returns:
+    ///
+    /// A rustls_str containing the SNI field. The returned value is valid
+    /// until the next time a method is called on the `rustls_accepted`.
```
djc, line 222: I think this is unnecessarily restrictive? Most of the methods on `rustls_accepted` do not mutate, so having pointers into all of the server name, the sig scheme and the cipher suites should be fine.

src/acceptor.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// contain internal NUL bytes and is not guaranteed to contain valid
+    /// UTF-8.
+    ///
+    /// This will be a zero-length rustls_slice bytes in these cases:
+    ///   - i is greater than the number of offered ALPN protocols.
+    ///   - The client did not offer the ALPN extension.
+    ///   - The `accepted` parameter was already transformed into a connection
+    ///      with rustls_accepted_into_connection.
+    ///   
+    /// The returned value is valid until the next time a method is called
```
djc, line 346: Maybe make it explicit that the `'static` lifetime on the return value is thus fake and must not be relied upon?

PR #256, opened 2022-04-13
Avoid creating references to out params

### LINKED ISSUE #245: Possible undefined behavior in out parameter handling
A lot of our functions take out parameters, e.g.:

```c
rustls_result rustls_connection_read(struct rustls_connection *conn,
                                     uint8_t *buf,
                                     size_t count,
                                     size_t *out_n);
```

The implementation is usually like:

```rust
            let out_n: &mut size_t = try_mut_from_ptr!(out_n);
            ...
            *out_n = n_read;
```

I think this creates undefined behavior if the caller passes a pointer to an uninitialized size_t, since we're then creating a reference to uninitialized memory. It would be better to keep out_n as a raw pointer, do the null check at the top of the function, and then dereference it in an `unsafe` block at the end of the function.

Another possibility would be to define an `OutParam` wrapper type, something like:

```rust
#[repr(transparent)]
struct OutParam<T> {
  inner: NonNull<MaybeUninit<T>>
}

impl<T> OutParam<T> {
  fn from_ptr(p: *mut T) -> Option<Self> {...}
  fn write(T) { ... }
}
```

That would allow us to use the type system to ensure we've checked for null at the top. And would also allow us, on the Rust side, to clearly mark out parameters (though they are already marked rather clearly by `*mut` and occurring at the end of a function signature).

jsha:
Creating a reference to uninitialized memory is undefined behavior. Callers of this library are likely to pass pointers to uninitialized memory as out params. We should support that by not turning out params into references.

Instead, keep them as raw pointers and use an unsafe block where we actually write the param.

Fixes #245

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -47,8 +47,9 @@ impl rustls_certificate {
     ) -> rustls_result {
         ffi_panic_boundary! {
             let cert = try_ref_from_ptr!(cert);
-            if out_der_data.is_null() { return NullParameter }
-            if out_der_len.is_null() { return NullParameter }
+            if out_der_data.is_null() || out_der_len.is_null() {
+                return NullParameter
+            }
             let der = cert.as_ref();
             unsafe {
                 *out_der_data = der.as_ptr();
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -16,8 +16,8 @@ use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 use crate::error::rustls_result;
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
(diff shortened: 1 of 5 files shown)


jsha:
@tgeoghegan @djc if either of you is available to review this, I'd appreciate it. Thanks!

jsha:
I updated the issue with some related links: https://github.com/rustls/rustls-ffi/issues/245#issuecomment-1101712972. In particular, Miri can't catch this problem. From the [Miri README](https://github.com/rust-lang/miri#miri):

> In particular, Miri does currently not check that integers/floats are initialized or that references point to valid data.

djc:
This looks reasonable to me (though I haven't reviewed the unsafe invariants in depth).

PR #257, opened 2022-04-26
Clarify comment for rustls_certified_key_build

jsha:
It does not take ownership of the pointers it is passed, and the
resulting struct should usually be freed right after creating a config
with it.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -280,6 +280,17 @@ impl rustls_certified_key {
     /// the original caller has called `rustls_certified_key_free`, other objects
     /// may retain a pointer to the object. The memory will be freed when all
     /// references are gone.
+    ///
+    /// This function does not take ownership of any of its input pointers. It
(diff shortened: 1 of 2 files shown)


djc (approved):

src/cipher.rs
```diff
@@ -280,6 +280,17 @@ impl rustls_certified_key {
...
     /// references are gone.
+    ///
+    /// This function does not take ownership of any of its input pointers. It
+    /// parses the pointed-to data and makes a copy of the result. You may
+    /// free the cert_chain and private_key pointers after calling it.
+    ///
+    /// Typically, you will build a `rustls_certified_key`, use it to create a
+    /// `rustls_server_config` (which increments the reference count), and then
+    /// immediately call `rustls_certified_key_free`. That leaves the
+    /// `rustls_server_config` in posession of the sole reference, so the
```
djc, line 291: "posession" => "possession"

PR #258, opened 2022-05-11
Makefile: remove dependency on header file

jsha:
This makes header file updating a manual step during development,
so end-users building from source don't need cbindgen for `make`
to work.


djc (approved):

icing (commented):
Nice, thanks.

Issue #244, opened 2021-11-30
Make "into" patterns safer


jsha:
When building a server config or a client config, we use the Rust "into" pattern- we start with an object of one type (`*rustls_server_config_builder`), and we consume it in order to produce an object of another type (`*rustls_server_config)`. That means for that object's lifecycle, there are two ways out:

 - Call `rustls_server_config_builder_free` to free it early (for instance if there was an error).
 - Call `rustls_server_config_builder_build` to turn it into a `*rustls_server_config`.
 
If the user does _both_ of these, it's undefined behavior (accessing uninitialized memory). If the user does neither of these, it's a memory leak. This sort of requirement to do exactly one of two things, in all paths through your code, introduces a lot of potential for mistakes. In the spirit of safety, we'd like to particularly minimize the potential for mistakes that result in undefined behavior.
 
Proposal: For objects that get consumed as part of their lifecycle, instead of storing them as Box<Foo>, store them as Box<Option<Foo>>[^1]. When the Foo in consumed, [`take`](https://doc.rust-lang.org/nightly/std/option/enum.Option.html#method.take) it out of the Option. Document that `foo_free` must always be called, regardless of whether the Foo was consumed. If any methods are called on a `foo` after it has been consumed, return a new rustls_result code "AlreadyConsumed". For methods that don't return rustls_result, return a default value (false, 0, NULL, etc), as we already do for panics and null pointer inputs.

This takes the potential for undefined behavior and reduces it to a potential for memory leaks, and a small potential for incorrect application behavior if you call methods on a consumed object and rely on the default values returned. The cost is one extra word of storage and an extra check in each method on a consumable object. I think that cost should be pretty negligible.

/cc @icing @tgeoghegan @djc for feedback

[^1]: This does not participate in the [Null Pointer Optimization](https://doc.rust-lang.org/std/option/#representation), and we can't do Option<Box<Foo>> instead. But that's fine.

djc:
Seems reasonable to me!

icing:
tl;dr: +1

The builder objects, as you explained, have a special `state`. Typical objects go from `created+init` to `used` to `freed`. The `builder` has the additional `converted into` and `free` is no longer allowed.

Adding an additional, internal indirection, so that the Rust side can detect invalid free calls seems like a  very good idea. Especially since the costs are very minor.

[...]

PR #260, opened 2022-05-14
Fix lifecycle of rustls_root_cert_store

jsha:
Comments said that rustls_root_cert_store was represented as an Arc under the hood, but that's not true. It's represented as a Box. Instead, it's rustls_client_cert_verifier and rustls_client_cert_verifier_optional that are wrapped in an Arc.

Also, updated a couple of functions that take rustls_client_cert_verifier to accept a const one.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -16,8 +16,8 @@ use rustls_pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 use crate::error::rustls_result;
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
-    ffi_panic_boundary, try_mut_from_ptr, try_ref_from_ptr, try_slice, ArcCastPtr, BoxCastPtr,
-    CastConstPtr, CastPtr,
+    ffi_panic_boundary, try_box_from_ptr, try_mut_from_ptr, try_ref_from_ptr, try_slice,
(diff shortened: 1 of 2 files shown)


kevinburke (approved):

src/cipher.rs
```diff
@@ -529,11 +523,13 @@ impl rustls_client_cert_verifier {
     /// can be used in several rustls_server_config instances. Must be freed by
     /// the application when no longer needed. See the documentation of
     /// rustls_client_cert_verifier_free for details about lifetime.
+    /// This copies the contents of the rustls_root_cert_store. It does not take
```
djc, line 526: Why doesn't it take ownership like the Rust version does?

jsha: It looks like we didn't specifically discuss it when we landed this implementation: https://github.com/rustls/rustls-ffi/pull/98. However, I think it's good to keep the clone:

The "into" pattern is dangerous in C because it risks a use-after-free. It's not dangerous in Rust because the language prevents you from use-after-free. When it's not a significant performance hit, it makes sense to do a little cloning to increase safety. This is not a significant performance hit because, once created, the rustls_client_cert_verifier can be passed to many different rustls_config_builders (since it's an Arc internall).

Related: #244.

djc: Okay, I guess.

djc (approved):

PR #261, opened 2022-05-19
CI: Set clippy to deny warnings

jsha:
Thanks to @djc for pointing out that we don't currently do this.


djc (approved):
I'm confused that this isn't catching the redundant clone?

jsha:
The redundant clone was introduced on the `acceptor` branch (#243) - not yet merged. And I fixed in it my latest push to that branch.

PR #262, opened 2022-06-04
SNI hostname getter: set `out_n = 0` if unavailable

amesgen:
Tiny bug fix PR.

The docs of `rustls_server_connection_get_sni_hostname` state:

> Returns Ok with *out_n == 0 if there is no SNI hostname available on this connection


jsha (approved):
Looks good, thanks!

PR #263, opened 2022-06-06
Fix double free on building a server config when a client verifier got set

amesgen:
Otherwise, we get a double free when building a config from a config builder.


jsha (approved):
Good catch. Thanks!

Issue #297, opened 2023-03-22
Breaking changes in upstream error types.


cpu:
There were several upstream breaking API changes in Rustls related to error types:

* https://github.com/rustls/rustls/pull/1182 
* https://github.com/rustls/rustls/pull/1172
* https://github.com/rustls/rustls/pull/1215
* https://github.com/rustls/rustls/pull/1229
* https://github.com/rustls/rustls/pull/1198

The upstream changes will have to be adapted to `error.rs` and `client.rs`. 

* `rustls::internal::msgs::enums::AlertDescription` needs to be changed to the non-internal import.
* `InvalidCertificateEncoding`, `InvalidCertificateSignatureType`, `InvalidCertificateSignature`, `InvalidCertificateData`, `CorruptMessage`,  `CorruptMessagePayload`, `PeerIncompatibleError`, `PeerMisbehavedError` have been removed and need to be adapted to the more fine-grained error types that replaced them.

cpu:
@jsha Do you have any thoughts on how the `rustls_result` enum should adapt to the upstream changes: https://github.com/rustls/rustls-ffi/blob/55df12216f0004af12c5967725a0b46d9c4bd466/src/error.rs#L137-L152

For e.g. `CertInvalidEncoding = 7117`. The upstream `InvalidCertificateEncoding` entry in the `Error` enum this was being mapped to previously was removed and now we have `InvalidCertificate(x)` where `x` is something more specific like `CertificateError::BadEncoding`. 

A couple questions I have:
* Should all new additions use distinct `u32` values not previously used, leaving gaps for the removed types?
* Should there be a `u32` per more specific `CertificateError` variant, or just for the top level `InvalidCertificate`? 

I'm trying to implement some of these changes without having had time to do a lot of reading of the existing code, or the git blame. Any guidance is welcome :-)

jsha:
> Should all new additions use distinct u32 values not previously used, leaving gaps for the removed types?

Yes, please.

> Should there be a u32 per more specific CertificateError variant, or just for the top level InvalidCertificate?

Yes. The error design here is more-or-less "combinatorial explosion," since I looked at the error enum in rustls at the time and decided that splitting out errors for sub-variants was reasonable in scope.

For the `Other` sub-variant, we will have to drop the `Arc<_>` and any information in it, which is a bummer but the best approach for now.

[...]

PR #300, opened 2023-03-22
Small AlertDescription and NoneVerifier error fixes.

cpu:
This branch pulls out two easy fixes that don't require any adjustment of the `rustls_result` enum. The remaining fixes will take a little bit more care so I thought I'd put these up for review first. Updates #297 

### error: fix AlertDescription import path. - 269f18b697cac6885781a16f705c19c64aae0284
The `AlertDescription` enum was moved out of `rustls::internal::msgs::enums` into the top-level API. This commit updates the rustls-ffi import to match.

### client: update NoneVerifier err return type. - c73624966a09d3652def5f5c3cab569919c78a5b
The upstream `error::InvalidCertificateSignature` type was replaced with  a differentiated `InvalidCertificate` error that has a `CertificateError::BadSignature` variant. 

This commit updates the rustls-ffi `NoneVerifier` to use this more specific error type as its return from `verify_server_cert`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,8 +9,8 @@ use std::time::SystemTime;
 use libc::{c_char, size_t};
 use rustls::client::{ResolvesClientCert, ServerCertVerified, ServerCertVerifier};
 use rustls::{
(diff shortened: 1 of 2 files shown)


jsha (approved):
🛠️ !

PR #301, opened 2023-03-22
session: remove impl StoresClientSessions

jsha:
This trait no longer exists upstream. It was replaced by ClientSessionStore.

rustls-ffi does not yet export bindings for configuring client session support, only server session support. So we can safely remove this for now.


cpu (approved):
:scissors: :scissors: :scissors:

PR #302, opened 2023-03-22
Support IP addresses in server names

### LINKED ISSUE #292: ServerName can now have two kinds of value
In [rustls_verify_server_cert_params](https://docs.rs/rustls-ffi/0.9.2/rustls_ffi/client/struct.rustls_verify_server_cert_params.html), we treat ServerName as if it will only contain a DnsName. However, as of rustls 0.21, we can also see IpAddress. We need to update the interface to account for that.

jsha:
This actually involved basically no changes in functionality. It updates documentation to be clear that IP addresses are allowed, and updates some field and parameter names from dns_name or sni_name to server_name.

Note: there's one fix still needed after this. `InvalidDnsNameError => write!(f, "hostname was either malformed or an IP address (rustls does not support certificates for IP addresses)"),` should become "InvalidDnsNameError => write!(f, "server name was malformed (not a valid hostname or IP address)". But that was likely to cause conflicts with @cpu's in-progress fix for #297 so I'm omitting it for now.

Fixes #292

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -430,11 +430,12 @@ impl ResolvesServerCert for ResolvesServerCertFromChoices {
 }
 
 /// The TLS Client Hello information provided to a ClientHelloCallback function.
-/// `sni_name` is the SNI servername provided by the client. If the client
-/// did not provide an SNI, the length of this `rustls_string` will be 0. The
-/// signature_schemes carries the values supplied by the client or, should
-/// the client not use this TLS extension, the default schemes in the rustls
-/// library. See: <https://docs.rs/rustls/0.20.0/rustls/internal/msgs/enums/enum.SignatureScheme.html>.
+/// `server_name` is the value of the ServerNameIndication extension provided
+/// by the client. If the client did not send an SNI, the length of this
+/// `rustls_string` will be 0. The signature_schemes field carries the values
+/// supplied by the client or, if the client did not send this TLS extension,
+/// the default schemes in the rustls library. See:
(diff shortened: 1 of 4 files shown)


src/server.rs
```diff
@@ -430,11 +430,12 @@ impl ResolvesServerCert for ResolvesServerCertFromChoices {
...
-/// did not provide an SNI, the length of this `rustls_string` will be 0. The
-/// signature_schemes carries the values supplied by the client or, should
-/// the client not use this TLS extension, the default schemes in the rustls
-/// library. See: <https://docs.rs/rustls/0.20.0/rustls/internal/msgs/enums/enum.SignatureScheme.html>.
+/// `server_name` is the value of the ServerNameIndication extension provided
+/// by the client. If the client did not send an SNI, the length of this
+/// `rustls_string` will be 0. The signature_schemes field carries the values
+/// supplied by the client or, if the client did not send this TLS extension,
+/// the default schemes in the rustls library. See:
+/// <https://docs.rs/rustls/0.20.0/rustls/internal/msgs/enums/enum.SignatureScheme.html>.
```
cpu, line 438: Are you leaving these doc ref's at 0.20.0 or should they be updated? I was updating some I came across in https://github.com/rustls/rustls-ffi/pull/303/files

cpu (approved):
Looks great! Thanks :-)

src/server.rs
jsha: I'm leaving them for now and figure we'll do one big batch of updates. Speaking of which, I'll file an issue for that and add it to the tracker.

cpu: SGTM. I can revert that part of my diff when I pick it up again tomorrow!

PR #303, opened 2023-03-22
error: adapt rustls_result for upstream changes.

### LINKED ISSUE #297: Breaking changes in upstream error types.

cpu:
This commit adapts the `rustls_result` enum based on the changes upstream. Notably the `CertInvalidEncoding`, `CertInvalidSignatureType` and `CertInvalidData` errors have been folded into a new `CertificateError` enum that we expand out into its possible variants. Similarly the `CorruptMessage` and `CorruptMessagePayload` types have been replaced by a `InvalidMessage` enum that we expand out.

Resolves #297.


src/error.rs
```diff
@@ -213,26 +267,59 @@ pub(crate) fn map_error(input: rustls::Error) -> rustls_result {
...
+            InvalidMessage::MissingData(_) => MessageMissingData,
+            InvalidMessage::MissingKeyExchange => MessageMissingKeyExchange,
+            InvalidMessage::NoSignatureSchemes => MessageNoSignatureSchemes,
+            InvalidMessage::TrailingData(_) => MessageTrailingData,
+            InvalidMessage::UnexpectedMessage(_) => MessageUnexpectedMessage,
+            InvalidMessage::UnknownProtocolVersion => MessageUnknownProtocolVersion,
+            InvalidMessage::UnsupportedCompression => MessageUnsupportedCompression,
+            InvalidMessage::UnsupportedCurve(_) => MessageUnsupportedCurve,
+            InvalidMessage::UnsupportedKeyExchangeAlgorithm(_) => MessageUnsupportedCompression,
+            _ => General,
```
cpu, line 299: I was surprised I needed this `_` arm since I thought the matches above were exhaustive, but the compiler complains otherwise. Maybe more investigation needed.

(Same for the `Error::InvalidCertificate(e) => match e` lower)

src/error.rs
```diff
@@ -70,10 +78,17 @@ pub(crate) fn cert_result_to_error(result: rustls_result) -> rustls::Error {
...
+        CertExpired => InvalidCertificate(CertificateError::Expired),
+        CertNotYetValid => InvalidCertificate(CertificateError::NotValidYet),
+        CertRevoked => InvalidCertificate(CertificateError::Revoked),
+        CertUnhandledCriticalExtension => InvalidCertificate(CertificateError::UnhandledCriticalExtension),
+        CertUnknownIssuer => InvalidCertificate(CertificateError::UnknownIssuer),
+        CertBadSignature => InvalidCertificate(CertificateError::BadSignature),
+        CertNotValidForName => InvalidCertificate(CertificateError::NotValidForName),
+        CertInvalidPurpose => InvalidCertificate(CertificateError::InvalidPurpose),
+        CertApplicationVerificationFailure => InvalidCertificate(CertificateError::ApplicationVerificationFailure),
+        CertOtherError => InvalidCertificate(CertificateError::Other(Arc::<dyn std::error::Error>::new("".into()))), // TODO(cpu): busted!
```
cpu, line 91: I was having a hard time figuring out the right way to contort an empty `&str` into a `Arc<dyn StdError + Send + Sync>`. I suspect I'm missing something obvious. This attempt doesn't compile :-)

cpu (commented):
@jsha Here's where I have to stop for the day. I self-reviewed to call out a couple places I know I've missed the mark. Hoping that this is a helpful start. I will iterate further tomorrow.

src/error.rs
```diff
@@ -301,25 +388,58 @@ impl Display for rustls_result {
...
         InvalidParameter => write!(f, "a parameter had an invalid value"),
-        CertInvalidData => write!(f, "invalid certificate data found"),
         UnexpectedEof => write!(f,  "peer closed TCP connection without first closing TLS connection"),
         PlaintextEmpty => write!(f,  "no plaintext available; call rustls_connection_read_tls again"),
         AcceptorNotReady => write!(f, "rustls_acceptor not ready yet; read more TLS bytes into it"),
         AlreadyUsed => write!(f, "tried to use a rustls struct after it had been converted to another struct"),
 
+        // TODO(@cpu): It feels like CertificateError should implement Display, but it does not
+        //             presently so we can't call CertificateError::BadEncoding.fmt(f) (for example)
+        //             since that will be a Debug fmt.
```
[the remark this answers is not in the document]
jsha: The way upstream gets away with this is that the Display implementation for rustls::Error prints the Debug `{:?}` output for the sub-variant, which is just its name: https://github.com/rustls/rustls/blob/6831835c56696d61024f2ae1261b41edad54e685/rustls/src/error.rs#L396-L398

Several of the other variants of Error do the same thing. The way I've dealt with it so far is to construct a `rustls::Error` with the appropriate variant and sub-variant, and `fmt` that. You can see an example in the AlertReceived errors [below](https://github.com/rustls/rustls-ffi/pull/303/files#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eR454-R475). It probably would be nice for upstream to provide a friendlier error here, and I'm not opposed to overriding here with slightly friendlier errors, as you've done.

src/error.rs
jsha: Not simple! There are two `From` impls that should get us there, and `Into` / `From` are supposed to apply recursively, but I think sometimes the language needs a little help:

https://doc.rust-lang.org/std/error/trait.Error.html#impl-From%3C%26str%3E-for-Box%3Cdyn%20Error%20+%20Send%20+%20Sync%20+%20%27a%2C%20Global%3E

https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-From%3CBox%3CT%2C%20Global%3E%3E-for-Arc%3CT%3E

This ugly thing works for me:

```
        CertOtherError => InvalidCertificate(CertificateError::Other(
            Into::<Box<dyn std::error::Error + Send + Sync>>::into("").into(),
        )),
```

There may be a better way but I'm not finding it at the moment.

src/error.rs
jsha: The upstream `InvalidMessage` enum [has the #[non_exhaustive] attribute](https://github.com/rustls/rustls/blob/6831835c56696d61024f2ae1261b41edad54e685/rustls/src/error.rs#L97), so it's an [error to match against it in any match statement that lacks an `_`.](https://doc.rust-lang.org/reference/attributes/type_system.html).

Rather than bailing out to General, we should probably define a `MessageInvalidOther` for this case.

src/error.rs
cpu: Thank you! I was trying to `Box` in some of my other quick attempts at this but I was misinterpreting the problem and boxing the arc which didn't make a lot of sense (or solve the problem). It's comforting to know I was at least reaching in the right general direction.

src/error.rs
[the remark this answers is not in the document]
cpu: For the variants where the friendly message wasn't adding much above the name of the type I deferred to the upstream `Error` display. For the variants where there was an embedded field I left the simplified messages.

cpu:
@jsha I think this branch is ready for review now. The Rust parts all build/test cleanly. The remaining CI failures are from `src/client.c` using the old `dns_name` struct field that was renamed. I'll fix that in a separate PR since it's unrelated to the error work.

cpu:
> cpu cpu force-pushed the cpu-297-rustls_result-update branch from 2f9b3d4 to c02b806

This was me remembering the note you left in #302 - I've rolled that `InvalidDnsNameError` change into this work.

cpu:
>  cpu force-pushed the cpu-297-rustls_result-update branch from c02b806 to 96e62ba 

Fmt fix.

src/error.rs
```diff
@@ -290,93 +382,181 @@ impl Display for rustls_result {
...
-        CertSCTTimestampInFuture => Error::InvalidSct(sct::TimestampInFuture).fmt(f),
-        CertSCTUnsupportedVersion => Error::InvalidSct(sct::UnsupportedSctVersion).fmt(f),
-        CertSCTUnknownLog => Error::InvalidSct(sct::UnknownLog).fmt(f),
+            // These variants are local to this glue layer.
+            rustls_result::Ok => write!(f, "OK"),
+            Io => write!(f, "I/O error"),
+            NullParameter => write!(f, "a parameter was NULL"),
+            InvalidDnsNameError => write!(
+                f,
+                "server name was malformed (not a valid hostname or IP address"
```
jsha, line 391:
```suggestion
                "server name was malformed (not a valid hostname or IP address)"
```

cpu: Good catch, thanks! I added a fix and force pushed. LMK if you'd prefer to see separate review fix commits that get squashed at the end.

cpu: > Generally speaking I prefer separate review fix commits, and always use squash merges. But for small stuff like this I'm not particularly picky.

:+1: 

> Also I recently learned a nice trick: In the "user force-pushed the branch branch..." message, "force-pushed" is clickable and will take you to a diff between the old branch and the force-pushed branch.

Oh hey, TIL. That is handy.

PR #306, opened 2023-03-23
docs: update docs.rs links to prefer /latest.

### LINKED ISSUE #304: Update doc links for 0.21
Our documentation links out to rustls documentation for many things. For instance, `rustls_client_config_builder` links to https://docs.rs/rustls/0.20.0/rustls/struct.ConfigBuilder.html. We should update these links for the 0.21 release.

While we're at it, we should consider what version to link to:

  1. Link to the exact version of rustls we currently depend on. For consistency, this requires updating all doc links each time we bump the version of rustls, which we haven't done consistently so far.
  2. Link to `/latest/`. This stays up-to-date but will result in 404s if a future semver-incompatible version of rustls removes something we link to.
  3. Link to `/~0.21/`. This is a [special docs.rs redirection](https://docs.rs/about/redirections) that will always link to the latest version in the `0.21` series.

So far I like (2) best. The issue of future breakage is a fairly small one, and linking to `/latest/` is nice and simple.

cpu:
Previously the Rustls docs linked to docs.rs with URLs containing specific versions of the target crate (e.g.
`docs.rs/rustls/0.20.0/...`). In this commit all such links are replaced with to use `/latest` instead of a fixed version.

This provides a simple way to keep things up to date and the risk of future breakage is fairly small since this crate is kept in close synchronicity with the crates it links to.

Resolves #304.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -24,7 +24,7 @@ use std::ops::Deref;
 
 /// An X.509 certificate, as used in rustls.
 /// Corresponds to `Certificate` in the Rust API.
-/// <https://docs.rs/rustls/0.20.0/rustls/struct.Certificate.html>
+/// <https://docs.rs/rustls/latest/rustls/struct.Certificate.html>
 pub struct rustls_certificate {
     // We use the opaque struct pattern to tell C about our types without
     // telling them what's inside.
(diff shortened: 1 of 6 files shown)


jsha:
Thanks for this! With the recent errors merge, this now has a conflict in src/error.rs.

cpu:
>  this now has a conflict in src/error.rs.

All fixed :broom:

jsha (approved):
📜

PR #316, opened 2023-04-28
Consolidate test certificates

jsha:
Previously some tests used certificates in the root of the repository, and others used ones under testdata/. Consolidate on testdata/ and also reissue those certificates to extend their expiration date.

This also fixes tests failing on main due to an expired localhost certificate.


cpu (approved):
This looks good to me. Once I cherry-picked the runner OS version fix from https://github.com/rustls/rustls-ffi/pull/317 onto this branch it was all green :tada:

PR #317, opened 2023-05-01
CI: upgrade runner OS, 18.04 -> 20.04.

cpu:
While reviewing https://github.com/rustls/rustls-ffi/pull/316 I noticed that CI was still not passing across the board. The root cause is that the Ubuntu 18.04 runner image that was being used for some of the GitHub actions workflows is no longer supported, resulting in those workflows getting stuck in a "waiting for runner" state (See https://github.com/actions/runner-images/issues/6002).

This commit updates the runner OS from 18.04 to 20.04.


jsha (approved):
Thanks for figuring that out!

PR #319, opened 2023-05-30
tests/server.c: fix segfault when printing usage

ctz:
```
$ ./target/server
usage: ./target/server cert.pem key.pem

Listen on port 8443 with the given cert and key.
Segmentation fault (core dumped)
```

This was because the `cleanup` label skipped over the initialisation of `certified_key`, meaning the pointer was uninitialised during `rustls_certified_key_free`.

### DIFF
--- a/tests/server.c
+++ b/tests/server.c
@@ -302,6 +302,7 @@ main(int argc, const char **argv)
     rustls_server_config_builder_new();
   const struct rustls_server_config *server_config = NULL;
   struct rustls_connection *rconn = NULL;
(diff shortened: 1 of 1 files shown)


jsha (approved):
Thanks for spotting that, and for the fix! I've checked this locally and it works great.

@cpu FYI CMake is only needed for Windows support. On Linux and macOS `make` should work fine.

cpu:
> @cpu FYI CMake is only needed for Windows support. On Linux and macOS make should work fine.

Ahh! That makes things easier. Thanks

PR #320, opened 2023-06-29
Track rustls 0.22 API changes

ctz:
Just parking this here.

PR #321, opened 2023-06-29
Test client/server mTLS support.

cpu:
## Description

In preparation for supporting client certificate revocation checking w/ CRLs it felt prudent to update the `tests/client.c` and `tests/server.c` pair of test programs to allow configuration/testing of mutually authenticated TLS (mTLS). 

### tests: common read_file and load_cert_and_key.
Previously only the `tests/server.c` code needed to load a `rustls_certified_key` (for the server cert/keypair). In a subsequent
commit the `tests/client.c` code will need to do the same for optionally providing a `rustls_certified_key` for client certificate
authentication. 

In preparation, this commit lifts the `read_file` and `load_cert_and_key` helper functions from `tests/server.c` into
`tests/common.c` (updating `tests/common.h` to match) where both client and server test programs can use the shared code.

### tests/client: add client certificate support.
This commit updates `tests/client.c` to allow setting two new env vars, `AUTH_CERT` and `AUTH_KEY`.

If neither are set, the program works as it did before: no client certificate is sent for mTLS.

If one but not both of these env vars are set, the program will error: they must both be provided.

If both are set, the `AUTH_CERT` and `AUTH_KEY` files are loaded into a `rustls_certified_key` and the built `rustls_client_config` will be configured to offer client certificate authentication with the server using the cert/key pair.

### tests/server: add optional required client cert auth.
This commit updates the `tests/server.c` program so that if an `AUTH_CERT` env var is provided the server will be configured to require clients provide a client certificate issued that chains to the `AUTH_CERT` certificate authority. If no `AUTH_CERT` env var is set the server works as it did before, ignoring client certificate authentication.


cpu:
> rustls-ffi / ensure-header-updated (pull_request) Failing after 9s

I can't reproduce this diff locally. I suspect it's a difference in `bindgen` versions between CI (xx) and what I have locally (0.24.3)

> rustls-ffi / Build+test (clang, 1.57.0, ubuntu-20.04) (pull_request) Failing after 2m

This looks like a dependency with an incompatible MSRV and also failing on `main`.

> rustls-ffi / Windows (pull_request) Failing after 1m

These look like clang warnings that are also failing on `main`.

I'll chase some of these CI issues in a separate branch :broom:

cpu:
> I'll chase some of these CI issues in a separate branch 🧹

https://github.com/rustls/rustls-ffi/pull/320

cpu:
> cpu force-pushed the cpu-test-mtls-support branch from ef0d20f to 11a1ab7 

Rebased to pick up the CI fixes.

ctz (approved):

cpu:
Another test only jam & this one's been up for a bit so I'm going to merge as-is with the same disclaimer as the others: happy to revise down the road if there's any new feedback.

PR #323, opened 2023-07-03
WIP: client certificate revocation checking support.

cpu:
**Note to reviews: this is a draft, it builds on yet-to-be-merged work. Notably, https://github.com/rustls/rustls-ffi/pull/321.**

Content new to this branch begins at be8730c8036c2243b7f132baff8c4ccca04f40cd

I would recommend holding off reviewing this PR while the upstream Rustls portion is still in-flux.

TODO:

* [ ] Wait for Rustls https://github.com/rustls/rustls/pull/1326 to be merged
* [ ] Update based on shape of upstream API that lands
* [ ] Error updates - need to map CRL specific errors to `rustls_result`.
* [ ] Double checking consistency/safety of changes.
* [ ] Update Cargo patch to use `main` branch of Rustls.
* [ ] Wait for new webpki release,
* [ ] Wait for rustls release.
* [ ] Remove Cargo patches.

### deps: update pemfile, use unreleased Rustls, webpki.
Updates `pemfile` from 0.2.1 to 1.0.3 to pick up support for reading DER
encoded CRLs from .pem files.

Updates webpki to use tip of main, picking up unreleased CRL support.

Updates rustls to use a fork/branch that adds WIP CRL support.

### tidy: remove usage of removed upstream SCT features.
The upstream Rustls project has dropped the minimal SCT support it offered. This commit tracks that change in rustls-ffi, removing the dep on the `sct` crate and removing related features.

### cipher: switch client cert verifiers to mutable ptrs.
In order to support adding CRLs to a constructed `rustls_client_cert_verifier` or `rustls_client_cert_verifier_optional` we need to change the constructor return type from `*const` to `*mut`. Corresponding destructors are updated as well.

### cipher: add client verifier CRL pem fns.
This commit updates the `rustls_client_cert_verifier` and `rustls_client_cert_verifier_optional` API surface to include a function for loading CRLs from a PEM file.

### server: support reading CRL PEM for client auth.
This commit updates the `tests/server.c` example program to support reading one or more CRLs from a single PEM encoded CRL file, provided via `AUTH_CRL`. This option is only processed when the server is performing mandatory client authentication (e.g. `AUTH_CERT` was provided).

### tests: add CRL mTLS test.
This commit adds a simple test CRL (`testdata/test.crl.pem`) that lists the `testdata/localhost/cert.pem` certificate as revoked, but _not_ the `testdata/example.com/cert.pem` certificate.

The `client-server.py` integration test driver is then updated with a suite that will start the server binary in a mode that requires mTLS, and that loads the test crl. Two connection attempts are made with the client binary: one using the `example.com` client cert that isn't expected to error, and one using the `localhost` client cert that _is_ expected to error (since it's revoked).


src/cipher.rs
```diff
@@ -517,10 +519,11 @@ pub struct rustls_client_cert_verifier {
     _private: [u8; 0],
 }
 
-impl CastConstPtr for rustls_client_cert_verifier {
+impl CastPtr for rustls_client_cert_verifier {
     type RustType = AllowAnyAuthenticatedClient;
 }
 
+impl BoxCastPtr for rustls_client_cert_verifier {}
 impl ArcCastPtr for rustls_client_cert_verifier {}
```
cpu, line 527: I'm a little bit skeptical these cahnges to the `*Ptr` traits for the verifiers are correct. Earmarked to dig into the associated macros for closer study.

In particular: worried the change from Arc -> Box on the constructor side will interact poorly with multiple handles to a verifier being used to add CRLs, then being turned back to Arc to give to Webpki :thinking:

cpu:
Going to close this WIP and replace it with a new one. We ended up dropping the `add_crl` approach on the verifier struct upstream for a more natural builder API. I will rework and open a fresh version in the next day or two.

PR #324, opened 2023-07-07
Rustls 0.21.5, client verifier CRL support.

cpu:
### chore: ignore clion/jetbrains dir, venv dir.

Small quality of life improvement for developers using clion and a venv for the Python tests.

### deps: update rustls ecosystem dependencies.

Brings in the latest Rustls and supporting libraries.

* Rustls 0.21.0 -> 0.21.5
* Webpki 0.22 -> Rustls webpki 0.101.0
* Rustls pemfile 0.2.1 -> 1.0.3

### cipher: rename verifier types to match upstream.

This commit renames the `rustls_client_cert_verifier` and `rustls_client_cert_verifier_optional` types to better match their upstream Rustls equivalents: `rustls_allow_any_authenticated_client_verifier` and `rustls_allow_any_anonymous_or_authenticated_client_verifier`. While wordier, these names are easier to understand and to map back to their Rust equivalents.

Associated functions (`_new` and `_free`) are updated to match the new naming scheme.

### cipher: client cert verifier w/ CRLs support.

This commit reworks both the `rustls_client_cert_verifier_optional` and `rustls_client_cert_verifier` structs to use a separate builder pattern. This enables optionally providing the builder CRLs from PEM files. Afterwards the intermediate builder can be turned into the standard `const *` verifiers to provide to the client configuration.

Thanks to @ctz for suggesting this approach and sketching it out with me in a draft branch I've tided up and extended here.

### server: support reading CRL PEM for client auth.

This commit updates the `tests/server.c` example program to support reading one or more CRLs from a single PEM encoded CRL file, provided via `AUTH_CRL`. This option is only processed when the server is performing mandatory client authentication (e.g. `AUTH_CERT` was provided).

### tests: add CRL mTLS test.
This commit adds a simple test CRL (`testdata/test.crl.pem`) that lists the `testdata/localhost/cert.pem` certificate as revoked, but _not_ the `testdata/example.com/cert.pem` certificate.

The `client-server.py` integration test driver is then updated with a suite that will start the server binary in a mode that requires mTLS, and that loads the test crl. Two connection attempts are made with the client binary: one using the `example.com` client cert that isn't expected to error, and one using the `localhost` client cert that _is_ expected to error (since it's revoked).


cpu:
> rustls-ffi / Clippy nightly (optional) (pull_request) 

This failure is nightly clippy saying:
```
error: usage of `Arc<T>` where `T` is not `Send` or `Sync`
Error:    --> src/lib.rs:396:23
    |
396 |         Arc::into_raw(Arc::new(src)) as *const _
    |                       ^^^^^^^^^^^^^
    |
    = help: consider using `Rc<T>` instead or wrapping `T` in a std::sync type like `Mutex<T>`
```

Ctz spotted that:

1. This is also happening on `main`.
2. It looks to be this false positive issue: https://github.com/rust-lang/rust-clippy/issues/11076

src/error.rs
```diff
@@ -166,7 +167,15 @@ u32_enum_builder! {
...
         CertSCTUnsupportedVersion => 7322,
-        CertSCTUnknownLog => 7323
+        CertSCTUnknownLog => 7323,
+
+        // From InvalidCertRevocationList
+        // https://docs.rs/rustls/0.21.3/rustls/enum.Error.html#variant.InvalidCertRevocationList
+        //
+        // TODO(XXX): due to a small oversight the 0.21.3 release did not export the
+        //   InvalidCertRevocationList's CertRevocationListError enum. Once this is fixed + released
+        //   upstream we should flatten the sub-variants here.
```
cpu, line 177: Fumbling through a maintenance release in https://github.com/rustls/rustls/pull/1348 to hopefully get this sorted before merge.

cpu: This was fixed in [v0.21.4](https://github.com/rustls/rustls/releases/tag/v%2F0.21.4) - I will update this branch to provide better error specificity using that fix shortly.

cpu:
> cpu force-pushed the jbp-cpu-rustls-0.21.3 branch from bf791cd to 536ddce 

Updated to Rustls 0.21.5 (_ignore the branch name :sweat_smile:_), addressed the TODO's I left RE: error specificity now that the CRL error sub-variants are properly exported. I think this is ready for review.

ctz (approved):
i'm perhaps a bit compromised on this, but i think this looks good.

reviewing this makes me think the rustls api for choosing between mandatory and optional client auth is a bit inflexible. but that's not an issue for here.

cpu:
@jsha Is this a PR you would be interested in reviewing? Since both Ctz and myself contributed code it would be nice to have an independent set of eyes on it if you're available.

jsha:
Yep! Planning to check it out a little now and a little tomorrow.

src/cipher.rs
```diff
@@ -509,6 +511,97 @@ impl rustls_root_cert_store {
...
 }
 
+/// A builder for a `rustls_client_cert_verifier`. This builder object can be used to configure
+/// certificate revocation lists, and then turned into a `rustls_client_cert_verifier` once ready.
+pub struct rustls_client_cert_verifier_builder {
+    _private: [u8; 0],
+}
+
+// XXX: contained value is consumed even on error, so this can contain None. but the caller
+// still needs to free it
```
jsha, line 521: This is Good, Actually.

There's a problem where Rust's "into" patterns can lead to really complicated lifecycle management on the C side. I wrote up the problem here: https://github.com/rustls/rustls-ffi/issues/244, and implemented the solution here: https://github.com/rustls/rustls-ffi/pull/243/files#diff-5451b831ef642c0ec3aab8375dd75414f81b677f5f638a3d44d1ff1d226117daR59-R61. The basic idea is that if you created something, you have to free it, whether you (a) converted it using an "into' or builder call, (b) failed to convert it using an "into" or builder call, or (c) errored out before trying to convert it.

In the Acceptor PR I did `RustType = Option<X>`; here you have a newtype. I think the `Option<X>` is slightly nicer but don't feel super stongly; however, we should do the same in both places.

src/cipher.rs
```diff
@@ -509,6 +511,97 @@ impl rustls_root_cert_store {
...
+        verifier: *mut rustls_client_cert_verifier_builder,
+        crl_pem: *const u8,
+        crl_pem_len: size_t,
+    ) -> rustls_result {
+        ffi_panic_boundary! {
+            let client_cert_verifier_holder = match BoxCastPtr::to_box(verifier) {
+                None => {
+                    return NullParameter;
+                },
+                Some(x) => Box::leak(x), // nb. does not take ownership of the box.
```
jsha, line 567: I think is incorrect and will cause a memory leak of the `AllowAnyAuthenticatedClientHolder` even after the contents of its options are taken.

Elsewhere we use `try_mut_from_ptr!` to produce an `&mut` reference. I think that's the right semantics here; someone else (the C code) continues to hold and own the Box, but we are mutating the contents in this function.

So that would look something like:

    let client_cert_verifier_holder: &mut AllowAnyAuthenticatedClientHolder = try_mut_ptr!(verifier)

src/cipher.rs
```diff
@@ -509,6 +511,97 @@ impl rustls_root_cert_store {
...
+
+            let crl_pem: &[u8] = try_slice!(crl_pem, crl_pem_len);
+            let crls_der = match crls(&mut Cursor::new(crl_pem)) {
+                Ok(vv) => vv.into_iter().map(UnparsedCertRevocationList).collect::<Vec<UnparsedCertRevocationList>>(),
+                Err(_) => return rustls_result::CertificateRevocationListParseError,
+            };
+
+            let client_cert_verifier = match client_cert_verifier_holder.0.take() {
+                None => {
+                    return NullParameter;
```
jsha, line 578: I think the `AlreadyUsed` result code (introduced in the Acceptor PR) would be a better fit here.

src/cipher.rs
```diff
@@ -517,26 +610,37 @@ pub struct rustls_client_cert_verifier {
...
     ) -> *const rustls_client_cert_verifier {
         ffi_panic_boundary! {
-            let store: &RootCertStore = try_ref_from_ptr!(store);
-            let client_cert_verifier = AllowAnyAuthenticatedClient::new(store.clone());
+            let mut client_cert_verifier_holder = match BoxCastPtr::to_box(builder) {
+                None => {
+                    return null() as *const _;
+                },
+                Some(x) => x,
+            };
```
jsha, line 637: As above, I think `try_mut_from_ptr!` is more appropriate here. Thanks to the magic of macros, it also handles early-returning null if the input parameter is null.

src/cipher.rs
```diff
@@ -557,6 +661,101 @@ impl rustls_client_cert_verifier {
...
+);
+
+impl CastPtr for rustls_client_cert_verifier_optional_builder {
+    type RustType = AllowAnyAnonymousOrAuthenticatedClientHolder;
+}
+
+impl BoxCastPtr for rustls_client_cert_verifier_optional_builder {}
+
+impl rustls_client_cert_verifier_optional_builder {
+    /// Create a new optional client certificate verifier builder for the root store.
```
jsha, line 683: I'm finding that the adjectives here are hard to parse. Which is optional: the client certificate, the verifier, or the builder? I happen to know that the client certificate (which will be supplied by the handshake) is optional.

Looking back, probably the fault is in the naming of the existing types (client_cert_verifier_optional and client_cert_verifier). Since we're making breaking changes maybe we should align these with the names of the Rust types? E.g. `rustls_allow_any_authenticated_client_builder_new` and `rustls_allow_any_anonymous_or_authenticated_client_new`? A mouthful to be sure but has the advantage of consistency; and the matching comments might be a little easier to understand?

Also a nit: "using the root store" rather than "for the root store".

jsha (changes requested):
Generally looks great, thanks so much for working on this.

I agree with ctz that the upstream API might be better; maybe it would even benefit from a builder pattern? E.g.:

```
let client_cert_verifier = VerifierBuilder::new().with_crl(xyz).allow_anonymous();
let client_cert_verifier2 = VerifierBuilder::new().with_crl(xyz).require_certificate();
```

In terms of how this would be used, I'd expect the CRL(s) to be updated with some frequency, but the comments here and on https://github.com/rustls/rustls/pull/1326 don't make it immediately clear how to do that. Is the idea that you're always using the Acceptor API, and each time you select / build a ServerConfig, you provide it with a client cert verifier that is provided with the most up-to-date CRLs you've got?

cpu:
> Generally looks great, thanks so much for working on this.

Thanks for the review! Much appreciated. I'll work through the feedback ASAP.

> I agree with ctz that the upstream API might be better; maybe it would even benefit from a builder pattern?

That does seem better. I think my preference would be to try and land this branch as-is and then look at iterating on the Rustls-side API as part of the next release since I suspect it will be a breaking change and we're queuing a few of those for a 0.22.x release.

Does that seem reasonable or would you prefer we hold this work to avoid reworking the API twice on the ffi side?

> In terms of how this would be used, I'd expect the CRL(s) to be updated with some frequency, but the comments here and on https://github.com/rustls/rustls/pull/1326 don't make it immediately clear how to do that. Is the idea that you're always using the Acceptor API, and each time you select / build a ServerConfig, you provide it with a client cert verifier that is provided with the most up-to-date CRLs you've got?

Yes I think that's the best approach with the existing implementation :+1: 

My original goal was to have the verifier configured with a `CrlProvider` trait (https://github.com/rustls/webpki/blob/3f57243a26b80bfebba6b25d07b914bc49ed813e/src/verify_cert.rs#L21-L29) that could abstract the process of providing CRLs as needed. That ended up being a dead end; I couldn't find a good way to express that the lifetime of the produced CRL needed to last the duration of the overall verification. Changing things so that the verifier provided the CRLs as an argument to webpki when initiating the verification resolved this very naturally (and matches how intermediates/trust anchors were handled).

I suspect there's room for improvements here but it was starting to feel like the energy folks had to dedicate to this effort was waning and the simpler approach was something we could get out the door.

Does that make sense? I'm hopeful there's still utility to this feature even with that limitation in mind.

jsha:
> my preference would be to try and land this branch as-is and then look at iterating on the Rustls-side API as part of the next release

Yep, very agree on this.

> > always using the Acceptor API, and each time you select / build a ServerConfig, you provide it with a client cert verifier that is provided with the most up-to-date CRLs you've got?
> Yes I think that's the best approach with the existing implementation

Sounds good. This does mean the server needs to know in advance all the CRLs it could possibly expect to see in any client certificate, but I think that is probably a reasonable requirement given how I expect client certificates are used in practice.

src/cipher.rs
cpu: I switched to using `Option<X>` instead of a newtype. I agree it seems slightly nicer than the newtype :+1:

src/cipher.rs
cpu: Thanks for catching that (I guess it's too subtle for valgrind?)

I switched over to using `try_mut_from_ptr!` throughout.

src/cipher.rs
cpu: Switched to `AlreadyUsed`. Thanks!

src/cipher.rs
cpu: Done!

src/cipher.rs
cpu: >  Since we're making breaking changes maybe we should align these with the names of the Rust types?

That sounds good to me. I agree that the wordyness is a fair cost to pay for having easier to understand types that can be better mapped back to their Rustls equivalents. 

To try and keep this easy to review I broke the renaming into a separate commit ahead of introducing the builder type + CRL support.

> Also a nit: "using the root store" rather than "for the root store".

Fixed :+1:

cpu:
> @github-actions rustls-ffi / ensure-header-updated (pull_request) Failing after 28s

Oops, forgot to regenerate the `.h` after going back to fix-up a comment review point I missed.

I'll update + push.

src/cipher.rs
```diff
@@ -509,102 +511,296 @@ impl rustls_root_cert_store {
...
 
+/// A builder for a `rustls_allow_any_authenticated_client_verifier`. This builder object can be
+/// used to configure certificate revocation lists, and then turned into a
+/// `rustls_allow_any_authenticated_client_verifier` once ready.
+pub struct rustls_allow_any_authenticated_client_builder {
+    _private: [u8; 0],
+}
+
+impl CastPtr for rustls_allow_any_authenticated_client_builder {
+    // XXX: contained value is consumed even on error, so this can contain None. but the caller
```
jsha, line 522: In my personal lexicon, XXX means "resolve before merge." I think this comment can be deleted now that this is an intentional part of the API? Or turned into a "note"?

src/cipher.rs
```diff
@@ -509,102 +511,296 @@ impl rustls_root_cert_store {
...
+            let store: &RootCertStore = try_ref_from_ptr!(store);
+            let client_cert_verifier = Some(AllowAnyAuthenticatedClient::new(store.clone()));
+            BoxCastPtr::to_mut_ptr(client_cert_verifier)
+        }
+    }
+
+    /// Add one or more certificate revocation lists (CRLs) to the client certificate verifier by
+    /// reading the CRL content from the provided buffer of PEM encoded content.
+    ///
+    /// This function may return an error if the provided buffer is not valid PEM encoded content,
```
jsha, line 553:
```suggestion
    /// This function returns an error if the provided buffer is not valid PEM encoded content,
```

Super nitpicky but I think it's better to guarantee it _will_ return an error. :D

src/cipher.rs
```diff
@@ -509,102 +511,296 @@ impl rustls_root_cert_store {
...
-impl ArcCastPtr for rustls_client_cert_verifier {}
+impl ArcCastPtr for rustls_allow_any_authenticated_client_verifier {}
 
-impl rustls_client_cert_verifier {
-    /// Create a new client certificate verifier for the root store. The verifier
-    /// can be used in several rustls_server_config instances. Must be freed by
+impl rustls_allow_any_authenticated_client_verifier {
+    /// Create a new allow any authenticated client certificate verifier from a builder.
+    ///
+    /// The builder is consumed and is no longer valid.
```
jsha, line 617:
```suggestion
    /// The builder is consumed and cannot be used again, but must still be freed.
```

src/cipher.rs
```diff
@@ -509,102 +511,296 @@ impl rustls_root_cert_store {
...
+        builder: *mut rustls_allow_any_authenticated_client_builder,
+        crl_pem: *const u8,
+        crl_pem_len: size_t,
+    ) -> rustls_result {
+        ffi_panic_boundary! {
+            let client_cert_verifier_builder: &mut Option<AllowAnyAuthenticatedClient> = try_mut_from_ptr!(builder);
+
+            let crl_pem: &[u8] = try_slice!(crl_pem, crl_pem_len);
+            let crls_der = match crls(&mut Cursor::new(crl_pem)) {
+                Ok(vv) => vv.into_iter().map(UnparsedCertRevocationList).collect::<Vec<UnparsedCertRevocationList>>(),
```
jsha, line 566:
```suggestion
            let crls_der: Vec<UnparsedCertRevocationList> = match crls(&mut Cursor::new(crl_pem)) {
                Ok(vv) => vv.into_iter().map(UnparsedCertRevocationList).collect(),
```

This is a stylistic thing, but I prefer type annotations on the variable declaration over turbofish. Some reasons:

 - Turbofish involves a lot of punctuation, including extra nesting of `<>`, which is hard to read.
 - Turbofish usually winds up in a nested expression, and so contributes to rightward pressure and line wrapping.
  - As long as we're explicitly writing out the type rather than allowing it to be inferred, it's better to write it in one consistent place up front. Otherwise, the reader has to look through potentially many nested branches to find the turbofish operator.

This is slightly aside from the question of how frequently to use type annotations on variable declarations _in general_, a matter on which I know opinions differ. :D

There's one other case of this in `rustls_client_cert_verifier_optional_builder_add_crl` on line 712.

src/cipher.rs
```diff
@@ -509,102 +511,296 @@ impl rustls_root_cert_store {
...
+/// A builder for a `rustls_allow_any_anonymous_or_authenticated_client_verifier`. This builder
+/// object can be used to configure certificate revocation lists, and then turned into a
+/// `rustls_allow_any_anonymous_or_authenticated_client_verifier` once ready.
+pub struct rustls_allow_any_anonymous_or_authenticated_client_builder {
+    _private: [u8; 0],
+}
+
+impl CastPtr for rustls_allow_any_anonymous_or_authenticated_client_builder {
+    // XXX: contained value is consumed even on error, so this can contain None. but the caller
+    // still needs to free it
```
jsha, line 667: Same comment as above: remove this XXX?

jsha (changes requested):
Looking great! Almost there.

src/cipher.rs
cpu: Switched to a `NOTE: ...` - it might be slightly redundant with the API but I found the comment helpful as someone relatively new to the C/Rust boundary and expectations w.r.t `free`.

src/cipher.rs
cpu: Can't argue with that :-D Fixed.

src/cipher.rs
cpu: Fixed. Thanks.

src/cipher.rs
cpu: Also made into a `NOTE:`.

src/cipher.rs
cpu: >This is a stylistic thing, but I prefer type annotations on the variable declaration over turbofish.

I switched this over, but also want to mention that this preference is opposite the feedback I've been getting on my rustls + webpki PRs. As you note I think it often comes down to a stylistic choice but the arguments you make in favour of explicit annotations seem sensible to me. 

I don't have a strong preference one way or the other. It might be something we should try to hash out in a `CONTRIBUTING.md` update over in the main Rustls repo to try and get some consistency and to avoid new contributors having to flip/flop based on who reviews their PR first :-) I've been collecting up some other notes I'd like to turn into a PR at some point. I can tag you when that happens if you'd like.

cpu:
> rustls-ffi / ensure-header-updated (pull_request) Failing

D'oh. `cbindgen` version mismatch between local (0.24.5) and CI (0.24.3). I'll switch to matching CI and regenerate.

cpu:
>> D'oh. cbindgen version mismatch between local (0.24.5) and CI (0.24.3). I'll switch to matching CI and regenerate.
>
> @github-actions rustls-ffi / ensure-header-updated (pull_request) Failing 

I regenerated with 0.24.3 and on the next run CI was using `cbindgen` 0.24.5! I think separate from this PR we might need to explicitly pin `cbindgen` to one version. I'm not sure why it's changing between runs but it seems to require some luck to match up.

jsha (approved):
🎸

cpu:
I can open a PR updating the `Cargo.toml` version to 0.11.0 and updating the `CHANGELOG.md` tomorrow morning. I want to make sure I take some care capturing the changes for the changelog.

cpu:
> I can open a PR updating the Cargo.toml version to 0.11.0 and updating the CHANGELOG.md tomorrow morning.

https://github.com/rustls/rustls-ffi/pull/338

PR #325, opened 2023-07-07
server: use into for SignatureScheme.

cpu:
This branch resolves a `TODO` in `server.rs`'s `sigschemes` function, replacing a `map`'d `match` with `into`.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -585,27 +585,7 @@ impl rustls_server_config_builder {
 
 // Turn a slice of u16's into a vec of SignatureScheme as needed by rustls.
 fn sigschemes(input: &[u16]) -> Vec<SignatureScheme> {
-    use rustls::SignatureScheme::*;
-    input
-        .iter()
-        .map(|n| match n {
-            // TODO: Once rustls 0.20.0+ is released, we can use `.into()` instead of this match.
-            0x0201 => RSA_PKCS1_SHA1,
-            0x0203 => ECDSA_SHA1_Legacy,
-            0x0401 => RSA_PKCS1_SHA256,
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -585,27 +585,7 @@ impl rustls_server_config_builder {
...
-            0x0603 => ECDSA_NISTP521_SHA512,
-            0x0804 => RSA_PSS_SHA256,
-            0x0805 => RSA_PSS_SHA384,
-            0x0806 => RSA_PSS_SHA512,
-            0x0807 => ED25519,
-            0x0808 => ED448,
-            n => SignatureScheme::Unknown(*n),
-        })
-        .collect()
+    input.iter().map(|n| (*n).into()).collect()
```
cpu, line 588: It would be nice to use `map(Into::into)` here but the trait `From<&u16>` is not implemented for `SignatureScheme`, just `From<u16>` and I couldn't find a nicer way to deref except with a map that explicilty `(*n)`'s. Maybe I'm overlooking a trick?

cpu:
> rustls-ffi / Clippy nightly (optional) (pull_request) Failing after 39s

As mentioned in [this other comment](https://github.com/rustls/rustls-ffi/pull/324#issuecomment-1625593611) I believe this is a FP.

src/server.rs
ctz: perhaps `Iter::copied` first? that should give an iter over `u16` rather than `&u16`

cpu: :bulb: That was exactly what I was missing. Thanks @ctz!

cpu:
This feels trivial enough to merge with one +1 :rocket:

Issue #327, opened 2023-07-10
Format C code with clang-format


ctz:
We have a checked-in `.clang-format`, so let's:

- run that in CI (`clang-format --dry-run --Werror ` on all .c/.h files), and expose that as `make format-check` or something
- let people run it locally with `make format`?

I suggest that we exclude `src/rustls.h` from this because it's already the subject of format rules from cbindgen?

cpu:
I don't mind picking up this issue but it might be nice to land some of the open PRs first to avoid a lot of churn rebasing each branch afterwards.

cpu:
https://github.com/rustls/rustls-ffi/pull/329

PR #329, opened 2023-07-10
clang-format all C code, enforce formatting in CI.

cpu:
## Description

This branch builds on the already checked-in `.clang-format`, making it easier to use locally and enforcing that C code in-repo is always formatted using it w/ a CI check.

I haven't attempted to extend the Windows Makefile with similar functionality on the assumption that all development is happening on Unix-like-systems and not Windows. 

Resolves https://github.com/rustls/rustls-ffi/issues/327

### Makefile: rm empty space.
Simple tidy commit.

### Makefile: fix PHONY targets.
The `all`, `clean`, and `test` aren't traditional file targets: they don't produce a built artifact on disk from other files. If there were
to be a file with a matching name found in the workdir these targets wouldn't perform their jobs as expected, thinking the workdir was up-to-date.

This commit marks these targets as "PHONY" targets, telling Make that they're not associated with produced files and are meta-targets for running helpful commands.

### Makefile: add format-check and format targets.
This commit adds two new PHONY targets:

* `make format-check` for enforcing that all .c/.h files are correctly formatted. This will be run in CI in subsequent commits once it will pass.

* `make format` for reformatting all .c/.h files according to the clang-format profile. This can be run by developers to format code.

Both targets use `find` to locate all suitable `.h` and `.c` files. Crucially we configure `find` to ignore the `target` build directory, as well as the `src/rustls.h` file that is generated by `cbindgen`.

### tests: universally apply clang-format.
This commit uses the new `make format` target to re-format all of the .c/.h files using `clang-format`. Afterwards the `make format-check` passes, where previously it flagged the re-formatted files in this commit as needing formatting.

### ci: enforce C formatting w/ make format-check.
This commit extends the existing format task in CI that was enforcing the use of `cargo fmt` to also call `make format-check` to enforce the use of `clang-format` for C code.


cpu:
>  cpu force-pushed the cpu-clang-format-all-the-things branch from 4d0f519 to 21e200e 

Resolved conflicts.

Issue #333, opened 2023-07-11
usage of Arc<T> where T is not Send or Sync


jsha:
Currently `cargo clippy` produces this error:

```
$ cargo clippy 
    Checking rustls-ffi v0.10.0 (/home/jsha/rust/rustls-ffi)
error: usage of `Arc<T>` where `T` is not `Send` or `Sync`
   --> src/lib.rs:396:23
    |
396 |         Arc::into_raw(Arc::new(src)) as *const _
    |                       ^^^^^^^^^^^^^
    |
    = help: consider using `Rc<T>` instead or wrapping `T` in a std::sync type like `Mutex<T>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `#[deny(clippy::arc_with_non_send_sync)]` on by default
```

The relevant code is:

```
pub(crate) trait CastConstPtr {
    type RustType;

    fn cast_const_ptr(ptr: *const Self) -> *const Self::RustType {
        ptr as *const _
    }
}

pub(crate) trait CastConstPtr {
    type RustType;

    fn cast_const_ptr(ptr: *const Self) -> *const Self::RustType {
        ptr as *const _
    }
}

/// Anything that qualifies for CastPtr also automatically qualifies for
/// CastConstPtr. Splitting out CastPtr vs CastConstPtr allows us to ensure
/// that Arcs are never cast to a mutable pointer.
impl<T, R> CastConstPtr for T
where
    T: CastPtr<RustType = R>,
    R: Send,
{
    type RustType = R;
}

pub(crate) trait ArcCastPtr: CastConstPtr + Sized + Send + Sync {
    ...
    fn to_const_ptr(src: Self::RustType) -> *const Self {
        Arc::into_raw(Arc::new(src)) as *const _
    }
```

I think the clippy finding is a good one. Somewhere along the line we should probably be guaranteeing that the `RustType` for an `ArcCastPtr` is at least `Send`. I tried a couple of quick stabs at expressing this in the type system but haven't yet succeeded, so I'm filing this issue to track it.

PR #334, opened 2023-07-11
clippy: fix on nightly toolchain

jsha:
Right now on the nightly toolchain, clippy gives an error. This doesn't block merges because we mark the nightly clippy check as optional, and only require the clippy check for a specified toolchain version.

I've filed #333 to address the error. In this PR I disable that particular clippy lint and bump the required toolchain version to 1.73.0.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -177,10 +177,14 @@ jobs:
       - name: Install rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.67.1
+          toolchain: 1.70.0
           components: clippy
       - name: Check clippy
-        run: cargo clippy --workspace -- -D warnings
+        # We allow unknown lints here because sometimes the nightly job
+        # (below) will have a new lint that we want to suppress.
+        # If we suppress (e.g. #![allow(clippy::arc_with_non_send_sync)]),
+        # we would get an unknown-lint error from older clippy versions.
+        run: cargo clippy --workspace -- -D warnings -A unknown-lints
 
   clippy-nightly-optional:
     name: Clippy nightly (optional)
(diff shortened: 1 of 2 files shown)


cpu (approved):
Thanks for the quick fix :+1:

jsha:
In CI, was getting:

```
error: could not download nonexistent rust version `1.73.0-x86_64-unknown-linux-gnu`: could not download file from 'https://static.rust-lang.org/dist/rust-1.73.0-x86_64-unknown-linux-gnu.tar.gz.sha256' to '/home/runner/.rustup/tmp/ez9sp9x7zg6d3ws7_file': http request returned an unsuccessful status code: 404
```

I realized I pulled the latest version from my `nightly` toolchain locally, which of course is not a released version. Bumped it down to 1.72.0.

jsha:
Heh, and [1.71 is still in beta](https://raw.githubusercontent.com/rust-lang/rust/master/RELEASES.md). Bumped it further down to 1.70. Also I realized we have the same fixed version for the `cargo fmt` check so I bumped that as well.

jsha:
Now that version 1.70.0 is running for the "required" clippy, I get this:

```
error: unknown lint: `clippy::arc_with_non_send_sync`
 --> src/lib.rs:5:10
  |
5 | #![allow(clippy::arc_with_non_send_sync)]
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `-D unknown-lints` implied by `-D warnings`
```

In the latest revision, I turn on `-A unknown-lints` for the required clippy (1.70.0) only. This allows unrecognized lints in the older clippy; but they will trigger an advisory error in the newer clippy (even though it won't block merging).

ctz (approved):

cpu (approved):
> In the latest revision, I turn on -A unknown-lints for the required clippy (1.70.0) only. This allows unrecognized lints in the older clippy; but they will trigger an advisory error in the newer clippy (even though it won't block merging).

Sounds like a reasonable workaround. Thanks for chasing this down.

PR #335, opened 2023-07-13
ci: explicitly pin cbindgen version for diff check.

cpu:
This branch installs an explicitly pinned version of `cbindgen` before performing the diff check of the `ensure-header-updated` workflow step.

Previously we used the version of `cbindgen` that was installed on the base GitHub runner image, which can be hard to match to the version used in local development and may change out from under us.

Since there can be non-semantic differences in output between point releases of `cbindgen` that will fail the diff check it's important developers can predict the exact version that will used in CI in order to match it when generating an updated `.h` locally.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -111,6 +111,12 @@ jobs:
           persist-credentials: false
       - name: Install nightly rust toolchain
         uses: dtolnay/rust-toolchain@nightly
+      - name: Install cbindgen
+        # Pin the installed version of cbindgen so that local usage can be
+        # reliably matched to CI. There can be non-semantic differences in
+        # output between point releases of cbindgen that will fail this check
+        # otherwise.
+        run: cargo install cbindgen --force --version 0.24.5
       - run: touch src/lib.rs
       - run: cbindgen --version
       - run: make src/rustls.h


.github/workflows/test.yaml
```diff
@@ -111,6 +111,12 @@ jobs:
           persist-credentials: false
       - name: Install nightly rust toolchain
         uses: dtolnay/rust-toolchain@nightly
+      - name: Install cbindgen
+        # Pin the installed version of cbindgen so that local usage can be
+        # reliably matched to CI. There can be non-semantic differences in
+        # output between point releases of cbindgen that will fail this check
+        # otherwise.
+        run: cargo install cbindgen --force --version 0.24.5
```
cpu, line 119: Opened as a draft because I'm not sure this is the best approach:

* It will slow down CI having to build cbindgen from src
* We have to maintain the version pin in the yaml
* It seems like in most runs it's replacing the same version with the same version, e.g.

> Replaced package `cbindgen v0.24.5` with `cbindgen v0.24.5` (executable `cbindgen`)

Maybe a better tact is going back through the runs where I was seeing CI occasionally use 0.24.3 and figure out why that was happening...

jsha: Now I'm really confused; where was cbindgen coming from before? It's not part of the Rust toolchain. And the https://github.com/dtolnay/rust-toolchain action doesn't seem to mention it explicitly.

Perhaps it's something that GHA runners happen to have preinstalled as a convenience to users? Weird if so; and definitely suggests we should pin a specific version.

jsha: Indeed! https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#packages

cpu: > Now I'm really confused; where was cbindgen coming from before? It's not part of the Rust toolchain.

Oh weird! I had been operating under the assumption it was part of the nightly toolchain :dizzy_face: Of course that doesn't make sense since I had to explicitly install it locally.

> Perhaps it's something that GHA runners happen to have preinstalled as a convenience to users? 

Yup, that looks to be the case:

https://github.com/actions/runner-images/blob/48a11cfaddc00044ad4af1a9162d872f3d695b47/images/linux/Ubuntu2204-Readme.md?plain=1#L159

> Weird if so; and definitely suggests we should pin a specific version.

OK - I'll flip this out of draft status. I think it's probably the best option. I don't think a dedicated 3rd party action like https://github.com/marketplace/actions/cbindgen-action will be worthwhile in this instance.

jsha: This looks good but the commit message should be updated to reflect that cbindgen is not part of the nightly toolchain.

cpu: Fixed up the commit message + the PR description.

jsha (approved):

PR #341, opened 2023-08-10
track upstream Rustls 0.22.x alpha changes.

cpu:
## Description

This branch tracks the breaking API changes that have accrued in the upstream Rustls, Webpki and Rustls-pemfile repos for the pending 0.22.0 release.

I've chosen to try and break categories of changes into unique commits, but this means that intermediate commits won't build: they're all-or-nothing with the updated dependency versions.

TODO:
* [x] removing cargo patch once rustls 0.22.0-alpha.0 published on crates.io
* [x] rustdoc comments
* [x] better error handling for verifier builder
* [x] debug and fix crash for server w/ AUTH_CERT
* [x] rebase for #353 
* [x] rebase for https://github.com/rustls/rustls/pull/1552
* [x] update for rustls alpha.6
  * [x] webpki client verifier updates
    * [x] revocation depth
    * [x] unknown status policy
    * [x] root hints
  * [x] support new server cert verifier builder
    * [x] roots
    * [x] CRLs
    * [x] revocation depth
    * [x] unknown status policy
* [x] plan for custom crypto providers? - deferred for now
* [x] plan for aws-lc-rs provider?  - deferred for now
* [x] ????
* [x] profit


src/cipher.rs
```diff
@@ -510,296 +511,145 @@ impl rustls_root_cert_store {
...
-                Some(x) => x,
+            let mut builder = WebPkiClientVerifier::builder(client_verifier_builder.roots)
+                .with_crls(client_verifier_builder.crls);
+            if client_verifier_builder.allow_anonymous {
+                builder = builder.allow_unauthenticated();
+            }
+
+            let verifier = match builder.build() {
+                Ok(v) => v,
+                Err(_) => return null() as *const _, // TODO(@cpu): handle this better.
```
cpu, line 639: @jsha Is the right way to handle the fallibility of the `build()` operation to rework this function to return a `rustls_result` and use an "out pointer arg" for the builder? :thinking: 

As-is when returning `*const rustls_client_cert_verifier` the best I can do is give back `NULL` when there's an error and it eats up whatever the underlying `Err(err)` was.

cpu:
> cpu requested review from jsha and ctz now

Tagged for early feedback, but there's no rush to review this. The upstream 0.22.0 release is a little bit out yet.

jsha:
>  Is the right way to handle the fallibility of the build() operation to
rework this function to return a rustls_result and use an "out pointer arg"
for the builder?

Yep, the pattern I've used is "rustls_result always gets the return
position if there is a possible error." Functions that can't error can
return a pointer for convenience.

And to refine on that a bit further: functions that can't error _for
reasons other than null inputs or internal panics_ can return a pointer for
convenience.

cpu:
> Yep, the pattern I've used is "rustls_result always gets the return position if there is a possible error." Functions that can't error can return a pointer for convenience.

:ballot_box_with_check: Thanks! I will fix that up.

cpu:
> :ballot_box_with_check: Thanks! I will fix that up.

Reworked the error handling, added a first pass at updated rustdoc comments.

src/cipher.rs
```diff
@@ -510,296 +509,181 @@ impl rustls_root_cert_store {
...
-    pub extern "C" fn rustls_client_cert_verifier_optional_builder_new(
+    pub extern "C" fn rustls_web_pki_client_cert_verifier_builder_new(
         store: *const rustls_root_cert_store,
-    ) -> *mut rustls_allow_any_anonymous_or_authenticated_client_builder {
+    ) -> *mut rustls_web_pki_client_cert_verifier_builder {
         ffi_panic_boundary! {
-            let store: &RootCertStore = try_ref_from_ptr!(store);
-            let client_cert_verifier = Some(AllowAnyAnonymousOrAuthenticatedClient::new(store.clone()));
-            BoxCastPtr::to_mut_ptr(client_cert_verifier)
+            let store: Arc<RootCertStore> = try_arc_from_ptr!(store);
```
cpu, line 584: I'm seeing a crash running the client server integration tests that I think is happening on this line. Specifically it's happening with the first test that sets an `AUTH_CERT` for the server to enforce mandatory client authentication. 

```
$> make clean && make && AUTH_CERT=testdata/minica.pem gdb --args ./target/server testdata/localhost/cert.pem testdata/localhost/key.pem
<snipped>
(gdb) run
Starting program: /home/daniel/Code/Rust/rustls-ffi/target/server testdata/localhost/cert.pem testdata/localhost/key.pem
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/nix/store/ld03l52xq2ssn4x0g5asypsxqls40497-glibc-2.37-8/lib/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
0x00000000004d3fc9 in std::panicking::try::h608246b5500614b5 ()
(gdb) bt
#0  0x00000000004d3fc9 in std::panicking::try::h608246b5500614b5 ()
#1  0x000000000040a35d in rustls_web_pki_client_cert_verifier_builder_new ()
#2  0x00000000004062ef in main (argc=<optimized out>, argv=<optimized out>) at tests/server.c:293
(gdb) q
```

Will debug further.

cpu:
:construction: Work in progress :construction: 

* Updated this branch for the latest available alphas
* Rolled in the `pki-types` changes from upstream
* Broke the client/server tests - [see this comment](https://github.com/rustls/rustls-ffi/pull/341#discussion_r1343078550)

src/cipher.rs
cpu: <details>
<summary>better backtrace:</summary>

```
$> make clean && make PROFILE=debug && AUTH_CERT=testdata/minica.pem gdb --args ./target/server testdata/localhost/cert.pem testdata/localhost/key.pem
<snipped>
(gdb) run
Starting program: /home/daniel/Code/Rust/rustls-ffi/target/server testdata/localhost/cert.pem testdata/localhost/key.pem
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/nix/store/ld03l52xq2ssn4x0g5asypsxqls40497-glibc-2.37-8/lib/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
alloc::sync::{impl#22}::clone<rustls::webpki::anchors::RootCertStore> (self=0x7fffffff4d50) at /build/rustc-1.72.0-src/library/alloc/src/sync.rs:1530
1530	/build/rustc-1.72.0-src/library/alloc/src/sync.rs: No such file or directory.
(gdb) bt
#0  alloc::sync::{impl#22}::clone<rustls::webpki::anchors::RootCertStore> (self=0x7fffffff4d50) at /build/rustc-1.72.0-src/library/alloc/src/sync.rs:1530
#1  0x0000000000406c74 in rustls_ffi::ArcCastPtr::to_arc<rustls_ffi::cipher::rustls_root_cert_store> (ptr=0x6be410) at src/lib.rs:377
#2  0x000000000040996b in rustls_ffi::try_arc_from<rustls_ffi::cipher::rustls_root_cert_store, rustls::webpki::anchors::RootCertStore> (from=0x6be410) at src/lib.rs:456
#3  0x0000000000407541 in rustls_ffi::cipher::{impl#16}::rustls_web_pki_client_cert_verifier_builder_new::{closure#0} () at src/cipher.rs:584
#4  0x0000000000414d0e in std::panicking::try::do_call<rustls_ffi::cipher::{impl#16}::rustls_web_pki_client_cert_verifier_builder_new::{closure_env#0}, *mut rustls_ffi::cipher::rustls_web_pki_client_cert_verifier_builder> (data=0x7fffffff4ee0)
    at /build/rustc-1.72.0-src/library/std/src/panicking.rs:500
#5  0x000000000041609b in __rust_try ()
#6  0x0000000000413e13 in std::panicking::try<*mut rustls_ffi::cipher::rustls_web_pki_client_cert_verifier_builder, rustls_ffi::cipher::{impl#16}::rustls_web_pki_client_cert_verifier_builder_new::{closure_env#0}> (f=...) at /build/rustc-1.72.0-src/library/std/src/panicking.rs:464
#7  0x0000000000411c97 in std::panic::catch_unwind<rustls_ffi::cipher::{impl#16}::rustls_web_pki_client_cert_verifier_builder_new::{closure_env#0}, *mut rustls_ffi::cipher::rustls_web_pki_client_cert_verifier_builder> (f=...) at /build/rustc-1.72.0-src/library/std/src/panic.rs:142
#8  0x0000000000408ad3 in rustls_ffi::cipher::rustls_web_pki_client_cert_verifier_builder::rustls_web_pki_client_cert_verifier_builder_new (store=0x6be410) at src/cipher.rs:583
#9  0x00000000004046bf in main (argc=<optimized out>, argv=<optimized out>) at tests/server.c:293
```
</details>

cpu: https://github.com/rust-lang/rust/blob/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/alloc/src/sync.rs#L1500-L1535

cpu: It's certainly the case that I've messed up the `Arc<RootCertStore>` handling introduced in this branch, but I'm yet sure what changed between this most recent force push and https://github.com/rustls/rustls-ffi/commit/07532681e838e6e1d8ecd9890273e14eeacb22b2 when it at least appeared to work :thinking: 

The test seems to work if I introduce a clumsy diff like this, but I'm not confident it's a well reasoned change:
```diff
diff --git a/src/cipher.rs b/src/cipher.rs
index b5e9eb4..b62bdeb 100644
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -578,12 +578,13 @@ impl rustls_web_pki_client_cert_verifier_builder {
     /// ownership of the pointed-to data.
     #[no_mangle]
     pub extern "C" fn rustls_web_pki_client_cert_verifier_builder_new(
-        store: *const rustls_root_cert_store,
+        store: *mut rustls_root_cert_store,
     ) -> *mut rustls_web_pki_client_cert_verifier_builder {
         ffi_panic_boundary! {
-            let store: Arc<RootCertStore> = try_arc_from_ptr!(store);
+            //let store: Arc<RootCertStore> = try_arc_from_ptr!(store);
+            let store: &mut RootCertStore = try_mut_from_ptr!(store);
             let builder = ClientCertVerifierBuilder {
-                roots: store,
+                roots: Arc::new(store.clone()),
                 crls: Vec::default(),
                 allow_anonymous: false,
             };
```

jsha: You're on the right track. I wasn't able to get this completely working before having to sign off, but two key things:

 - try_arc_from_ptr is what you want for any time you want to do clones of an arc. See docs on [ArcCastPtr::to_arc](https://github.com/rustls/rustls-ffi/blob/6c24d4829e1d596ce03e6097145e023be29acf78/src/lib.rs#L356-L371), which it calls under the hood. It does some refcounting fixups.
 - Only one of BoxCastPtr or ArcCastPtr should be implemented for a type; the fact that BoxCastPtr was implemented prevented the type system from catching the problem here.

The below doesn't compile yet, but should indicate a good direction for progress yet; happy to help more tomorrow.

```
diff --git a/src/cipher.rs b/src/cipher.rs
index b5e9eb4a..1325f8d3 100644
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -443,8 +443,6 @@ impl CastPtr for rustls_root_cert_store {
     type RustType = RootCertStore;
 }
 
-impl BoxCastPtr for rustls_root_cert_store {}
-
 impl ArcCastPtr for rustls_root_cert_store {}
 
 impl rustls_root_cert_store {
@@ -453,10 +451,10 @@ impl rustls_root_cert_store {
     /// Caller must add root certificates with rustls_root_cert_store_add_pem.
     /// <https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.empty>
     #[no_mangle]
-    pub extern "C" fn rustls_root_cert_store_new() -> *mut rustls_root_cert_store {
+    pub extern "C" fn rustls_root_cert_store_new() -> *const rustls_root_cert_store {
         ffi_panic_boundary! {
             let store = rustls::RootCertStore::empty();
-            BoxCastPtr::to_mut_ptr(store)
+            ArcCastPtr::to_const_ptr(&store)
         }
     }
 
@@ -503,7 +501,7 @@ impl rustls_root_cert_store {
     #[no_mangle]
     pub extern "C" fn rustls_root_cert_store_free(store: *mut rustls_root_cert_store) {
         ffi_panic_boundary! {
-            let store = try_box_from_ptr!(store);
+            let store = try_arc_from_ptr!(store);
             drop(store)
         }
     }
@@ -519,7 +517,7 @@ impl CastPtr for rustls_client_cert_verifier {
     type RustType = Arc<dyn ClientCertVerifier>;
 }
 
-impl BoxCastPtr for rustls_client_cert_verifier {}
+impl ArcCastPtr for rustls_client_cert_verifier {}
 
 impl rustls_client_cert_verifier {
     /// Free a `rustls_client_cert_verifier` previously returned from
@@ -583,7 +581,7 @@ impl rustls_web_pki_client_cert_verifier_builder {
         ffi_panic_boundary! {
             let store: Arc<RootCertStore> = try_arc_from_ptr!(store);
             let builder = ClientCertVerifierBuilder {
-                roots: store,
+                roots: store.clone(),
                 crls: Vec::default(),
                 allow_anonymous: false,
             };
diff --git a/src/lib.rs b/src/lib.rs
index c002925f..9c0b5f56 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -329,7 +329,8 @@ where
 }
 
 // An implementation of BoxCastPtr means that when we give C code a pointer to the relevant type,
-// it is actually a Box.
+// it is actually a Box. At most one of BoxCastPtr or ArcCastPtr should be implemented for a given
+// type.
 pub(crate) trait BoxCastPtr: CastPtr + Sized {
     fn to_box(ptr: *mut Self) -> Option<Box<Self::RustType>> {
         if ptr.is_null() {
diff --git a/src/server.rs b/src/server.rs
index 4a46319f..94cd271b 100644
--- a/src/server.rs
+++ b/src/server.rs
@@ -166,7 +166,7 @@ impl rustls_server_config_builder {
     ) {
         ffi_panic_boundary! {
             let builder: &mut ServerConfigBuilder = try_mut_from_ptr!(builder);
-            let verifier: Arc<dyn ClientCertVerifier> = try_ref_from_ptr!(verifier).clone();
+            let verifier: Arc<dyn ClientCertVerifier> = try_arc_from_ptr!(verifier);
             builder.verifier = verifier;
         }
     }
```

cpu: > You're on the right track. I wasn't able to get this completely working before having to sign off, but two key things:

Thanks Jsha!

> Only one of BoxCastPtr or ArcCastPtr should be implemented for a type; the fact that BoxCastPtr was implemented prevented the type system from catching the problem here.

Oh interesting! I had done the same in my earlier revision that seemed to work. I suspect that it was just more subtly broken than the current branch :grimacing: 

I will pick up your diff today and see where I can get. Appreciate the guidance :bow:

cpu: I have a few questions after playing with the diff some more. Sorry if some of this is obvious. I'm still trying to get the right mental model in place and find the dance between Rust types, the helper macros, and the C semantics tricky.

I've pushed some revisions and I think it's closer: the rust code builds. The `server.c` code fails from discarding the `const` qualifier on the `rustls_root_cert_store` pointer based on the "Root Cert store `ArcCastPtr`" problem discussed below.

### Root Cert store `ArcCastPtr`

```diff
-impl BoxCastPtr for rustls_root_cert_store {}
-
 impl ArcCastPtr for rustls_root_cert_store {}
 ```
 
The `ArcCastPtr` change on `rustls_root_cert_store` implies changing `rustls_root_cert_store_new` and `rustls_root_cert_store_free` to work with `*const rustls_root_cert_store`, but that's not compatible with the needs of `rustls_root_cert_store_add_pem` where we want a `*mut rustls_root_cert_store` to use with `try_mut_from_ptr!`, right? I'm not sure how to reconcile this now that I understand `BoxCastPtr` and `ArcCastPtr` are mutually exclusive.

### Explicit `Arc<RootCertStore>` clone

```diff
 impl rustls_client_cert_verifier {
     /// Free a `rustls_client_cert_verifier` previously returned from
@@ -583,7 +581,7 @@ impl rustls_web_pki_client_cert_verifier_builder {
         ffi_panic_boundary! {
             let store: Arc<RootCertStore> = try_arc_from_ptr!(store);
             let builder = ClientCertVerifierBuilder {
-                roots: store,
+                roots: store.clone(),
                 crls: Vec::default(),
                 allow_anonymous: false,
             };
```

Clippy finds the `clone` here unnecessary. Is it missing something?

### `Arc<Arc<dyn ClientCertVerifier>>`

```diff
diff --git a/src/server.rs b/src/server.rs
index 4a46319f..94cd271b 100644
--- a/src/server.rs
+++ b/src/server.rs
@@ -166,7 +166,7 @@ impl rustls_server_config_builder {
     ) {
         ffi_panic_boundary! {
             let builder: &mut ServerConfigBuilder = try_mut_from_ptr!(builder);
-            let verifier: Arc<dyn ClientCertVerifier> = try_ref_from_ptr!(verifier).clone();
+            let verifier: Arc<dyn ClientCertVerifier> = try_arc_from_ptr!(verifier);
             builder.verifier = verifier;
         }
     }
```

Using `try_arc_from_ptr!` here gives us an `Arc<Arc<dyn ClientCertVerifier>>`, whereas `try_ref_from_ptr!` returns the `Arc<dyn ClientCertVerifier>` the `builder.base.with_client_cert_verifier` fn that the `verifier` field is ultimately used for requires. I thought perhaps changing the `rustls_client_cert_verifier`'s `RustType` to be `dyn ClientCertVerifier` instead of `Arc<dyn ClientCertVerifier>` would be the way to use `try_arc_from_ptr!` here, but then the size of the type isn't known at compile time. Using a reference introduces complicates related to the lifetime of the reference, so it doesn't seem like the right approach either. I think this is the first instance of a `CastPtr` for a trait instead of a struct so there isn't existinv code to crib from. 

### Out parameter for an `ArcCastPtr`

In [earlier discussions](https://github.com/rustls/rustls-ffi/pull/341#issuecomment-1673836497) we landed on using an out parameter pattern for the client cert verifier's build operation, since its fallible and needs to use the traditional return for a `rustls_result`.

The existing instances of this pattern are using types that are `BoxCastPtr`, but the new builder is now only `ArcCastPtr` and so can't use the `BoxCastPtr::set_mut_ptr` helper. Is the correct approach to set the out parameter type to be `*mut *const rustls_client_cert_verifier` (as opposed to `*mut *mut rustls_client_cert_verifier` ala the other `BoxCastPtr` instances) and then use an unsafe block similar to the`set_mut_ptr` helper? Something like:

```rust
 unsafe {
      *verifier_out = ArcCastPtr::to_const_ptr(verifier)
  }
```

jsha: > that's not compatible with the needs of rustls_root_cert_store_add_pem where we want a *mut rustls_root_cert_store to use with try_mut_from_ptr!, right?

Yep, that's right; so there's some more API design work to be done here. The basic idea of BoxCastPtr / ArcCastPtr is "any time we give C a pointer of type X, it really corresponds to a Box<T> under the hood" or "... really corresponds to an Arc<T> under the hood".

If you look at it from the Rust side, you have the same problem: If you've got an `Arc<T>` you can't modify it. Instead, you do all the modification you need ahead of time, and then wrap the resulting T in an `Arc`. That's why our `Arc<T>` types are represented on the C side with a `*const`, to clarify that "can't modify" situation.

So I think what we need here is another builder pattern: create a root_cert_store_builder, add certs to it, then "build" it by wrapping it in an Arc.

(more soon)

jsha: > Clippy finds the clone here unnecessary. Is it missing something?

Clippy is correct here. Instead of cloning the Arc we got out of `ArcCastPtr::to_arc`, we can simply hand over ownership of it to the ClientCertVerifierBuilder. By doing that, the Arc will no longer be dropped at the end of the function (because it was moved instead), so the refcount won't be decremented. It will remain at least 2 (1 for the C side, 1 for this particular reference on the Rust side).

> Out parameter for an ArcCastPtr

Your approach here is correct. I think to refine on it, we should add a `set_mut_ptr` method on the ArcCastPtr trait and use that.

> Arc<Arc<dyn ClientCertVerifier>>

Looking at this in more detail now.

jsha: > whereas try_ref_from_ptr! returns the Arc<dyn ClientCertVerifier>

Actually, `try_ref_from_ptr!(verifier:  *const rustls_client_cert_verifier)` is returning an `&Arc<dyn ClientCertVerifier>`, which is correct and what I'd expect. We can then call `clone()` on that reference (since [`Arc::clone`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.clone) takes `&self`).

Perhaps the other site where I proposed `try_arc_from_ptr!` would be equally well served by `try_ref_from_ptr!` plus `clone()`; we might actually be able to do away with `try_arc_from_ptr!` entirely, which would be a nice clarification.

cpu: I think I've rolled in all of this feedback/learning now :rocket: The commit history has gotten a little bit messy but otherwise I think it's in a workable state now :sweat_smile: 

> Clippy is correct here. Instead of cloning the Arc we got out of ArcCastPtr::to_arc, we can simply hand over ownership of it to the ClientCertVerifierBuilder

This ended up falling away after switching to `try_ref_from_ptr!` in more places.

> Actually, try_ref_from_ptr!(verifier:  *const rustls_client_cert_verifier) is returning an &Arc<dyn ClientCertVerifier>, which is correct and what I'd expect. We can then call clone() on that reference

Thanks, that's what I landed on too. I should have been more precise in my description :+1: 

> Your approach here is correct. I think to refine on it, we should add a set_mut_ptr method on the ArcCastPtr trait and use that.

Done :ballot_box_with_check: 

> we might actually be able to do away with try_arc_from_ptr! entirely, which would be a nice clarification.

I think I've managed to achieve this :crossed_fingers:

jsha:
I've paged back in some more of this context and I'm realizing there's a challenge / mis-design about how the `CastConstPtr` / `ArcCastPtr` traits work.

`CastConstPtr` means "on the Rust side, we can turn `const *rustls_foo` (C) into `*const rustls::Foo` (Rust) with a simple `as` cast."

`ArcCastPtr` means "on the Rust side, `const *rustls_foo` is the output of [Arc::into_raw](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw) called on a `rustls::Foo`." And because `CastConstPtr` is a trait bound, it also carries the same meaning.

Those definitions seem potentially conflicting because the layout of an `Arc<rustls::Foo>` is different from the layout of a `*const rustls::Foo`. The [layout of an `Arc` is defined here](https://doc.rust-lang.org/1.72.0/src/alloc/sync.rs.html#249-252) and [here on ArcInner](https://doc.rust-lang.org/1.72.0/src/alloc/sync.rs.html#326-339).

The tricky thing is that `Arc::into_raw` returns a pointer to the `data` field of ArcInner - that's past the beginning of the struct by two `usize`s (the strong and weak counts). That's okay, because [`Arc::from_raw`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.from_raw) knows how to then adjust backwards in memory by two `usize`s to reconstruct the original Arc. And of course it's more useful for `into_raw` to return a pointer to T than a pointer to an private ArcInner struct that happens to contain your T at some offset.

So, given a `const * rustls_foo` there are two casts we can perform:

 1. A simple `as` cast to `*const rustls_foo`, pointing to the inner data
 2. `Arc::from_raw`, which gets us a real Arc, including the strong and weak counts. But this runs into some subtle problems when we drop that `Arc`. Dropping decrements the strong count, and will very often result in freeing the pointed-to memory. That's why we have ArcCastPtr::to_arc.

So far, we've been allowing both, but I think in practice we should forbid (1) and always reconstruct the "real" type, to avoid confusion. In other words, `ArcCastPtr` should be disjoint from `CastConstPtr` rather than a subset of it.

The other thing to mention here is that the `RustType` associated type in `CastPtr` / `CastConstPtr` / `BoxCastPtr` / `ArcCastPtr` is supposed to be a pointee type. That is, it should not include any `*`, `&`, `Box`, or `Arc` (unless it happens to be doubly-indirect). For instance, here we have the definition for `rustls_certified_key` / `rustls::CertifiedKey`:

https://github.com/rustls/rustls/blob/ce31cd5559fcf1bc938c11c0363a4ef9f0c129f1/src/cipher.rs#L264-L269

```rust
impl CastPtr for rustls_certified_key {
    type RustType = CertifiedKey;
}

impl ArcCastPtr for rustls_certified_key {}
```

That represents a C-side type of `const *rustls_certified_key` and a Rust-side type of `Arc<rustls::CertifiedKey>`. Note that the associated type is `type RustType = CertifiedKey`, not `type RustType Arc<CertifiedKey>`. That's because in the helper functions (e.g. `cast_const_ptr`), we need to be able to tack on the correct punctuation (e.g. to say the return type is `*const Self::RustType`). It's not the clearest way to express things, but it's the way to make the type system do what we want from it.

In this branch, the definition for `rustls_client_cert_verifier` is:

```rust
impl CastPtr for rustls_client_cert_verifier {
    type RustType = Arc<dyn ClientCertVerifier>;
}

impl ArcCastPtr for rustls_client_cert_verifier {}
```

But ideally that should look like:

```rust
impl CastPtr for rustls_client_cert_verifier {
    type RustType = dyn ClientCertVerifier;
}

impl ArcCastPtr for rustls_client_cert_verifier {}
```

The problem is, as you ran into: `dyn ClientCertVerifier` is unsized (`!Sized`). The rule with unsized types is: you can only store them in pointers / smart pointers. For instance `Box<dyn ClientCertVerifier>` is fine; so is `Arc<dyn ClientCertVerifier>` and `&dyn ClientCertVerifier`. That should work fine for us, since we're really only dealing in pointers.

However, our implementations of `ArcCastPtr` / `ConstCastPtr` have `Sized` bounds. Why? I think this is honestly a mistake I made when implementing. I saw an error message saying I couldn't do something because xyz didn't have a size known at compile time, so I added a `+ Sized` bound and that fixed it. Really, the fix should have been something like this:

```rust
 pub(crate) trait CastPtr {
-    type RustType;
+    type RustType: ?Sized;
```

The deal is that [type parameters and associated types are `Sized` by default](https://doc.rust-lang.org/stable/reference/special-types-and-traits.html?#sized). In other words that `type RustType` was really acting like `type RustType: Sized`. But we can explicitly disavow the default `Sized` bound using the special `?Sized` bound.

Adding the `?Sized` bound to `RustType` in both `CastPtr` and `CastConstPtr` allows us to remove the `+ Sized` bound for `ArcCastPtr` and `BoxCastPtr`, which in turn allows us to say `RustType = dyn ClientCertVerifier`.

However, then we run into some other problems: Our `set_mut_ptr` methods take an owned T as a parameter. That's fine when `T` is sized, but when `T` is `dyn ClientCertVerifier`, that doesn't work; function parameters _must_ be sized, and there's no way around that requirement. You can't disavow the `Sized` requirement for function parameters.

We also run into this delightfully esoteric error message, which I vaguely see the direction of but don't really know how to fix:

```
error[E0606]: casting `*const <Self as CastConstPtr>::RustType` as `*const Self` is invalid
   --> src/lib.rs:402:9  
    |
402 |         Arc::into_raw(Arc::new(src)) as *const _
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: vtable kinds may not match
```

I think the net result of all of this is that our best bet might be to double-indirect things: Give the C side a `Box<Arc<dyn ClientCertVerifier>>`. That way we don't need to worry about the complexity of `dyn` types.

Below is a patch I was noodling on where I added the `?Sized` bound. It doesn't work and is probably a dead end, but might be interesting:

```patch
diff --git a/src/cipher.rs b/src/cipher.rs
index dce36138..43cdaed2 100644
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -582,7 +582,7 @@ pub struct rustls_client_cert_verifier {
 }
 
 impl CastPtr for rustls_client_cert_verifier {
-    type RustType = Arc<dyn ClientCertVerifier>;
+    type RustType = dyn ClientCertVerifier;
 }
 
 impl ArcCastPtr for rustls_client_cert_verifier {}
@@ -738,7 +738,7 @@ impl rustls_web_pki_client_cert_verifier_builder {
                 Err(e) => return error::map_verifier_builder_error(e),
             };
 
-            ArcCastPtr::set_mut_ptr(verifier_out, verifier);
+            ArcCastPtr::set_mut_ptr(verifier_out, verifier.as_ref());
 
             rustls_result::Ok
         }
diff --git a/src/lib.rs b/src/lib.rs
index 2c31dd10..75232a54 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,7 +301,7 @@ mod tests {
 /// Implementing this is required in order to use `try_ref_from_ptr!` or
 /// `try_mut_from_ptr!`.
 pub(crate) trait CastPtr {
-    type RustType;
+    type RustType: ?Sized;
 
     fn cast_mut_ptr(ptr: *mut Self) -> *mut Self::RustType {
         ptr as *mut _
@@ -311,7 +311,7 @@ pub(crate) trait CastPtr {
 /// CastConstPtr represents a subset of CastPtr, for when we can only treat
 /// something as a const (for instance when dealing with Arc).
 pub(crate) trait CastConstPtr {
-    type RustType;
+    type RustType: ?Sized;
 
     fn cast_const_ptr(ptr: *const Self) -> *const Self::RustType {
         ptr as *const _
@@ -324,6 +324,8 @@ pub(crate) trait CastConstPtr {
 impl<T, R> CastConstPtr for T
 where
     T: CastPtr<RustType = R>,
+    T: ?Sized,
+    R: ?Sized,
 {
     type RustType = R;
 }
@@ -331,7 +333,7 @@ where
 // An implementation of BoxCastPtr means that when we give C code a pointer to the relevant type,
 // it is actually a Box. At most one of BoxCastPtr or ArcCastPtr should be implemented for a given
 // type.
-pub(crate) trait BoxCastPtr: CastPtr + Sized {
+pub(crate) trait BoxCastPtr: CastPtr {
     fn to_box(ptr: *mut Self) -> Option<Box<Self::RustType>> {
         if ptr.is_null() {
             return None;
@@ -354,7 +356,7 @@ pub(crate) trait BoxCastPtr: CastPtr + Sized {
 // An implementation of ArcCastPtr means that when we give C code a pointer to the relevant type,
 // it is actually a Arc. At most one of BoxCastPtr or ArcCastPtr should be implemented for a given
 // // type.
-pub(crate) trait ArcCastPtr: CastConstPtr + Sized {
+pub(crate) trait ArcCastPtr: CastConstPtr {
     /// Sometimes we create an Arc, then call `into_raw` and return the resulting raw pointer
     /// to C. C can then call rustls_server_session_new multiple times using that
     /// same raw pointer. On each call, we need to reconstruct the Arc. But once we reconstruct the Arc,
diff --git a/src/server.rs b/src/server.rs
index bd0c1064..979f2699 100644
--- a/src/server.rs
+++ b/src/server.rs
@@ -166,7 +166,7 @@ impl rustls_server_config_builder {
     ) {
         ffi_panic_boundary! {
             let builder: &mut ServerConfigBuilder = try_mut_from_ptr!(builder);
-            let verifier = try_ref_from_ptr!(verifier);
+            let verifier = try_arc_from_ptr!(verifier);
             builder.verifier = verifier.clone();
         }
     }
```

cpu:
> I think the net result of all of this is that our best bet might be to double-indirect things: Give the C side a Box<Arc<dyn ClientCertVerifier>>. That way we don't need to worry about the complexity of dyn types.

I took a crack at this. Here's what [the diff](https://github.com/rustls/rustls-ffi/pull/341/commits/f12aa122840eac719c948ce7dea49b003f3cff0e) entailed (_writing this out to check my own understanding_):

* `rustls_client_cert_verifier` is changed to implement `BoxCastPtr` instead of `ArcCastPtr`.
* The `RustType` for the `CastPtr` impl for `rustls_client_cert_verifier` remains `Arc<dyn ClientCertVerifier>` - we're OK with it not being a "pointee" type, because it's double-indirect. That is, a `Box` over the `Arc<dyn ClientCertVerifier>`.
* The `rustls_client_cert_verifier_free` fn impl changes to operate on a `*mut` input pointer argument, which we free by using `BoxCastPtr::to_box` to convert from `*mut rustls_client_cert_verifier` to `Box<Arc<dyn ClientCertVerifier>>`. Internally this ends up being an `as *mut` conversion and the `Box<Arc<dyn ClientCertVerifier>>` is dropped with the stack frame, freeing the `Box` and appropriately decrementing the `Arc` reference counters.
* The `rustls_web_pki_client_cert_verifier_builder_build` fn is changed to accept a `*mut` out-pointer. We build a `Arc<dyn ClientCertVerifier>` with the Rustls `rustls::webpki::ClientCertVerifierBuilder`'s `build` fn, and then assign it to the out pointer with  `BoxCastPtr::set_mut_ptr`. Internally this is constructing a `Box` around the `Arc<dyn ClientCertVerifier>`,  and calling `Box::into_raw` to get the `*mut` pointer to the `Box` for the out pointer assignment.

Separately, I had the same issue with the new `rustls_root_cert_store` adjustment: it was implementing `CastPtr` with `RustType = Arc<RootCertStore>`. I _think_ we don't need the double indirection here since `RootCertStore` isn't a `dyn` trait. So for this instance I tacked on [a diff](f0f97b1fe8fdf18c91524bf401e44eeba26d0146) that:

* Changes the `CastPtr` impl to use `RustType = RootCertStore` - no issues here, `RootCertStore` is `Sized`.
* The `rustls_root_cert_store_builder_build` fn that creates a `rustls_root_cert_store` uses `ArcCastPtr::set_mut_ptr` to wrap the `RootCertStore` in an `Arc<RootCertStore>`. Internally it converts the `Arc<RootCertStore>` to a `*const` written to the out pointer.
* The `rustls_web_pki_client_cert_verifier_builder_new` fn that takes a `*const rustls_root_cert_store` argument now uses `try_arc_from_ptr!` instead of `try_ref_from_ptr!` to reconstitute a `Arc<RootCertStore>` for the `ClientCertVerifierBuilder` being constructed. 

I think this avoids the issues you were highlighting :crossed_fingers:

jsha:
Yep, your summary all looks good. A couple of super pedantic points:

> remains Arc<dyn ClientCertVerifier> - we're OK with it not being a "pointee" type, because it's double-indirect. That is, a Box over the Arc<dyn ClientCertVerifier>.

Really `Arc<dyn ClientCertVerifier>` _is_ a pointee type in this situation because the Box points to it; but yeah this all makes sense.

> The rustls_client_cert_verifier_free fn impl changes to operate on a *mut input pointer argument, which we free by using BoxCastPtr::to_box to convert from *mut rustls_client_cert_verifier to Box<Arc<dyn ClientCertVerifier>>. Internally this ends up being an as *mut conversion 

`to_box` does do an `as *mut` conversion; it follows that with `Box::from_raw`, which is how we get back to a state of "here's a Rust object with ownership, which will cause memory to be freed when it is dropped at the end of the function."

> It's worrying that the mis-designed implementation "works".

I agree; I need to poke at it a bit more and see what the deal is, and if there's anything to improve in terms of detection. It's possible that we weren't exercising "interesting" code paths for `Arc` because we only created one of a thing and didn't clone it?

cpu:
> Yep, your summary all looks good.

Great, thanks for taking a look. I can work on tidying up the commit history now that it seems like we're stabilizing on an overall approach.

I think there might also be one work item left to consider (maybe separate from this branch?): making `ArcCastPtr` disjoint from `CastConstPtr` based on your earlier comment:

> So far, we've been allowing both, but I think in practice we should forbid (1) and always reconstruct the "real" type, to avoid confusion. In other words, ArcCastPtr should be disjoint from CastConstPtr rather than a subset of it.

Should I give that a go?

> Really Arc<dyn ClientCertVerifier> is a pointee type in this situation because the Box points to it

Ah ok, good point (_pun not intended :laughing:_)

> to_box does do an as *mut conversion; it follows that with Box::from_raw, which is how we get back to a state of "here's a Rust object with ownership, which will cause memory to be freed when it is dropped at the end of the function."

ACK :ballot_box_with_check: 

> It's possible that we weren't exercising "interesting" code paths for Arc because we only created one of a thing and didn't clone it?

That would make sense :thinking:

jsha:
> It's worrying that the mis-designed implementation "works"... Do you have any intuition about why that might be?

I looked back at one of the earlier revisions here: 6e988afb55cc8763d52b66eb9f06f191ac4335bf. And it turns out you had already implemented the doubly-indirect solution we eventually landed on: allocate a `Box<Arc<dyn ClientCertVerifier>>`, turn that outer `Box` into a raw `*mut` pointer and hand it over to C:

```rust
impl CastPtr for rustls_client_cert_verifier {
    type RustType = Arc<dyn ClientCertVerifier>;
}

impl BoxCastPtr for rustls_client_cert_verifier {}
```

https://github.com/rustls/rustls-ffi/commit/6e988afb55cc8763d52b66eb9f06f191ac4335bf#diff-48d27698cc708c7efe770fd897e4885efaa0a15cae30d54936068603ba03bbd9R170

```rust
let verifier: Arc<dyn ClientCertVerifier> = try_ref_from_ptr!(verifier).clone();
```

For more detail on what's happening in that line of code, it could be:

```
let verifier_ref: &Arc<dyn ClientCertVerifier> = try_ref_from_ptr!(verifier);
let verifier = verifier_ref.clone();
```

That's taking advantage of the fact that we can get infinitely many read-only references to the contents of a Box, even though we don't _own_ the box. It's also taking advantage of the fact that an `Arc` can be cloned from an `&Arc`.

This could also have been:

```
let verifier_box: Box<Arc<dyn ClientCertVerifier>> = try_box_from_ptr!(verifier); // would also need verifier to be a `*mut`, not a *const
let verifier_ref: &Arc<dyn ClientCertVerifier> = verifier_box.as_ref();
let verifier: Arc<dyn ClientCertVerifier> = verifier.clone();
...
Box::leak(verifier_box); // make sure we don't free the box on drop, since the C side still owns it
```

In practice we only use `try_box_from_ptr!` when we want to really take ownership of the pointer, because it's too easy to leave off the `Box::leak`.

Anyhow, long story short: the mis-design was that it was (and is) possible to implement `BoxCastPtr` and `ArcCastPtr` for the same type, which enables conversions of a pointer to _either_ `Box<T>` or `Arc<T>` even though it's truly only one of those. In your earlier revision, the pointer was truly a Box and we didn't get any valgrind or other errors because you never actually invoked the conversion to `Arc<T>`. The pointer was created with:

```rust

**Comment by cpu** [comment]
>  I can work on tidying up the commit history 

Rebased on `main` and tidied up. As called out in the PR desc the intermediate commits don't build/test cleanly, but it felt like it would be easier to review this way. I suggest we squash merge instead of the usual rebase merge once this branch is ready to go.

I'll try to keep it up to date with the 0.22.x progress in the main Rustls repo.

**Comment by cpu** [comment]
> cpu force-pushed the cpu-rustls-0.22-alpha branch from 48d9e81 to 363c445 

Rebased to rework w/ the new `Castable` trait.

File: src/cipher.rs [hunk 3]
```diff
@@ -23,21 +23,22 @@ use crate::{
...
-/// <https://docs.rs/rustls/latest/rustls/struct.Certificate.html>
-pub struct rustls_certificate {
+/// Corresponds to `CertificateDer` in the Rust pki-types API.
+/// <https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.CertificateDer.html>
+pub struct rustls_certificate<'a> {
     // We use the opaque struct pattern to tell C about our types without
     // telling them what's inside.
     // https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
     _private: [u8; 0],
+    _marker: PhantomData<&'a ()>,
```
cpu, line 34: Calling this out as a more interesting change:

The `RustType` of the `rustls_certificate` type's `Castable` impl gains a lifetime as of the switch to `CertificateDer`, which means that the `rustls_certificate` type needs a lifetime too. 

That in turn requires adding a `PhantomData` marker since otherwise it would be an unbounded lifetime not referenced by a field. Based on [the FFI guide section on opaque structs](https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs) I think adding the phantom data reference is the correct solution here.

src/cipher.rs
```diff
@@ -568,298 +569,184 @@ impl rustls_root_cert_store {
...
+    // https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
     _private: [u8; 0],
 }
 
-impl Castable for rustls_allow_any_authenticated_client_verifier {
-    type Ownership = OwnershipArc;
-    type RustType = AllowAnyAuthenticatedClient;
+pub(crate) struct ClientCertVerifierBuilder {
+    roots: Arc<RootCertStore>,
+    crls: Vec<CertificateRevocationListDer<'static>>,
```
cpu, line 615: Leaving a note here: I'm not sure that using a `'static` lifetime is the correct choice for this type. I'll give this more thought.

cpu: I think I've convinced myself this is correct:

The `CertificateRevocationListDer` type is defined in `pki-types` as a newtype around `Der<'a>`. That in turn is a newtype around `DerInner<'a>` which is an enum with two variants:

```
#[derive(Clone, PartialEq)]
enum DerInner<'a> {
    #[cfg(feature = "alloc")]
    Owned(Vec<u8>),
    Borrowed(&'a [u8]),
}
```

The lifetime that gets carried through the newtypes is specific to the `Borrowed` variant, but we're using the `Owned(Vec<u8>)` variant throughout, matching what was previously `Vec<UnparsedCertRevocationList>`.

Using `'static` as the lifetime binds us to either sticking with the owned variant (that has no associated lifetime) as we've done here, or a `'static` lifetime for borrowed data we can treat equivalent to an owned type from a lifetime perspective.

jsha: Yep, that's exactly right; sorry, I meant to reply on this thread earlier but I forgot!

Also worth noting there's a secondary but similar meaning for `'static` when used in a trait bound, which means "[the receiver can hold onto the type as long as they want](https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html#trait-bound)"

cpu: > Yep, that's exactly right

Great! Glad I came to the right conclusion :-)

> Also worth noting there's a secondary but similar meaning for 'static when used in a trait bound

Ah interesting! I think I've seen that but I don't know that I had a crisp understanding of what it meant.

cpu:
@jsha @ctz Since we're getting close to a 0.22 Rustls release I've updated this branch to use the latest alpha and caught up on the various API changes. I think it's ~reviewable in the current state if you have time for feedback.

Right now the rustls-ffi side is written to assume *ring* as the crypto provider. I think worst case we could release in that state and maintain status-quo, but to get the most out of the new 0.22 release we might want to come up with a plan for how to expose the choice of using `aws-lc-rs`, or a custom crypto provider through this API. Have either of you given this any thought? The `aws-lc-rs` support is probably not too much of a lift. Getting the crypto provider API right might be trickier.

jsha:
Excellent, thanks! Will take a look.

And yeah, it would be good to figure out how to expose the crypto provider choice. I'll do some thinking about that.

cpu:
> cpu force-pushed the cpu-rustls-0.22-alpha branch from 05b1789 to e6b1953 

Resolved conflicts, tidied the commit history a little bit (e.g. using the latest alphas from the start), and adjusted to use `try_take!` where applicable.

src/cipher.rs
```diff
@@ -472,319 +486,651 @@ impl rustls_root_cert_store {
...
+    /// `rustls_web_pki_client_cert_verifier_builder_add_crl` is used to add certificate revocation
+    /// lists (CRLs) to the builder. If CRLs are added, revocation checking will be performed
+    /// for the entire certificate chain unless
+    /// `rustls_web_pki_client_cert_verifier_only_check_end_entity_revocation` is used. Unknown
+    /// revocation status for certificates considered for revocation status will be treated as
+    /// an error unless `rustls_web_pki_client_cert_verifier_allow_unknown_revocation_status` is
+    /// used.
     ///
-    /// This object can then be used to load any CRLs.
+    /// Anonymous unauthenticated clients will not be permitted unless
```
jsha, line 712:
```suggestion
    /// Unauthenticated clients will not be permitted unless
```

I think the change from `anonymous` to `unauthenticated` is a good one; this is a missed spot.

src/cipher.rs
```diff
@@ -472,319 +486,651 @@ impl rustls_root_cert_store {
...
 }
 
-impl Castable for rustls_allow_any_authenticated_client_builder {
+pub(crate) struct ClientCertVerifierBuilder {
+    roots: Arc<RootCertStore>,
+    root_hint_subjects: Vec<DistinguishedName>,
+    crls: Vec<CertificateRevocationListDer<'static>>,
+    revocation_depth: RevocationCheckDepth,
+    revocation_policy: UnknownStatusPolicy,
+    allow_anonymous: bool,
```
jsha, line 686:
```suggestion
    allow_unauthenticated: bool,
```

(and corresponding changes anywhere this field is referenced)

src/cipher.rs
```diff
@@ -472,319 +486,651 @@ impl rustls_root_cert_store {
...
+            client_verifier_builder.root_hint_subjects.clear();
+            rustls_result::Ok
+        }
+    }
+
+    /// Add additional distinguished names to the list of trust anchor hint subjects.
+    ///
+    /// By default, the client cert verifier will use the subjects provided by the root cert
+    /// store configured for client authentication. Calling this function will add to these
+    /// existing hint subjects. Calling this function with empty `subjects` will have no
```
jsha, line 853:
```suggestion
    /// existing hint subjects. Calling this function with an empty `store` will have no
```

src/cipher.rs
```diff
@@ -472,319 +486,651 @@ impl rustls_root_cert_store {
...
+            let client_verifier_builder = try_take!(client_verifier_builder);
 
-            let client_cert_verifier = match client_cert_verifier_builder.take() {
-                None => {
-                    return null() as *const _;
-                },
-                Some(x) => x,
+            let mut builder = WebPkiClientVerifier::builder(client_verifier_builder.roots)
+                .with_crls(client_verifier_builder.crls);
+            if matches!(client_verifier_builder.revocation_depth, RevocationCheckDepth::EndEntity) {
```
jsha, line 891: I think this would be better expressed as an exhaustive match with `RevocationCheckDepth::Chain => {}` as the other branch. That makes it clear when reading this that there's only one other possibility. Same below with UnknownStatusPolicy::Allow / Deny.

src/cipher.rs
```diff
@@ -472,319 +486,651 @@ impl rustls_root_cert_store {
...
+
+/// A built server certificate verifier that can be provided to a `rustls_client_config_builder`
+/// with `rustls_client_config_builder_set_server_verifier`.
+pub struct rustls_server_cert_verifier {
+    _private: [u8; 0],
+}
+
+impl Castable for rustls_server_cert_verifier {
+    type Ownership = OwnershipBox;
+    type RustType = Arc<dyn ServerCertVerifier>;
```
jsha, line 1123: The same comment I mentioned above about fat pointers should be here; or we should put something in the CONTRIBUTING that explains it and link to that.

src/cipher.rs
```diff
@@ -398,72 +407,77 @@ impl rustls_certified_key {
...
 
         Ok(rustls::sign::CertifiedKey::new(parsed_chain, signing_key))
     }
 }
 
-/// A root certificate store.
-/// <https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html>
-pub struct rustls_root_cert_store {
+/// A `rustls_root_cert_store` being constructed. A builder can be modified by,
+/// adding trust anchor root certificates with `rustls_root_cert_store_builder_add_pem`.
```
jsha, line 444:
```suggestion
/// A `rustls_root_cert_store` being constructed. A builder can be modified by
/// adding trust anchor root certificates with `rustls_root_cert_store_builder_add_pem`.
```

jsha (changes requested):
Looking good!

src/cipher.rs
```diff
@@ -747,3 +812,155 @@ impl rustls_web_pki_client_cert_verifier_builder {
...
+
+/// A built server certificate verifier that can be provided to a `rustls_client_config_builder`
+/// with `rustls_client_config_builder_set_server_verifier`.
+pub struct rustls_server_cert_verifier {
+    _private: [u8; 0],
+}
+
+impl Castable for rustls_server_cert_verifier {
+    type Ownership = OwnershipBox;
+    type RustType = Arc<dyn ServerCertVerifier>;
```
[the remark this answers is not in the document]
jsha: I wound up writing this up for CONTRIBUTING.md: https://github.com/rustls/rustls-ffi/pull/364

Assuming we wind up merging that, let's just link to it from here.

src/client.rs
```diff
@@ -312,66 +364,16 @@ impl rustls_client_config_builder {
         }
     }
 
-    /// Use the trusted root certificates from the provided store.
-    ///
-    /// This replaces any trusted roots already configured with copies
-    /// from `roots`. This adds 1 to the refcount for `roots`. When you
-    /// call rustls_client_config_free or rustls_client_config_builder_free,
-    /// those will subtract 1 from the refcount for `roots`.
+    /// Configure the server certificate verifier.
```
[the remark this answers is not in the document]
cpu: Added.

src/cipher.rs
```diff
@@ -398,72 +407,77 @@ impl rustls_certified_key {
...
 
-impl rustls_root_cert_store {
-    /// Create a rustls_root_cert_store. Caller owns the memory and must
-    /// eventually call rustls_root_cert_store_free. The store starts out empty.
-    /// Caller must add root certificates with rustls_root_cert_store_add_pem.
-    /// <https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.empty>
+impl rustls_root_cert_store_builder {
+    /// Create a `rustls_root_cert_store_builder`. Caller owns the memory and must
+    /// eventually call `rustls_root_cert_store_builder_build`, then free the
+    /// resulting `rustls_root_cert_store`.
```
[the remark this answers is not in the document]
cpu: Fixed up.

src/cipher.rs
```diff
@@ -747,3 +812,155 @@ impl rustls_web_pki_client_cert_verifier_builder {
...
+
+impl Castable for rustls_web_pki_server_cert_verifier_builder {
+    type Ownership = OwnershipBox;
+    type RustType = Option<ServerCertVerifierBuilder>;
+}
+
+impl ServerCertVerifierBuilder {
+    /// Create a `rustls_web_pki_server_cert_verifier_builder`. Caller owns the memory and must
+    /// eventually call `rustls_web_pki_server_cert_verifier_builder_build`, then free the
+    /// resulting `rustls_server_cert_verifier`.
```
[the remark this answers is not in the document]
cpu: Fixed up (also for the client verifier)

src/cipher.rs
```diff
@@ -747,3 +812,155 @@ impl rustls_web_pki_client_cert_verifier_builder {
...
+        crl_pem_len: size_t,
+    ) -> rustls_result {
+        ffi_panic_boundary! {
+            let server_verifier_builder: &mut Option<ServerCertVerifierBuilder> = try_mut_from_ptr!(builder);
+
+            let crl_pem: &[u8] = try_slice!(crl_pem, crl_pem_len);
+            let crls_der: Result<Vec<CertificateRevocationListDer>, _> =  crls(&mut Cursor::new(crl_pem)).collect();
+            let crls_der = match crls_der{
+                Ok(vv) => vv,
+                Err(_) => return rustls_result::CertificateRevocationListParseError,
```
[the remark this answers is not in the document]
cpu: > We should detect that condition here and turn it into an error.

Done, and also for the client verifier version.

In both cases I returned `rustls_result::CertificateRevocationListParseError` - it didn't feel significant enough to make a dedicated error but I can revisit if you disagree.

src/cipher.rs
```diff
@@ -747,3 +812,155 @@ impl rustls_web_pki_client_cert_verifier_builder {
...
+            };
+
+            let server_verifier_builder = match server_verifier_builder {
+                None => return AlreadyUsed,
+                Some(v) => v,
+            };
```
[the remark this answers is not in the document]
cpu: Done (and also for the client verifier version).

src/cipher.rs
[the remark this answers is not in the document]
cpu: Thanks! I approved #364 earlier today so I put in a link to where the section will appear once that's merged. I also added the text to the equivalent `Castable` impl for the client cert verifier.

src/cipher.rs
cpu: Fixed up throughout.

src/cipher.rs
cpu: Done!

src/cipher.rs
cpu: Fixed. Thanks!

src/cipher.rs
cpu: Makes sense. I've updated each instance, for the verifier depth and unknown revocation status, for both the client cert and server cert verifier impls.

src/cipher.rs
cpu: Done.

src/cipher.rs
cpu: Fixed. Thanks.

cpu:
> cpu force-pushed the cpu-rustls-0.22-alpha branch from 0b7c359 to 49b4eb9 

Addressed review feedback, updated with the `try_arc_from_ptr` -> `try_clone_arc` change.

jsha (approved):
This looks great. I'm inclined to go ahead and merge this to main, so we can iterate further (e.g. adding crypto providers) without you needing to constantly rebase. 0.22 is coming up soon, and if we really need to do another rustls-ffi release targeting 0.21, we can do it off of a branch. What do you think?

src/cipher.rs
```diff
@@ -430,34 +430,44 @@ impl rustls_certified_key {
     }
 }
 
-/// A root certificate store.
-/// <https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html>
-pub struct rustls_root_cert_store {
+/// A `rustls_root_cert_store` being constructed. A builder can be modified by
```
ctz, line 443: paragraph break after headline?

src/cipher.rs
```diff
@@ -430,34 +430,44 @@ impl rustls_certified_key {
...
+    type RustType = Option<RootCertStoreBuilder>;
 }
 
-impl rustls_root_cert_store {
-    /// Create a rustls_root_cert_store. Caller owns the memory and must
-    /// eventually call rustls_root_cert_store_free. The store starts out empty.
-    /// Caller must add root certificates with rustls_root_cert_store_add_pem.
-    /// <https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.empty>
+impl rustls_root_cert_store_builder {
+    /// Create a `rustls_root_cert_store_builder`. Caller owns the memory and may
```
ctz, line 465: paragraph break after headline?

ctz (approved):
i agree on merging this.

src/cipher.rs
cpu: Fixed. Thanks!

src/cipher.rs
cpu: Fixed. Thanks!

cpu:
When CI finishes I'll squash merge this. Since the individual commits don't build/test cleanly this feels better than rebase merging for this situation.

Issue #345, opened 2023-09-24
Building as shared library


kpcyrd:
Hello!

I've started looking into this repository while trying to build curl using `--with-rustls`, and getting this error:

```
configure: error: --with-rustls was specified but could not find rustls.
```

I've attempted to build a `librustls` package out of rustls-ffi but noticed it only contains an `.a` file:

```
drwxr-xr-x root/root         0 2023-09-24 21:07 usr/
drwxr-xr-x root/root         0 2023-09-24 21:07 usr/include/
-rwxr-xr-x root/root     79764 2023-09-24 21:07 usr/include/rustls.h
drwxr-xr-x root/root         0 2023-09-24 21:07 usr/lib/
-rwxr-xr-x root/root  30385654 2023-09-24 21:07 usr/lib/librustls.a
```

Is there a way to build a .so file instead?

Thanks!

cpu:
Hi @kpcyrd 

> Is there a way to build a .so file instead?

So far there hasn't been explicit support for building a dynamic library (see [this comment](https://github.com/rustls/rustls-ffi/pull/274#issuecomment-1684071375) from Jsha for the rationale). It sounds like it's workable with the yet-to-be-merged cargo-c workflow, but isn't something that's tested+supported outright.

cpu:
If you're feeling brave testing+feedback about using rustls-ffi as a .so would be helpful input :-)

kpcyrd:
I've used the patch from #274 and created this PKGBUILD:

```sh
# Maintainer: kpcyrd <kpcyrd[at]archlinux[dot]org>

pkgname=librustls
pkgver=0.11.0
pkgrel=1
pkgdesc="Use rustls from languages other than Rust"
arch=('x86_64')
license=('Apache-2.0' 'MIT')
url='https://github.com/rustls/rustls-ffi'
makedepends=(
  cargo-c
  rust
)
provides=('librustls.so')
options=(!lto)
source=(https://github.com/rustls/rustls-ffi/archive/v${pkgver}/${pkgname}-${pkgver}.tar.gz
        Cargo.lock
        shared-linking.patch)
sha256sums=('0eeac3b916286cce35a3f32f3fd11f54ad2584a32bb67ac41c0c563c7c62c98b'
            '06d3b4df897c45a3320e73dca2fcea1cde75f29b1c0a90f4930b403c7c7dc3b1'
[...]

PR #346, opened 2023-09-28
Commit Cargo.lock to repository

kpcyrd:
Related to #345


.github/workflows/test.yaml
```diff
@@ -122,6 +122,17 @@ jobs:
       - run: make src/rustls.h
       - run: git diff --exit-code
 
+  audit:
+    name: Audit Cargo.lock for vulnerable dependencies
```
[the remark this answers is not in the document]
kpcyrd: Done!

djc:
I think you'll want to make sure CI jobs use `--locked` to that the `Cargo.lock` is actually getting tested.

cpu (approved):
This looks good to me and seems in keeping with [the updated guidance from the Rust project](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html) on committing lockfiles :+1:

kpcyrd:
I've added `--locked` in multiple places, if `Cargo.lock` is compatible with `Cargo.toml` it should be picked up automatically, if it's out-of-sync multiple jobs would now start failing (including `make`).

Makefile
```diff
@@ -5,6 +5,7 @@ else
 endif
 
 CARGO ?= cargo
+CARGOFLAGS += --locked
```
cpu, line 8: I think the `Makefile.Windows` and `CMakeLists.txt` files might need similar updates.

kpcyrd: Done

cpu (approved):

cpu:
@jsha Do you have any reservations about checking in the `Cargo.lock` for this project?

jsha (approved):
Looks good to me. Thanks!

PR #353, opened 2023-10-13
rework `CastPtr`, `CastConstPtr`, `BoxCastPtr`, `ArcCastPtr`

### LINKED ISSUE #349: Prevent implementing both ArcCastPtr and BoxCastPtr on the same type
In https://github.com/rustls/rustls-ffi/pull/341#issuecomment-1745711553 we ran into a problem: both `ArcCastPtr` and `BoxCastPtr` were implemented for `rustls_client_cert_verifier`. This turned out fine because only the `Box`-related casts were used, but it could have turned out badly if, for instance, a pointer was created as a `Box` and later (incorrectly) cast to an `Arc`.

We should see if we can use the Rust type system to make it impossible to implement both of these for the same type.

Also, right now we have helper traits `CastPtr` / `ConstConstPtr` that are trait bounds for `ArcCastPtr` and `BoxCastPtr`. So for each C-facing type, we need to do two trait impls. We could probably simplify things by removing the trait bound, and instead adding a blanket impl of `CastConstPtr` for all `ArcCastPtr` and all `BoxCastPtr`; and a blanket impl of `CastPtr` for all `BoxCastPtr`. That way we would only need to impl one of `ArcCastPtr` / `BoxCastPtr` for any given Rust type:

today:
```rust
impl CastPtr for rustls_root_cert_store_builder {
    type RustType = Option<RootCertStoreBuilder>;
}

impl BoxCastPtr for rustls_root_cert_store_builder {}
```

future?
```rust
impl BoxCastPtr for rustls_root_cert_store_builder {
    type RustType = Option<RootCertStoreBuilder>;
}
```

cpu:
### rework CastPtr, CastConstPtr, BoxCastPtr, ArcCastPtr

This commit reworks the existing `CastPtr`, `CastConstPtr`, `BoxCastPtr`, and `ArcCastPtr` traits into a new `Castable` trait with an associated `Ownership`, constrained to be a type implementing the new `OwnershipMarker` trait. Three implementations of this `OwnershipMarker` trait are offered: `OwnershipBox`, `OwnershipArc` and `OwnershipRef`.

The net result is that the type system is now able to enforce our invariant that a single `Castable` type can't be cast to a pointer of more than one type of ownership (e.g. both an `Arc` and a `Box`).  Implementing `Castable` for the same type with different `Ownership`'s or `RustType`'s will be rejected by the compiler as a conflicting trait implementation. We always implement a distinct type per-ownership model, and per rust type.

Along the way crate-internal documentation for the traits and associated free-standing helper `fs`s were reworked for clarity/consistency.

Resolves #349 

### lib: remove pub export of macros

Previously the helper macros in `src/lib.rs` were marked `macro_export`, making them part of the public API. Since these were
meant to be crate internal, we also annotated the macros as `doc(hidden)` to avoid them appearing in the API docs. I suspect this was done before `pub(crate)` visibility was an option.

This commit removes the `macro_export` and `doc(hidden)` attributes and uses a `pub(crate)` re-export to make the macros available to crate-internal users without making them part of the public API.

This also uncovered that the `try_mut_slice!` macro wasn't being used anywhere and so it is removed outright.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -18,8 +18,9 @@ use rustls_pemfile::{certs, crls, pkcs8_private_keys, rsa_private_keys};
 use crate::error::{map_error, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
-    ffi_panic_boundary, try_box_from_ptr, try_mut_from_ptr, try_ref_from_ptr, try_slice,
-    ArcCastPtr, BoxCastPtr, CastConstPtr, CastPtr,
+    ffi_panic_boundary, free_arc, to_arc_const_ptr, to_boxed_mut_ptr, try_box_from_ptr,
+    try_mut_from_ptr, try_ref_from_ptr, try_slice, Castable, OwnershipArc, OwnershipBox,
+    OwnershipRef,
 };
 use rustls_result::{AlreadyUsed, NullParameter};
 
@@ -33,7 +34,8 @@ pub struct rustls_certificate {
     _private: [u8; 0],
 }
 
-impl CastPtr for rustls_certificate {
+impl Castable for rustls_certificate {
+    type Ownership = OwnershipRef;
     type RustType = Certificate;
 }
 
@@ -66,7 +68,8 @@ pub struct rustls_supported_ciphersuite {
     _private: [u8; 0],
 }
 
-impl CastPtr for rustls_supported_ciphersuite {
+impl Castable for rustls_supported_ciphersuite {
+    type Ownership = OwnershipRef;
     type RustType = SupportedCipherSuite;
 }
 
@@ -262,11 +265,10 @@ pub struct rustls_certified_key {
     _private: [u8; 0],
 }
 
-impl CastPtr for rustls_certified_key {
+impl Castable for rustls_certified_key {
+    type Ownership = OwnershipArc;
     type RustType = CertifiedKey;
 }
-
-impl ArcCastPtr for rustls_certified_key {}
 
 impl rustls_certified_key {
     /// Build a `rustls_certified_key` from a certificate chain and a private key.
@@ -366,7 +368,7 @@ impl rustls_certified_key {
             } else {
                 new_key.ocsp = None;
             }
-            *cloned_key_out = ArcCastPtr::to_const_ptr(new_key);
+            *cloned_key_out = to_arc_const_ptr(new_key);
             rustls_result::Ok
         }
     }
@@ -380,7 +382,7 @@ impl rustls_certified_key {
     #[no_mangle]
     pub extern "C" fn rustls_certified_key_free(key: *const rustls_certified_key) {
         ffi_panic_boundary! {
-            rustls_certified_key::free(key);
+            free_arc(key);
         }
     }
 
@@ -443,11 +445,10 @@ pub struct rustls_root_cert_store {
     _private: [u8; 0],
 }
 
-impl CastPtr for rustls_root_cert_store {
+impl Castable for rustls_root_cert_store {
+    type Ownership = OwnershipBox;
     type RustType = RootCertStore;
 }
-
-impl BoxCastPtr for rustls_root_cert_store {}
 
 impl rustls_root_cert_store {
     /// Create a rustls_root_cert_store. Caller owns the memory and must
@@ -458,7 +459,7 @@ impl rustls_root_cert_store {
     pub extern "C" fn rustls_root_cert_store_new() -> *mut rustls_root_cert_store {
         ffi_panic_boundary! {
             let store = rustls::RootCertStore::empty();
-            BoxCastPtr::to_mut_ptr(store)
+            to_boxed_mut_ptr(store)
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -292,179 +292,259 @@ mod tests {
...
+/// since this would be a conflicting trait implementation and rejected by the compiler.
+///
+/// This trait allows us to avoid using `as` in most places, and ensures that when we cast, we're
+/// preserving const-ness, and casting between the correct types. Implementing this is required in
+/// order to use `try_ref_from_ptr!` or `try_mut_from_ptr!` and several other helpful cast-related
+/// conversion helpers.
+pub(crate) trait Castable {
+    /// Indicates whether to use `Box` or `Arc` when giving a pointer to C code for the underlying
+    /// `RustType`.
+    type CastType: CastTypeMarker;
```
cpu, line 333: "CastType" might not be the best name for this idea and the associated marker types but it's the one I picked for a first pass.

It feels a bit like this is more of a wrapper type for the cast, so perhaps the names should have "wrapper" somewhere?

src/cipher.rs
```diff
@@ -33,7 +33,8 @@ pub struct rustls_certificate {
     _private: [u8; 0],
 }
 
-impl CastPtr for rustls_certificate {
+impl Castable for rustls_certificate {
```
cpu, line 36: This type and `rustls_supported_ciphersuite` are unique in that unlike most of the other snake types they only implemented `CastPtr` and not one of `BoxCastPtr` or `ArcCastPtr`.

I think using `ArcCastPtr` makes sense here since the new `Castable` type requires choosing a `CastType`. It didn't feel like breaking out a separate representation for this case was worthwhile, but its possible I'm overlooking something that would require more care here.

src/connection.rs
```diff
@@ -545,9 +544,7 @@ impl rustls_connection {
     #[no_mangle]
     pub extern "C" fn rustls_connection_free(conn: *mut rustls_connection) {
         ffi_panic_boundary! {
-            let conn: &mut Connection = try_mut_from_ptr!(conn);
-            // Convert the pointer to a Box and drop it.
-            drop(unsafe { Box::from_raw(conn) });
+            free_box(conn);
```
cpu, line 547: I think this was a spot where we missed using the `BoxCastPtr::to_box` helper previously, and had re-implemented it inline.

src/cipher.rs
jsha: Currently `rustls_certificate` is only given to the user in two places:

 - https://github.com/rustls/rustls-ffi/blob/8caf0c8e2c239251f3d3d2f307ede461af50e984/src/rustls.h#L913-L921
 - https://github.com/rustls/rustls-ffi/blob/8caf0c8e2c239251f3d3d2f307ede461af50e984/src/rustls.h#L1355-L1368

In both of these cases, the underlying Rust type is a reference, not a Box or an Arc. I think what we need here is a new type, `RefCastPtr`, that expresses a non-owned reference.

If in the future we provide any methods that give the C code an owned pointer to a `rustls::Certificate` (a `Box<Certificate>` or an `Arc<Certificate>`), we would want to provide a separate C type (`rustls_owned_certificate` or `rustls_certificate_owned`) to prevent any confusion.

src/lib.rs
jsha: Yeah I definitely got turned around by the `CastType` name a couple times playing with this code.

How about `Ownership`? With possible values OwnershipBox, OwnershipArc, and OwnershipRef.

Or `PointerStyle`?

We could also potentially change `RustType` to `Pointee` to go with `PointerStyle`.

cpu: > How about Ownership? With possible values OwnershipBox, OwnershipArc, and OwnershipRef.

I like those :+1:

How do you feel about the top-level `Castable` name? Should we lay out some paint swatches for that bike shed while we're at it? :sweat_smile:

jsha:
I like Castable a lot!

src/cipher.rs
cpu: > In both of these cases, the underlying Rust type is a reference, not a Box or an Arc.

Ahh, ok that makes sense, and I think I understand why nothing blew up when I had this arranged incorrectly based on how the types are used.

>  I think what we need here is a new type, RefCastPtr, that expresses a non-owned reference.

I used the suggestion from the other comment thread for a `OwnershipRef` to use as the `Ownership` associated type in the `Castable` impl for these types. 

> This type and rustls_supported_ciphersuite are unique

There was actually a third instance too (also now fixed up): `rustls_client_hello`. All three now implement `Castable` with `Ownership = OwnershipRef`.

Since `try_ref_from_ptr` and the underlying `try_from` are generic over the `Ownership` of the `Castable` we can use `try_ref_from_ptr` for these types but not the arc or box specific variants. I think that's what we want :+1:

src/lib.rs
cpu: >> How do you feel about the top-level Castable name? 
>
> I like Castable a lot!

Great :man_artist: - Resolving this thread since I pushed a revision with the new `Ownership` names.

jsha (approved):
This is a great improvement, thanks!

cpu:
Thanks for all your help along the way :bow:

PR #360, opened 2023-11-15
lib: add try_take! macro, handling AlreadyUsed Options

cpu:
In several places we represent something that could be consumed as an `Option<T>`. When we try to use it, we `take()` the option, match the result, and return `rustls_result::AlreadyUsed` if `take()` returned `None`.

Since this pattern is becoming more common with the use of more builder patterns (particularly in https://github.com/rustls/rustls-ffi/pull/341) this commit adds a `try_take!` macro that can do this repetitive work for us.

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -12,7 +12,8 @@ use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::server::rustls_server_config;
 use crate::{
     ffi_panic_boundary, free_box, rustls_result, set_boxed_mut_ptr, to_box, to_boxed_mut_ptr,
-    try_arc_from_ptr, try_callback, try_mut_from_ptr, try_ref_from_ptr, Castable, OwnershipBox,
+    try_arc_from_ptr, try_callback, try_mut_from_ptr, try_ref_from_ptr, try_take, Castable,
+    OwnershipBox,
 };
 use rustls_result::NullParameter;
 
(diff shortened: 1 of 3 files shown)


cpu:
> rustls-ffi / Check minimum versions (pull_request) Failing after 34s

Ah, it looks like several CI tasks are failing in this branch (and also in `main`'s [scheduled CI runs](https://github.com/rustls/rustls-ffi/actions/runs/6877424258)) due to https://github.com/rust-lang/rust/issues/117693. We fixed this in Rustls w/ https://github.com/rustls/rustls/pull/1582, but that fix is only present in `0.22.0-alpha.4`. 

I addressed the breakage in https://github.com/rustls/rustls-ffi/pull/341 with the update to alpha-4 (https://github.com/rustls/rustls-ffi/pull/341/commits/4f9ecfd04da80295b04f124030f68585795207e1) but fixing `main` will require a Rustls point release in the 0.21.x series. Maybe worthwhile? `rustls-ffi` probably isn't the only downstream crate that will break from this :thinking:

jsha:
A point release for rustls 0.21.x would be great. In the meantime I think we can unbreak the rustls-ffi build by pinning a specific nightly rust release rather than the latest nightly. We may also be able to move some of the affected tasks to a stable release.

ctz (approved):

jsha (approved):
This is great. One piece of naming feedback: The other `try_` macros do non-mutating conversions. This one mutates, specifically calling `take()`. I think `try_take!` would be a better name and make it clearer what's happening.

cpu:
> I think try_take! would be a better name and make it clearer what's happening.

Good idea, renamed! The branch name has fallen out of date but I updated everything else. Thanks for the review.

PR #364, opened 2023-11-16
doc: document handling of dynamically sized types

jsha:
Following up on the conversation in https://github.com/rustls/rustls-ffi/pull/341#issuecomment-1749605419 about how to handle `Arc<dyn Foo>` in the Rustls API. I decided to document this in the CONTRIBUTING.md; we can link to this sub-heading as needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,3 +26,51 @@ If you're making changes to rustls-ffi, you'll need
 regenerate the header file:
 
     make src/rustls.h
+
+## Dynamically Sized Types
+
+Many types exposed in this API are wrapped in a `Box` or an `Arc`, which can be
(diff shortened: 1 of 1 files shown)


cpu (approved):
This looks great! I just had a couple small nits to flag.

I think extensive coverage of the `try_xxx` macros is probably too much detail, but maybe it would be worth a brief mention of their existence to direct an interested reader to their location in `lib.rs`? WDYT?

jsha:
> I think extensive coverage of the try_xxx macros is probably too much detail, but maybe it would be worth a brief mention of their existence to direct an interested reader to their location in lib.rs? WDYT?

This is a good idea but I don't think it's tightly bound to the DSTs explanation so I'll do it in a followup.

PR #368, opened 2023-11-24
WIP: track rustls alpha.5

cpu:
Opened as a draft pending publication of rustls alpha.5 in crates.io.

Brings in the change to use `pki_types` for `ServerName`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,12 @@ read_buf = ["rustls/read_buf"]
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "=0.22.0-alpha.4", features = [ "ring" ]}
-rustls-webpki = "0.102.0-alpha.6"
-pki-types = { package = "rustls-pki-types", version = "0.2.1", features = ["std"] }
+rustls = { version = "=0.22.0-alpha.5", features = [ "ring" ]}
+pki-types = { package = "rustls-pki-types", version = "0.2.2", features = ["std"] }
+webpki = { package = "rustls-webpki", version = "=0.102.0-alpha.7", features = ["std"] }
 libc = "0.2"
 sct = "0.7"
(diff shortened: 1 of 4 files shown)


src/client.rs
```diff
@@ -255,14 +255,22 @@ impl ServerCertVerifier for Verifier {
...
+        server_name: &pki_types::ServerName<'_>,
         ocsp_response: &[u8],
         _now: UnixTime,
     ) -> Result<ServerCertVerified, rustls::Error> {
         let cb = self.callback;
         let server_name: Cow<'_, str> = match server_name {
-            rustls::ServerName::DnsName(n) => n.as_ref().into(),
-            rustls::ServerName::IpAddress(ip) => ip.to_string().into(),
+            pki_types::ServerName::DnsName(n) => n.as_ref().into(),
+            // TODO(@cpu): HACK
```
cpu, line 265: Trying to think through the right change to make this more ergonomic. I think `pki_types::IpAddr` should `impl From<IpAddr> for std::net::IpAddr ? :thinking:

cpu: https://github.com/rustls/pki-types/pull/14

Cargo.toml
```diff
@@ -23,12 +23,12 @@ read_buf = ["rustls/read_buf"]
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "=0.22.0-alpha.4", features = [ "ring" ]}
-rustls-webpki = "0.102.0-alpha.6"
-pki-types = { package = "rustls-pki-types", version = "0.2.1", features = ["std"] }
+rustls = { version = "=0.22.0-alpha.6", features = [ "ring" ]}
+pki-types = { package = "rustls-pki-types", version = "0.2.3", features = ["std"] }
```
cpu, line 27: This changes to 0.2.3 instead of staying at 0.2.2 because we're using the `ServerName::to_str` method that's only available in 0.2.3+

jsha (approved):
Thanks!

PR #372, opened 2023-12-03
docs: update changelog

cpu:
This branch adds an updated `CHANGELOG.md` covering the to-be-published 0.12.0 release. Additionally it fixes one misnamed function caught during review of the changes since the last release, and a fn that was accidentally dropped from the `rustls.h` interface.

To generate the changelog content I pulled the 0.11.0 `src/rustls.h` and the `src/rustls.h` from `main`, ran them through `gcc` to remove comments and post-processed to remove whitespace and `#` lines. I then reviewed the diff and tried to account for all of the changes in the text:
```bash
gcc -fpreprocessed -E old.rustls.h | sed '/^\s*$/d' | grep -v '#' > old.h
gcc -fpreprocessed -E src/rustls.h | sed '/^\s*$/d' | grep -v '#' > new.h
diff new.h old.h
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,57 @@
 # Changelog
 
+## 0.12.0 (2023-12-03)
+
+This release updates to [Rustls 0.22], but does not yet expose support for
+customizing the cryptographic provider. This will be added in a future release,
+and 0.12.0 continues to use `*ring*` as the only cryptographic provider.
+
+[Rustls 0.22]: https://github.com/rustls/rustls/releases/tag/v%2F0.22.0
+
+### Added
+
+* `RUSTLS_RESULT_CLIENT_CERT_VERIFIER_BUILDER_NO_ROOT_ANCHORS` error code,
+  returned when a client cert verifier is being built that hasn't provided any
+  root trust anchors.
+* The server certificate verifier now supports CRL revocation checking through
+  policy and CRLs provided to the server certificate verifier builder.
+* Client certificate verifier builder now supports controlling CRL revocation
+  status check depth and unknown revocation policy.
+
+### Changed
+
+* The root certificate store constructor (`rustls_root_cert_store_new`) and the
+  function to add PEM content (`rustls_root_cert_store_add_pem`) have been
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,58 @@
...
+  a `rustls_root_cert_store`. The built `rustls_web_pki_server_cert_verifier`
+  can be provided to a client config builder with
+  `rustls_client_config_builder_set_server_verifier`.
+* CRL validation defaults to checking the full certificate chain, and treating
+  unknown revocation status as an error condition.
+
+### Removed
+
+* `RUSTLS_RESULT_CERT_SCT_*` error codes have been removed.
+* `rustls_certificate_get_der` has been removed.
```
cpu, line 54: I'm having trouble finding the commit where this was removed :thinking: I think it might have become redundant with the change to pki_types?

jsha: `git log -S rustls_certificate_get_der` shows me it was 58e2b58af3be910ddcdd49b9aaf11c398aabf314 / #341 (which I squash merged instead of rebase merged 😭 sorry).

cpu: I tried a similar methodology and also landed on that commit, but reviewing the diff I see it adding a lifetime to `rustls_certificate` but left the `rustls_certificate_get_der` fn. In fact, that fn is still present [in main](https://github.com/rustls/rustls-ffi/blob/e474895c39d296ffc1181baa9dcc6ddc5674947a/src/cipher.rs#L51-L68). 

So I think the real mystery is why it got dropped from the `cbindgen` output in `rustls.h` :thinking: :thinking:

cpu: > So I think the real mystery is why it got dropped from the cbindgen output in rustls.h 🤔 🤔

It's not a very satisfying root cause, but I've found that lifting the `rustls_certificate_get_der` out of `impl rustls_certificate` lets `cbindgen` pick up the function again.

I've added a commit to this branch that does that and updated the proposed changelog text to no longer say that `rustls_certificate_get_der` was removed.

jsha (approved):
Thanks!

PR #383, opened 2024-01-17
`cargo fmt` inside of ffi_panic_boundary! invocations

ctz:
This is a one-off, because I don't want to make CI require it (with the knock-on that people have to run random scripts wrapping `cargo fmt` in their workflow).

But even as a one-off I think it's a strict improvement?

The script is:

```shell
sed -i -e 's/ffi_panic_boundary! {/if true {/g' src/*.rs
cargo fmt
sed -i -e 's/if true {/ffi_panic_boundary! {/g' src/*.rs
```


cpu (approved):
Thank you! This is a nice improvement.

Issue #380, opened 2024-01-12
Missing pointer null checks


marco-vassena:
Functions [`set_boxed_mut_ptr`](https://github.com/rustls/rustls-ffi/blob/eeccabdf1ad0518c4a3e4b441665b80a23e03222/src/lib.rs#L490) and [`set_arc_mut_ptr`](set_arc_mut_ptr) assume that the caller pass non-null pointers. It'd be safer to explicitly check that the pointers are not null and return `rustls_result`.

jsha:
One thing that's tricky here: ideally we want all our NULL checks to be at the top of the function, so we can avoid doing any work or allocating anything before potentially bailing out. But `set_boxed_mut_ptr` / `set_arc_mut_ptr` are generally called at the end of the function.

One way to do this would be to keep the requirement that `set_boxed_mut_ptr` / `set_arc_mut_ptr` must not be called with NULL pointers, but additionally enforce that invariant with a panic. For functions that correctly check for NULL at the top, presumably the compiler would be smart enough to eliminate the second check. And for functions that forget to check for NULL, there would be a panic rather than undefined behavior.

Another possibility would be to take `&mut *mut T` instead of `*mut *mut T`, and rely on our existing `try_` macros to produce the reference, which we would then know is non-NULL.

jsha:
Yet another way to do this would be to make inner functions that _return_ T and then have a helper function that (a) checks the out pointer for NULL, (b) calls the inner function, and (c) assigns the result to the target of the out pointer.

ctz:
> Another possibility would be to take `&mut *mut T` instead of `*mut *mut T`, and rely on our existing `try_` macros to produce the reference, which we would then know is non-NULL.

I like this option. We can produce the reference at the top of the function (with the NULL check and early return), and that will eliminate the late failures.

cpu:
@marco-vassena Thanks for the report. We've implemented a fix in #402 that should make this a harder trap to fall into going forward.

PR #389, opened 2024-02-28
Prepare 0.13.0 release with Rustls 0.23

cpu:
Addresses upstream breaking API changes and updates associated Rustls dependencies to use the newest Rustls release.

Most notable for rustls-ffi are the changes in the `rustls_acceptor_accept` and `rustls_accepted_into_connection`
functions where a new `rustls_accepted_alert` out parameter is added.

This type wraps the upstream `AcceptedAlert` type and allows the caller to gather any to-be-written TLS data (most notably, TLS alerts) that was produced while trying to accept the client connection. The caller should handle the to-be-written alert bytes by calling `rustls_accepted_alert_write_tls` with the `rustls_accepted_alert` and a `rustls_write_callback` implementation to write the returned bytes to the connection before calling `rustls_accepted_alert_free`.

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -1,12 +1,14 @@
 use std::sync::Arc;
 
 use libc::{c_void, size_t, EINVAL, EIO};
-use rustls::server::{Accepted, Acceptor};
+use rustls::server::{Accepted, AcceptedAlert, Acceptor};
 use rustls::ServerConfig;
 
 use crate::connection::rustls_connection;
 use crate::error::{map_error, rustls_io_result};
-use crate::io::{rustls_read_callback, CallbackReader, ReadCallback};
+use crate::io::{
+    rustls_read_callback, rustls_write_callback, CallbackReader, CallbackWriter, ReadCallback,
+};
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::server::rustls_server_config;
 use crate::{
@@ -148,7 +150,15 @@ impl rustls_acceptor {
     /// out_accepted: An output parameter. The pointed-to pointer will be set
     ///   to a new rustls_accepted only when the function returns
     ///   RUSTLS_RESULT_OK. The memory is owned by the caller and must eventually
-    ///   be freed.
+    ///   be freed
+    /// out_alert: An output parameter. The pointed-to pointer will be set
+    ///   to a new rustls_accepted_alert only when the function returns
+    ///   a non-OK result. The memory is owned by the caller and must eventually
+    ///   be freed with rustls_accepted_alert_free. The caller should call
+    ///   rustls_accepted_alert_write_tls to write the alert bytes to the TLS
+    ///   connection before freeing the rustls_accepted_alert.
+    ///
+    /// At most one of out_accepted or out_alert will be set.
     ///
     /// Returns:
     ///
@@ -173,19 +183,26 @@ impl rustls_acceptor {
     pub extern "C" fn rustls_acceptor_accept(
         acceptor: *mut rustls_acceptor,
         out_accepted: *mut *mut rustls_accepted,
+        out_alert: *mut *mut rustls_accepted_alert,
     ) -> rustls_result {
         ffi_panic_boundary! {
             let acceptor: &mut Acceptor = try_mut_from_ptr!(acceptor);
             if out_accepted.is_null() {
                 return NullParameter;
             }
+            if out_alert.is_null() {
+                return NullParameter;
+            }
             match acceptor.accept() {
                 Ok(None) => rustls_result::AcceptorNotReady,
-                Err(e) => map_error(e),
                 Ok(Some(accepted)) => {
                     set_boxed_mut_ptr(out_accepted, Some(accepted));
                     rustls_result::Ok
                 }
+                Err((e, accepted_alert)) => {
+                    set_boxed_mut_ptr(out_alert, accepted_alert);
+                    map_error(e)
+                },
             }
         }
     }
@@ -266,7 +283,7 @@ impl rustls_accepted {
             let hello = accepted.client_hello();
             let signature_schemes = hello.signature_schemes();
             match signature_schemes.get(i) {
-                Some(s) => s.get_u16(),
+                Some(s) => u16::from(*s),
                 None => 0,
             }
         }
@@ -304,7 +321,7 @@ impl rustls_accepted {
             let hello = accepted.client_hello();
             let cipher_suites = hello.cipher_suites();
             match cipher_suites.get(i) {
-                Some(cs) => cs.get_u16(),
+                Some(cs) => u16::from(*cs),
                 None => 0,
             }
         }
@@ -365,6 +382,14 @@ impl rustls_accepted {
     /// out_conn: An output parameter. The pointed-to pointer will be set
     ///   to a new rustls_connection only when the function returns
     ///   RUSTLS_RESULT_OK.
+    /// out_alert: An output parameter. The pointed-to pointer will be set
+    ///   to a new rustls_accepted_alert when, and only when, the function returns
+    ///   a non-OK result. The memory is owned by the caller and must eventually
+    ///   be freed with rustls_accepted_alert_free. The caller should call
+    ///   rustls_accepted_alert_write_tls to write the alert bytes to
+    ///   the TLS connection before freeing the rustls_accepted_alert.
+    ///
+    /// At most one of out_conn or out_alert will be set.
     ///
     /// Returns:
     ///
@@ -395,18 +420,25 @@ impl rustls_accepted {
         accepted: *mut rustls_accepted,
         config: *const rustls_server_config,
         out_conn: *mut *mut rustls_connection,
+        out_alert: *mut *mut rustls_accepted_alert,
     ) -> rustls_result {
         ffi_panic_boundary! {
             let accepted: &mut Option<Accepted> = try_mut_from_ptr!(accepted);
             let accepted = try_take!(accepted);
             let config: Arc<ServerConfig> = try_clone_arc!(config);
+            if out_alert.is_null() {
+                return NullParameter;
+            }
             match accepted.into_connection(config) {
                 Ok(built) => {
                     let wrapped = crate::connection::Connection::from_server(built);
                     set_boxed_mut_ptr(out_conn, wrapped);
                     rustls_result::Ok
                 }
-                Err(e) => map_error(e),
+                Err((e, accepted_alert)) => {
+                    set_boxed_mut_ptr(out_alert, accepted_alert);
+                    map_error(e)
+                },
             }
         }
     }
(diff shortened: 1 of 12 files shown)


cpu:
> Investigate Windows cmake build failures

Something seems up with the `cmake` build tasks. They're both failing to link with unresolved external symbol errors:

```
rustls_ffi.lib(std-828e2398afa6ec92.std.37d137a6c0aa880e-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol __imp_WakeByAddressSingle referenced in function _ZN4core3ptr56drop_in_place$LT$std..thread..Packet$LT$$LP$$RP$$GT$$GT$17h27aa9d331296562bE [D:\a\rustls-ffi\rustls-ffi\build\tests\client.vcxproj]
rustls_ffi.lib(std-828e2398afa6ec92.std.37d137a6c0aa880e-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol __imp_WaitOnAddress referenced in function _ZN3std6thread4park17hc58846d9dce664faE [D:\a\rustls-ffi\rustls-ffi\build\tests\client.vcxproj]
```

cpu:
> I'm not sure why aws-lc-rs is still being built as a dependency of rustls with the no-default-features and feature opt-in we've done in this branch,

Ah! :bulb: It's `std`: https://github.com/rustls/rustls/pull/1818

cpu:
> Something seems up with the cmake build tasks.

Unrelated, but on my mind: https://github.com/rustls/rustls-ffi/issues/390 I'm not sure I understand why it makes sense to invest resources into the parallel cmake build infrastructure.

cpu:
>> I'm not sure why aws-lc-rs is still being built as a dependency of rustls with the no-default-features and feature opt-in we've done in this branch,
>
> Ah! 💡 It's std: https://github.com/rustls/rustls/pull/1818

I updated the git patch to pick up this change and was able to drop the CI changes adding `nasm` to the Windows build tasks.

cpu:
>  Finish support for Acceptor error alert handling.

This is related to https://github.com/rustls/rustls/pull/1811 and https://github.com/rustls/rustls/issues/1792 - the previous API swallowed alerts generated during `accept` or `into_connection`. As implemented in this branch, we do the same thing: dropping any alerts on the floor. The upstream Rustls implementation now returns `(Error, AcceptedAlert)`, and the caller must use [AcceptedAlert::write](https://rustls.github.io/rustls/prerelease/server/struct.AcceptedAlert.html#method.write) to potentially write TLS alerts to the wire, allowing for better handling here.

I'm not sure the best way to represent this for the FFI boundary. One idea: the `rustls_acceptor_accept` and `rustls_accepted_into_connection` fn can gain two new out parameters, one `*mut u8` for the data, and one `*mut size_t` for how much alert data was written. The wrapper would handle the error condition by writing from `AcceptedAlert` into the out buffer, and then returning a failure `rustls_io_result`. The caller would be responsible for copying the correct number of bytes from the out buffer to the socket. This seems awkward, but so is basically every C API... :laughing: 

Any other ideas?

jsha:
>  One idea: the rustls_acceptor_accept and rustls_accepted_into_connection fn can gain two new out parameters, one *mut u8 for the data, and one *mut size_t for how much alert data was written. The wrapper would handle the error condition by writing from AcceptedAlert into the out buffer, and then returning a failure rustls_result. The caller would be responsible for copying the correct number of bytes from the out buffer to the socket, but only for the error condition. This seems awkward, but so is basically every C API... 😆

I think an additional out parameter is the correct direction. But it needs to be an owned type with its own lifetime, so the C code can free it when done. If we ask the caller to provide a buffer as you're proposing, we run into problems like figuring out how big a buffer to allocate[1]. Since the Rust side is going to allocate a ChunkVecBuffer anyhow, we might as well use that.

So this looks like

 - expose the `AcceptedAlert` type as `*rustls_accepted_alert`
 - implement an accessor for the bytes inside it
 - implement `rustls_accepted_alert_free()`
 - add a second outparam for the two affected functions, documenting that at most one of the two outparams will be set, and that which one is set is fully determined by whether an error was returned.
 
[1]: But thinking about it a second time: can we put a small upper bound on the size of the alert bytes? If so, that makes the problem of the output buffer less annoying. We could tell people to allocate a small array on the stack and simply write to that, avoiding complications of malloc'ing and so on.

cpu:
> I think an additional out parameter is the correct direction. But it needs to be an owned type with its own lifetime

Aha, that does making arranging things easier :+1: 

> Since the Rust side is going to allocate a ChunkVecBuffer anyhow, we might as well use that.

I ended up introducing a new crate-internal type on our side for this (`AcceptedAlertBuffer`) . The `ChunkVecBuffer` in the upstream `AcceptedAlert` isn't accessible. With the current API in 0.23 you have to call [write](https://docs.rs/rustls/latest/rustls/server/struct.AcceptedAlert.html#method.write) to copy the data out.

> So this looks like ...

Other than the caveat about needing our own `Vec` I think the plan you proposed and what I've implemented in-branch were pretty much in-sync :+1: Thanks!

> [1]: But thinking about it a second time: can we put a small upper bound on the size of the alert bytes? If so, that makes the problem of the output buffer less annoying. 

I thought about this a little bit but the `AcceptedAlert`'s `ChunkVecBuffer` is the `sendable_tls` buffer from the connection's common data. It isn't scoped to alerts specifically, or guaranteed to be a specific number of alerts. It seemed fragile/dangerous to try and assume a safe static buffer size given that so I went with the other design hashed out in the comment thread.

cpu:
I think this is ready for review now. (@ctz maybe you would be interested in giving it a once-over as well?)

ctz (approved):
`rustls_accepted_alert` as a solution looks pretty nice to me.

src/acceptor.rs
```diff
@@ -149,7 +148,15 @@ impl rustls_acceptor {
...
-    ///   be freed.
+    ///   be freed
+    /// out_alert: An output parameter. The pointed-to pointer will be set
+    ///   to a new rustls_accepted_alert only when the function returns
+    ///   a non-OK result. The memory is owned by the caller and must eventually
+    ///   be freed with rustls_accepted_alert_free. The caller should call
+    ///   rustls_accepted_alert_bytes to get the alert bytes, writing them to
+    ///   the TLS connection before freeing the rustls_accepted_alert.
+    ///
+    /// Only one of out_accepted or out_alert will be set.
```
jsha, line 159:
```suggestion
    /// At most one of out_accepted or out_alert will be set.
```

src/acceptor.rs
```diff
@@ -149,7 +148,15 @@ impl rustls_acceptor {
     /// out_accepted: An output parameter. The pointed-to pointer will be set
     ///   to a new rustls_accepted only when the function returns
     ///   RUSTLS_RESULT_OK. The memory is owned by the caller and must eventually
-    ///   be freed.
+    ///   be freed
+    /// out_alert: An output parameter. The pointed-to pointer will be set
+    ///   to a new rustls_accepted_alert only when the function returns
+    ///   a non-OK result. The memory is owned by the caller and must eventually
```
jsha, line 154:
```suggestion
    /// out_alert: An output parameter. The pointed-to pointer will be set
    ///   to a new rustls_accepted_alert when, and only when, the function returns
    ///   a non-OK result. The memory is owned by the caller and must eventually
```

Note: this guarantee that it's always set is complicated by the fact that AcceptedAlert::new can error. It looks like that error is necessitated by the Write interface, but can't actually happen since we're writing to a Vec.

Normally I'd say "assert that there's no error writing to the Vec and panic if that's not true," but panicking isn't an option for us in FFI land. So I'd handle the (can't happen) error case by making an empty AceptedAlert.

One other possibility: the upstream `AcceptedAlert` doesn't actually expose a bytes accessor, it exposes a [`write`](https://docs.rs/rustls/latest/rustls/server/struct.AcceptedAlert.html#method.write) method that expects a `&mut dyn Write`. In rustls-ffi our equivalent of that is a [`rustls_write_callback`](https://docs.rs/rustls-ffi/latest/rustls_ffi/io/type.rustls_write_callback.html) plus a `void *userdata`. And it's likely the user already has that callback handy to write to whatever their underlying I/O is. What do you think of exposing the `write` method on AcceptedAlert instead of the byte accessor?

jsha (changes requested):

src/acceptor.rs
cpu: > What do you think of exposing the write method on AcceptedAlert instead of the byte accessor?

Yeah! That actually works out nicely and lets us drop the awkward `AcceptedAlertBuffer` type I had to invent before.

I've updated the branch, the docs, and the PR description to be arranged around using a `rustls_write_callback` for this job too. Thanks for the suggestion :+1:

ctz (approved):

src/acceptor.rs
```diff
@@ -426,6 +452,66 @@ impl rustls_accepted {
...
+    _private: [u8; 0],
+}
+
+impl Castable for rustls_accepted_alert {
+    type Ownership = OwnershipBox;
+    type RustType = AcceptedAlert;
+}
+
+impl rustls_accepted_alert {
+    /// Write some TLS bytes (e.g. alerts) to the network. The actual network I/O is
```
ctz, line 466: small nit: there should be zero or one alert. if there are more than that, it would be a bug in rustls.
```suggestion
    /// Write some TLS bytes (an alert) to the network. The actual network I/O is
```

ctz (approved):

src/acceptor.rs
cpu: Fixed up. Thanks :+1:

src/acceptor.rs
```diff
@@ -365,6 +379,14 @@ impl rustls_accepted {
...
     ///   to a new rustls_connection only when the function returns
     ///   RUSTLS_RESULT_OK.
+    /// out_alert: An output parameter. The pointed-to pointer will be set
+    ///   to a new rustls_accepted_alert when, and only when, the function returns
+    ///   a non-OK result. The memory is owned by the caller and must eventually
+    ///   be freed with rustls_accepted_alert_free. The caller should call
+    ///   rustls_accepted_alert_write_tls to write the alert bytes to
+    ///   the TLS connection before freeing the rustls_accepted_alert.
+    ///
+    /// Only one of out_conn or out_alert will be set.
```
jsha, line 389: This should also be updated to say "at most one."

cpu: Fixed.

cpu:
@jsha Are there any other changes you'd like to see in this branch?

src/acceptor.rs
```diff
@@ -173,6 +183,7 @@ impl rustls_acceptor {
     pub extern "C" fn rustls_acceptor_accept(
         acceptor: *mut rustls_acceptor,
         out_accepted: *mut *mut rustls_accepted,
+        out_alert: *mut *mut rustls_accepted_alert,
```
jsha, line 186: This needs a NULL check. Alternately we could treat NULL as a signal for "I'm not interested in sending alerts, don't do anything with this parameter" (and put that in the docs).

Same for rustls_accepted_into_connection below.

jsha (changes requested):
Oops, I thought I had approved this. Thanks for the ping. Had one more small piece of feedback; with that addressed, consider this approved by me. Thanks!

src/acceptor.rs
cpu: I added `NULL` checks for the `out_alert` parameters and folded that into the commit that introduced the params.

Afterwards I did a scan of all the other `set_boxed_mut_ptr` callsites and most were missing the same check for their `dst` param. I think prioritizing one of the systematic solutions from https://github.com/rustls/rustls-ffi/issues/380 would be appealing.

> Another possibility would be to take &mut *mut T instead of *mut *mut T, and rely on our existing try_ macros to produce the reference, which we would then know is non-NULL.

I briefly experimented with this but wasn't able to get it working quickly. I'll come back to it, but for the meantime to unblock this landing I added manual null checks throughout in a separate commit: https://github.com/rustls/rustls-ffi/pull/389/commits/7cc74e47c754fa45b62a07601a463e54c7163b00

cpu: > Alternately we could treat NULL as a signal for "I'm not interested in sending alerts, don't do anything with this parameter"

I'm lightly in favour of making the parameter mandatory to signal that handling the alerts is the recommended course of action so I went with the NULL check instead of this approach.

jsha (approved):
Looks good!

Issue #397, opened 2024-03-27
Investigate alternatives to use of nightly features (and other Gentoo build issues)


Kangie:
Gentoo has some requirements for building packages that are causing issues with my attempt to update our package to 0.12.1.

1. The use of nightly features to build a shared library

Gentoo users compile software locally, and our package manager will use the installed Rust to build this package.

As per #345 this fails when using `cbindgen` unless you use a nightly comipler or disable stability guarantees, but maybe we don't need to run `cbindgen` for release builds?

> We vendor the cbindgen [generated rustls.h](https://github.com/rustls/rustls-ffi/blob/main/src/rustls.h) in-repo, which would avoid needing cbindgen or nightly rustc, but I think the cargo-c build process is regenerating it as part of its build. Possibly one fix is figuring out if that's necessary at all.

2. Compilation outside the `src_compile` phase

If we attempt to work around this with `RUSTC_BOOTSTRAP=1`, I see compilation when `cargo cinstall` is run if the `src_test` phase has been called (i.e. if `cargo ctest` has been run between `cargo cbuild` and `cargo cinstall`)

```
 abi_x86_64.amd64: running multilib-minimal_abi_src_install
       Dirty cfg-if v1.0.0: the profile configuration changed
   Compiling cfg-if v1.0.0
       Dirty untrusted v0.9.0: the profile configuration changed
   Compiling untrusted v0.9.0
       Dirty rustls-pki-types v1.3.1: the profile configuration changed
   Compiling rustls-pki-types v1.3.1
       Dirty spin v0.9.8: the profile configuration changed
   Compiling spin v0.9.8
       Dirty log v0.4.21: the profile configuration changed
   Compiling log v0.4.21
       Dirty zeroize v1.7.0: the profile configuration changed
   Compiling zeroize v1.7.0
       Dirty base64 v0.21.5: the profile configuration changed
   Compiling base64 v0.21.5
       Dirty subtle v2.5.0: the profile configuration changed
   Compiling subtle v2.5.0
   ```

We don't allow this in Gentoo - compilation should happen in the appropriate `src_compile` phase.

Current state of the Gentoo Ebuild:

```ebuild
# Copyright 2022-2024 Gentoo Authors
# Distributed under the terms of the GNU General Public License v2

EAPI=8

CRATES="
        aho-corasick@1.1.1
        base64@0.21.5
        cc@1.0.83
        cfg-if@1.0.0
        getrandom@0.2.11
        libc@0.2.153
[...]

PR #398, opened 2024-03-28
Cargo: configure cargo-c to use vendored .h

cpu:
Presently we pre-generate the rustls.h header file using `cbindgen`, commit the result to `src/rustls.h`, and check that the generated result matches the checked-in result in CI. However, the new experimental cargo-c build regenerates its own header file using `cbindgen` by default unless told to do otherwise. We'd prefer it didn't do this because we're using a `cbindgen` feature that requires nightly rust.

This commit updates the `Cargo.toml` capi metadata to tell cargo-c to skip generation of its own header file. We then configure the pre-generated checked-in header file as an asset to be copied into the install include directory.

This better matches how the `Makefile` build allowed building the static lib without needing nightly rust or `cbindgen` and resolves https://github.com/rustls/rustls-ffi/issues/397.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ regex = "1.9.6"
 [package.metadata.capi.header]
 name = "rustls"
 subdirectory = false
+generation = false # Prefer a vendored .h
 
 [package.metadata.capi.library]
 name = "rustls"
@@ -50,3 +51,6 @@ rustflags = "-Cmetadata=rustls-ffi"
 [package.metadata.capi.pkg_config]
 name = "rustls"
 filename = "rustls"
+
+[package.metadata.capi.install.include]
+asset = [{from = "src/rustls.h", to = "" }]


cpu:
@Kangie I'd be interested to hear how this branch works for you. It seems to do the right thing in my local testing :crossed_fingers:

cpu:
@kpcyrd If you wouldn't mind testing I'm hoping this branch lets you drop using `RUSTC_BOOTSTRAP` in your Arch package.

Kangie:
I will try and test with this in the near future, but will be busy over this weekend :(

Pinging @thesamesam as the Gentoo rustls-ffi maintainer who might get a chance to look first!

cpu:
> rustls-ffi / Clippy nightly (optional) (pull_request) Failing after 18s

https://github.com/rustls/rustls-ffi/pull/399

thesamesam:
Thanks! This PR works for us in Gentoo. I'll pull it in.

thesamesam:
> Thanks! This PR works for us in Gentoo. I'll pull it in.

I lied! It looks like `rustls.h` may not be installed when this is applied?

kpcyrd:
I tried again with the new patch and it seems fine for me :+1:

cpu:
> I tried again with the new patch and it seems fine for me 👍

Excellent, thank you!

> (Looking at your output, you may want to update your cargo-c binary) :)

I'm using cargo-c 0.9.24+cargo-0.73.0 but it looks like [nixpkgs unstable updated to 0.9.29](https://github.com/NixOS/nixpkgs/commit/6d5c84fef68657f7187a0de3312db06459c8cb74) since I last rebuilt. I don't think anyone has started packaging .30 or .31 yet. I'll follow up on that.

ctz (approved):

jsha (approved):
Looks good, thanks!

cpu:
I'm going to merge this so I can include it in the 0.13 release. If it turns out there's still an issue with Gentoo builds we can tweak futher. I'll also backport the fix to a 0.12.2 point release.

cpu:
This was included in both [0.12.2](https://github.com/rustls/rustls-ffi/releases/tag/v0.12.2) and [0.13](https://github.com/rustls/rustls-ffi/releases/tag/v0.13.0)

lu-zero:
> Maybe @lu-zero has thoughts on this too? :)

I think the default `to` may be changed since `to=""` probably is more common. I'll think more about it and in case change it and document the default `to`. Thank you for pointing out `to=""` is a bit annoying.

lu-zero:
> > Maybe @lu-zero has thoughts on this too? :)
> 
> I think the default `to` may be changed since `to=""` probably is more common. I'll think more about it and in case change it and document the default `to`. Thank you for pointing out `to=""` is a bit annoying.

I just checked and behaves as intended. I would suggest to keep the include subdir since tends to be more future proof.

``` diff
 [package.metadata.capi.install.include]
-asset = [{from = "src/rustls.h", to = "" }]
+asset = [{from = "src/rustls.h" }]
```

cpu:
@lu-zero Ah! Interesting to know we could omit the `to`. I think I agree with you that we might as well keep it as-is for now. It's working as intended and I doubt we'll have reason to revisit soon.

PR #402, opened 2024-03-29
NULL safe set_boxed_mut_ptr/set_arc_mut_ptr

cpu:
Previously the `set_boxed_mut_ptr()` and `set_arc_mut_ptr()` helper fns used for assigning out parameters across the FFI boundary took `*mut *mut C` and `*mut *const C` for the destination argument `dst`. Using these safely required callers always verify that `dst != NULL`. In practice it's very easy to forget to do this and danger lurks!

We could modify these helpers to do `NULL` checking, but we tend to use them near the end of a function to assign a result in a success case and we would prefer `NULL` checking happen at the beginning of the function.

One proposed solution is to modify these setter functions to take `&mut *mut C` and `&mut *const C`. By using new helper macros to carefully construct a `&mut` from the input double pointer we can front-load the `NULL` check and the assignment in the set fns can proceed knowing there's no possibility for a `NULL` outer pointer.

This commit implements this strategy, updating the argument type of `set_boxed_mut_ptr` and `set_arc_mut_ptr` to take `&mut (*const|*mut) C`. New `try_mut_from_ptr_ptr` and `try_ref_from_ptr_ptr` macros allow converting from `*mut *mut C` and `*mut *const C` to the reference types, bailing early for `NULL`.

Resolves https://github.com/rustls/rustls-ffi/issues/380

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -13,10 +13,9 @@ use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::server::rustls_server_config;
 use crate::{
     ffi_panic_boundary, free_box, rustls_result, set_boxed_mut_ptr, to_box, to_boxed_mut_ptr,
-    try_callback, try_clone_arc, try_mut_from_ptr, try_ref_from_ptr, try_take, Castable,
-    OwnershipBox,
+    try_callback, try_clone_arc, try_mut_from_ptr, try_mut_from_ptr_ptr, try_ref_from_ptr,
+    try_take, Castable, OwnershipBox,
 };
-use rustls_result::NullParameter;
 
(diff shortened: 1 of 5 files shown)


cpu:
I think the fn names and rustdoc could use some bike shedding but I wanted to get this up for input before I log off. 💤

cpu:
@ctz Did you want to give this a review pass since you were interested in https://github.com/rustls/rustls-ffi/issues/380 ?

ctz (approved):
Mental note to pull these into openssl-compat

cpu:
Thanks for taking a look.

> Mental note to pull these into openssl-compat

#404 might be of interest in that area as well.

Issue #151, opened 2021-10-15
Implement macro for opaque struct + CastPtr + BoxCastPtr


jsha:
We have a recurring pattern in this repository:

```
pub struct rustls_server_config_builder {
    // We use the opaque struct pattern to tell C about our types without
    // telling them what's inside.
    // https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
    _private: [u8; 0],
}

impl CastPtr for rustls_server_config_builder {
    type RustType = ServerConfig;
}

impl BoxCastPtr for rustls_server_config_builder {}
```

We should define a macro such that we can simply all this boilerplate to:

    c_type!(rustls_server_config_builder, ServerConfig);

Or, to include BoxCastPtr:

    c_type_boxed!(rustls_server_config_builder, ServerConfig);

We should make sure that doccomments right above the macro invocation wind up in crustls.h associated with the struct declaration.

jsha:
I started work on this. One significant problem is that if we put a doccomment above the macro call, rustdoc treats that as an attempt to document the macro call (which it doesn't support), rather than an attempt to document the first item generated by the macro call. Perhaps we need the macro to accept doccomments as part of its input.

PR #404, opened 2024-03-30
arc_castable!, box_castable!, ref_castable! macros

cpu:
Throughout the project we follow a pattern of implementing safe bridge types for FFI by:

1. Creating a `snake_case_named` [opaque struct](https://github.com/rustls/rustls-ffi/blob/main/CONTRIBUTING.md#opaque-struct-pattern).
2. Implementing `Castable` for the struct:
  a. specifying a `RustType` (some native Rust type).
  b. specifying an `Ownership` (`OwnershipBox`, `OwnershipArc`, `OwnershipRef`).

In the case of `OwnershipRef` we may also need to add a `PhantomData` field referencing a lifetime to the opaque struct.

To reduce the boilerplate of doing all of the above over and over this commit adds three new macros (loosely inspired by the Rustls [`enum_builder!` macro](https://github.com/rustls/rustls/blob/d8d438aecc2f06ddc1e4975822580dd6b2f4c64a/rustls/src/msgs/macros.rs#L2)) that accomplish the task with less ceremony (as suggested in https://github.com/rustls/rustls-ffi/issues/151):  `arc_castable!`, `box_castable!`, and `ref_castable!`.

Care has been taken to avoid the downside [mentioned in 151](https://github.com/rustls/rustls-ffi/issues/151#issuecomment-963751363) about rustdoc comments. The macros ensure the doc comments we add are associated with the correct type and not the macro itself. We can verify this fact by noting that the generated `rustls.h` header file that contains these comments has not changed with the introduction of the macros.

Existing instances of the pattern are updated to use these macros with one exception: the [`rustls_client_hello` type](https://github.com/rustls/rustls-ffi/blob/1dec438490266ccad18db3477555759face1c080/src/server.rs#L422-L448) implements `Castable` with `OwnershipRef` for `rustls_client_hello<'a>`, but it isn't the same opaque struct pattern as used elsewhere and so is best implemented by hand.

I've spot checked the expansion of the macros against my expectations and was happy with the results. If you want to do the same thing, here's [a `cargo expand` of 10c5324](https://gist.github.com/cpu/888f13381ef86f084f447d7133791e69).

Resolves https://github.com/rustls/rustls-ffi/issues/151

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -21,23 +21,18 @@ use webpki::{RevocationCheckDepth, UnknownStatusPolicy};
 use crate::error::{self, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
-    ffi_panic_boundary, free_arc, free_box, set_arc_mut_ptr, set_boxed_mut_ptr, to_arc_const_ptr,
-    to_boxed_mut_ptr, try_clone_arc, try_mut_from_ptr, try_mut_from_ptr_ptr, try_ref_from_ptr,
-    try_ref_from_ptr_ptr, try_slice, try_take, Castable, OwnershipArc, OwnershipBox, OwnershipRef,
+    arc_castable, box_castable, ffi_panic_boundary, free_arc, free_box, ref_castable,
+    set_arc_mut_ptr, set_boxed_mut_ptr, to_arc_const_ptr, to_boxed_mut_ptr, try_clone_arc,
+    try_mut_from_ptr, try_mut_from_ptr_ptr, try_ref_from_ptr, try_ref_from_ptr_ptr, try_slice,
+    try_take,
 };
 use rustls_result::{AlreadyUsed, NullParameter};
 
-/// An X.509 certificate, as used in rustls.
-/// Corresponds to `CertificateDer` in the Rust pki-types API.
-/// <https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.CertificateDer.html>
-pub struct rustls_certificate<'a> {
-    _private: [u8; 0],
-    _marker: PhantomData<&'a ()>,
-}
-
-impl<'a> Castable for rustls_certificate<'a> {
-    type Ownership = OwnershipRef;
-    type RustType = CertificateDer<'a>;
+ref_castable! {
+    /// An X.509 certificate, as used in rustls.
+    /// Corresponds to `CertificateDer` in the Rust pki-types API.
+    /// <https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.CertificateDer.html>
+    pub struct rustls_certificate(CertificateDer<'a>);
 }
 
 /// Get the DER data of the certificate itself.
@@ -62,14 +57,9 @@ pub extern "C" fn rustls_certificate_get_der(
     }
 }
 
-/// A cipher suite supported by rustls.
-pub struct rustls_supported_ciphersuite {
-    _private: [u8; 0],
-}
-
-impl Castable for rustls_supported_ciphersuite {
-    type Ownership = OwnershipRef;
-    type RustType = SupportedCipherSuite;
+ref_castable! {
+    /// A cipher suite supported by rustls.
+    pub struct rustls_supported_ciphersuite(SupportedCipherSuite);
 }
 
 impl rustls_supported_ciphersuite {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -348,6 +352,92 @@ pub(crate) trait Castable {
...
+
+/// Defines a new [`Castable`] opaque struct with [`OwnershipArc`] ownership.
+///
+/// Expects to be invoked with a visibility specifier, the struct keyword, a struct name, and
+/// in parens, the Rust type that the opaque struct wraps (optionally with a lifetime specifier).
+/// Similar to defining a normal struct that wraps a single type as its `.0` member.
+///
+/// If a lifetime is specified, the opaque struct will be parameterized by that lifetime
+/// and a `PhantomData` field referencing the lifetime is added to the struct.
+macro_rules! ref_castable {
```
cpu, line 395: I made an attempt to fold the implementation of `ref_castable!` into `castable!` so the bulk of the macro body could be shared like `box_castable!` and `arc_castable!` but I wasn't happy with the result and so went back to having this macro live on its own. I've only written a handful of simple macros so its possible someone with more experience could refactor this out.

src/lib.rs
```diff
@@ -348,6 +352,92 @@ pub(crate) trait Castable {
...
+        $(#[$comment:meta])*
+        $enum_vis:vis struct $name:ident($rust_type:ty);
+    ) => {
+        crate::castable!(OwnershipArc $(#[$comment])* $enum_vis $name $rust_type);
+    };
+}
+
+pub(crate) use arc_castable;
+
+/// Defines a new [`Castable`] opaque struct with [`OwnershipArc`] ownership.
```
jsha, line 387:
```suggestion
/// Defines a new [`Castable`] opaque struct with [`OwnershipRef`] ownership.
```

src/lib.rs
```diff
@@ -348,6 +352,92 @@ pub(crate) trait Castable {
...
+///
+/// Expects to be invoked with a visibility specifier, the struct keyword, a struct name, and
+/// in parens, the Rust type that the opaque struct wraps (optionally with a lifetime specifier).
+/// Similar to defining a normal struct that wraps a single type as its `.0` member.
+///
+/// If a lifetime is specified, the opaque struct will be parameterized by that lifetime
+/// and a `PhantomData` field referencing the lifetime is added to the struct.
```
jsha, line 394:
```suggestion
/// If a lifetime parameter is specified, the opaque struct will be parameterized it
/// and a `PhantomData` field referencing the lifetime is added to the struct.
```

Nitpicking but I think it's important to be clear that it's a lifetime parameter rather than a specific lifetime (which can't be named).

src/lib.rs
```diff
@@ -348,6 +352,92 @@ pub(crate) trait Castable {
...
+
+/// Defines a new [`Castable`] opaque struct with [`OwnershipArc`] ownership.
+///
+/// Expects to be invoked with a visibility specifier, the struct keyword, a struct name, and
+/// in parens, the Rust type that the opaque struct wraps. Similar to defining a normal struct
+/// that wraps a single type as its `.0` member.
+macro_rules! arc_castable {
+    (
+        $(#[$comment:meta])*
+        $enum_vis:vis struct $name:ident($rust_type:ty);
```
jsha, line 379: Calling this `$enum_vis` is confusing because we're always working with `struct`. But probably we can do away with this variable since it will always be `pub`, match on `pub struct`, and emit `pub struct`.

src/lib.rs
```diff
@@ -348,6 +352,92 @@ pub(crate) trait Castable {
     type RustType;
 }
 
+/// Defines a new [`Castable`] opaque struct with [`OwnershipBox`] ownership.
+///
+/// Expects to be invoked with a visibility specifier, the struct keyword, a struct name, and
+/// in parens, the Rust type that the opaque struct wraps. Similar to defining a normal struct
+/// that wraps a single type as its `.0` member.
```
jsha, line 359: I would say "the Rust type pointed to by raw pointers of the opaque struct's type."

For the second sentence, I would say something like:

> Similar to a [newtype](https://doc.rust-lang.org/book/ch19-04-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction) with `#[repr(transparent)]`, but only allows conversions between Box and raw pointers.

(and similar in other comments with this verbiage)

jsha (commented):
This is a really sweet improvement, thanks! Overall looks great, just some documentation / style nits.

src/cipher.rs
```diff
@@ -943,16 +909,16 @@ impl rustls_web_pki_client_cert_verifier_builder {
...
+box_castable! {
+    /// A server certificate verifier being constructed. A builder can be modified by,
+    /// e.g. `rustls_web_pki_server_cert_verifier_builder_add_crl`. Once you're
+    /// done configuring settings, call `rustls_web_pki_server_cert_verifier_builder_build`
+    /// to turn it into a `rustls_server_cert_verifier`. This object is not safe
+    /// for concurrent mutation.
+    ///
+    /// See <https://docs.rs/rustls/latest/rustls/client/struct.ServerCertVerifierBuilder.html>
+    /// for more information.
+    pub(crate) struct rustls_web_pki_server_cert_verifier_builder(Option<ServerCertVerifierBuilder>);
```
jsha, line 921: I think this wasn't meant to become `pub(crate)` since it's part of hte public API.

src/lib.rs
cpu: Good catch, thanks! Fixed.

src/lib.rs
cpu: That's a good clarification. Applied.

src/lib.rs
cpu: Oops, that was copy-pasta from borrowing from the Rustls enum builder macro.

I've taken your suggestion to remove the parameter and just use `pub` throughout.

src/lib.rs
cpu: Fixed up in each case.

src/cipher.rs
cpu: Good catch! Thank you. Restored to `pub`.

Issue #235, opened 2021-11-15
Fix warnings in server.c / client.c Windows build


jsha:
```
tests/server.c(57): warning C4013: 'ws_strerror' undefined; assuming extern returning int
tests/server.c(57): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 2 has type 'int'
tests/server.c(57): warning C4313: 'fprintf': '%s' in format string conflicts with argument 2 of type 'int'
tests/server.c(62): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 2 has type 'int'
tests/server.c(62): warning C4313: 'fprintf': '%s' in format string conflicts with argument 2 of type 'int'
tests/server.c(98): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 1 has type 'int'
tests/server.c(98): warning C4313: 'fprintf': '%s' in format string conflicts with argument 1 of type 'int'
tests/server.c(110): warning C4477: 'fprintf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type 'size_t'
tests/server.c(110): note: consider using '%zd' in the format string
tests/server.c(129): warning C4477: 'fprintf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type 'size_t'
tests/server.c(129): note: consider using '%zd' in the format string
tests/server.c(135): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 1 has type 'int'
tests/server.c(135): warning C4313: 'fprintf': '%s' in format string conflicts with argument 1 of type 'int'
tests/server.c(168): warning C4477: 'fprintf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type 'size_t'
tests/server.c(168): note: consider using '%zd' in the format string
tests/server.c(331): error C2079: 'siga' uses undefined struct 'sigaction'
tests/server.c(332): error C2224: left of '.sa_handler' must have struct/union type
tests/server.c(333): warning C4013: 'sigaction' undefined; assuming extern returning int
tests/server.c(363): warning C4244: '=': conversion from 'SOCKET' to 'int', possible loss of data
tests/server.c(365): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 1 has type 'int'
[...]

PR #405, opened 2024-03-30
Fix two incompatible types warnings in example code

cpu:
Resolves https://github.com/rustls/rustls-ffi/issues/235

### client.c

The `rustls_verify_server_cert_callback` type is [defined as returning `uint32_t`](https://github.com/rustls/rustls-ffi/blob/1dec438490266ccad18db3477555759face1c080/src/rustls.h#L429-L430). 

In `client.c` we were mistakenly typing the `verify` callback as [returning `enum rustls_result`](https://github.com/rustls/rustls-ffi/blob/1dec438490266ccad18db3477555759face1c080/tests/client.c#L363-L364) when it should be `uint32_t`. This commit makes the required adjustment, silencing the warning.

### server.c

[On Linux](https://linux.die.net/man/2/setsockopt), `man 2 setsockopt` shows the `optval` argument as being `const void *`. [On Windows](
https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt), `setsockopt`'s `optval` argument is `const char*`. This mismatch produces an incompatible types warning when building the `server.c` example on Windows. 

The solution is straight-forward. Explicitly cast `optval` to `const char*`. On Windows this is the correct type. On Linux it will be implicitly converted to `const void *` and everyone is happy.

### DIFF
--- a/tests/client.c
+++ b/tests/client.c
@@ -360,7 +360,7 @@ cleanup:
   return ret;
 }
 
-enum rustls_result
+uint32_t
 verify(void *userdata, const rustls_verify_server_cert_params *params)
 {
   size_t i = 0;
--- a/tests/server.c
+++ b/tests/server.c
@@ -339,7 +339,9 @@ main(int argc, const char **argv)
   }
 
   int enable = 1;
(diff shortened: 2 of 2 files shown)


jsha (approved):
Cheers!

PR #413, opened 2024-04-04
Fix valgrind job

ctz:
`export VALGRIND="valgrind -q"` doesn't work as intended:

- in GH actions, environment alterations don't affect subsequent `run` steps
- in `tests/client_server.rs`, we require `VALGRIND` to be a single command to execute: no shell expansion takes place so no arguments are allowed

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -52,8 +52,7 @@ jobs:
           persist-credentials: false
       - name: Install valgrind
         run: sudo apt-get update && sudo apt-get install -y valgrind
-      - run: export VALGRIND="valgrind -q"
-      - run: make test integration
+      - run: VALGRIND=valgrind make test integration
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -52,8 +52,7 @@ jobs:
           persist-credentials: false
       - name: Install valgrind
         run: sudo apt-get update && sudo apt-get install -y valgrind
-      - run: export VALGRIND="valgrind -q"
-      - run: make test integration
+      - run: VALGRIND=valgrind make test integration
```
cpu, line 55: Good catch :sob: It looks like the `VALGRIND` end manipulation has been inoperable since it was introduced in 456d64bf8f42b76ee21b0f64f895cbf3187b22a5

cpu (approved):

PR #414, opened 2024-04-04
Fix valgrind warning in server.c

ctz:

PR #415, opened 2024-04-04
Makefile: limit scope of format/format-check

cpu:
We don't want `.git` or `target/` or other misc dirs to be subject to these phony targets.

Inspired by [ctz fixing the same](https://github.com/rustls/rustls-openssl-compat/pull/3/files/5c795fedea8801a6b0f6ff8e345bb11857831b13..1ef23d04bbcdf95b40c85ad1aba159a1f9956150#diff-f8c0f426bd38bf5be300f7bfe6b623a7742d7977625ece5ac664d9233ed3b69b) in the compat repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,17 +62,15 @@ clean:
 	rm -rf target
 
 format:
-	find . \
+	find src tests \
 		-name '*.[c|h]' \
-		! -path "./target/*" \
-		! -wholename './src/rustls.h' | \
+		! -wholename 'src/rustls.h' | \
 			xargs clang-format -i
 
 format-check:
-	find . \
+	find src tests \
 		-name '*.[c|h]' \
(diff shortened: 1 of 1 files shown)


ctz (approved):

cpu:
This is manifesting as a spurious format failure [over in this other branch](https://github.com/rustls/rustls-ffi/actions/runs/8557797432/job/23450909797?pr=414)

ctz:
Ah, I think this is now evading the `-wholename` filter

cpu:
> Ah, I think this is now evading the -wholename filter

Fixed up. I also removed the now-unneeded target dir exclusion.

cpu:
Simple test-only fix. Going to merge once CI passes so that #414 can be rebased.

PR #396, opened 2024-03-11
proj: remove Makefile.Windows

cpu:
_Note to reviewers: This builds on #395 - new commits starting at 542c0d69b1fa94c2c1401b28222646025e49a361_

CMake is the recommended way to build `rustls-ffi` for Windows platforms. This commit removes the manually curated Makefile specific to Windows to reduce maintenance burden. See additional discussion in https://github.com/rustls/rustls-ffi/issues/390

PR #419, opened 2024-04-08
 Add `rustls-platform-verifier` binding 

### LINKED ISSUE #417: Bindings to `rustls-platform-verifier`?
Any thoughts on including bindings to [`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier)? AFAICT the reasons listed in https://github.com/rustls/rustls-ffi/pull/110 (which removed `rustls-native-roots`) do not apply to it.

E.g. sth like this (maybe also taking in additional root certs):
```rust
#[no_mangle]
pub extern "C" fn rustls_platform_verifier_server_cert_verifier(
    verifier_out: *mut *mut rustls_server_cert_verifier,
) -> rustls_result {
    ffi_panic_boundary! {
        let verifier_out = try_mut_from_ptr_ptr!(verifier_out);
        let verifier = Arc::new(rustls_platform_verifier::Verifier::new());
        set_boxed_mut_ptr(verifier_out, verifier);
        rustls_result::Ok
    }
}
```

However, it still might be considered out-of-scope for this project, or it might be too new of a project to include here.

amesgen:
Closes #417 

Have not yet taken a closer look at the Windows failures

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.61.0 # MSRV - keep in sync with what rustls considers MSRV
+          - 1.64.0 # MSRV - keep in sync with what rustls considers MSRV
         os: [ubuntu-latest]
         # but only stable on macos/windows (slower platforms)
         include:
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -1161,6 +1161,23 @@ impl Castable for rustls_server_cert_verifier {
 }
 
 impl rustls_server_cert_verifier {
+    /// Create a new server certificate verifier that uses the system's root store and WebPKI via
+    /// `rustls-platform-verifier`.
+    ///
+    /// The verifier can be used in several `rustls_client_config` instances and must be freed by
+    /// the application when no longer needed.
+    #[no_mangle]
+    pub extern "C" fn rustls_platform_server_cert_verifier(
+        verifier_out: *mut *mut rustls_server_cert_verifier,
+    ) -> rustls_result {
+        ffi_panic_boundary! {
+            let verifier_out = try_mut_from_ptr_ptr!(verifier_out);
+            let verifier = Arc::new(rustls_platform_verifier::Verifier::new());
+            set_boxed_mut_ptr(verifier_out, verifier);
+            rustls_result::Ok
+        }
+    }
+
     /// Free a `rustls_server_cert_verifier` previously returned from
     /// `rustls_server_cert_verifier_builder_build`. Calling with NULL is fine. Must not be
     /// called twice with the same value.
--- a/src/rustls.h
+++ b/src/rustls.h
@@ -1251,6 +1251,15 @@ rustls_result rustls_web_pki_server_cert_verifier_builder_build(struct rustls_we
 void rustls_web_pki_server_cert_verifier_builder_free(struct rustls_web_pki_server_cert_verifier_builder *builder);
 
 /**
+ * Create a new server certificate verifier that uses the system's root store and WebPKI via
+ * `rustls-platform-verifier`.
+ *
+ * The verifier can be used in several `rustls_client_config` instances and must be freed by
+ * the application when no longer needed.
+ */
+rustls_result rustls_platform_server_cert_verifier(struct rustls_server_cert_verifier **verifier_out);
+
+/**
  * Free a `rustls_server_cert_verifier` previously returned from
  * `rustls_server_cert_verifier_builder_build`. Calling with NULL is fine. Must not be
  * called twice with the same value.
--- a/tests/client_server.rs
+++ b/tests/client_server.rs
@@ -104,6 +104,12 @@ fn client_server_integration() {
 fn standard_client_tests(valgrind: Option<String>) -> Vec<ClientTest> {
     vec![
         ClientTest {
+            name: "rustls-platform-verifier",
+            valgrind: valgrind.clone(),
+            env: vec![("RUSTLS_PLATFORM_VERIFIER", "1")],
+            expect_error: true,
+        },
+        ClientTest {
             name: "With CA_FILE",
             valgrind: valgrind.clone(),
             env: vec![("CA_FILE", "testdata/minica.pem")],
--- a/tests/static_libs.rs
+++ b/tests/static_libs.rs
@@ -46,7 +46,16 @@ fn expected_linker_parts() -> &'static [&'static str] {
     }
     #[cfg(target_os = "macos")]
     {
-        &["-liconv", "-lSystem", "-lc", "-lm"]
+        &[
+            "-framework",
+            "Security",
+            "-framework",
+            "CoreFoundation",
+            "-liconv",
+            "-lSystem",
+            "-lc",
+            "-lm",
+        ]
     }
     #[cfg(target_os = "windows")]
     {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,10 +12,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f1fdabc7756949593fe60f30ec81974b613357de856987752631dea1e3394c80"
+
+[[package]]
 name = "base64"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9475866fec1451be56a3c2400fd081ff546538961565ccb5b7142cbd22bc7a51"
+
+[[package]]
+name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bytes"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "514de17de45fdb8dc022b1a7975556c53c86f9f0aa5f534b98977b171857c2c9"
 
 [[package]]
 name = "cc"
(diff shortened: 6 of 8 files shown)


src/cipher.rs
```diff
@@ -1161,6 +1161,23 @@ impl Castable for rustls_server_cert_verifier {
...
 
 impl rustls_server_cert_verifier {
+    /// Create a new server certificate verifier that uses the system's root store and WebPKI via
+    /// `rustls-platform-verifier`.
+    ///
+    /// The verifier can be used in several `rustls_client_config` instances and must be freed by
+    /// the application when no longer needed.
+    #[no_mangle]
+    pub extern "C" fn rustls_platform_server_cert_verifier(
+        verifier_out: *mut *mut rustls_server_cert_verifier,
```
amesgen, line 1171: This could also take an (optional) `rustls_root_cert_store` to use for [`Verifier::new_with_extra_roots`](https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots); however, it doesn't exist on e.g. macOS and Windows, so one would either need a platform-conditional API or at least new error codes for these cases. WDYT?

cpu: There's an open issue (https://github.com/rustls/rustls-platform-verifier/issues/58) on rustls-platform-verifier to add support for extra roots for the other platforms.

I think we should keep this simple and avoid implementing it in -ffi until the feature is available more broadly. There's not very much platform-conditional code in this repo and it'd be great to avoid adding more if we can.

.github/workflows/test.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.61.0 # MSRV - keep in sync with what rustls considers MSRV
+          - 1.64.0 # MSRV - keep in sync with what rustls considers MSRV
```
cpu, line 25: I think this comment is a bit inaccurate now, the MSRV of Rustls is still 1.61, it's the MSRV of rustls-platform-verifier we're having to match now.

Perhaps best to reword the comment to say "keep in sync with what rustls and rustls-platform-verifier consider MSRV"?

src/cipher.rs
```diff
@@ -1161,6 +1161,23 @@ impl Castable for rustls_server_cert_verifier {
 }
 
 impl rustls_server_cert_verifier {
+    /// Create a new server certificate verifier that uses the system's root store and WebPKI via
+    /// `rustls-platform-verifier`.
```
cpu, line 1165:
```suggestion
    /// Create a new server certificate verifier that uses the system platform verifier via
    /// [`rustls-platform-verifier`].
    ///
    /// [`rustls-platform-verifier`]: https://github.com/rustls/rustls-platform-verifier
```

I think webpki is only in play for UNIX systems w/o a native platform verifier (e.g. linux, bsd, but not android). 

Since there's some nuance here I think it might be best to just say it uses the platform verifier crate and point to that crate's README for more information.

src/cipher.rs
```diff
@@ -1161,6 +1161,23 @@ impl Castable for rustls_server_cert_verifier {
...
+    ///
+    /// The verifier can be used in several `rustls_client_config` instances and must be freed by
+    /// the application when no longer needed.
```
cpu, line 1168:
```suggestion
    /// The verifier can be used in several `rustls_client_config` instances and must be freed by
    /// the application using `rustls_server_cert_verifier_free` when no longer needed.
```

src/rustls.h
```diff
@@ -1250,6 +1250,15 @@ rustls_result rustls_web_pki_server_cert_verifier_builder_build(struct rustls_we
...
+ * `rustls-platform-verifier`.
+ *
+ * The verifier can be used in several `rustls_client_config` instances and must be freed by
+ * the application when no longer needed.
+ */
+rustls_result rustls_platform_server_cert_verifier(struct rustls_server_cert_verifier **verifier_out);
+
 /**
  * Free a `rustls_server_cert_verifier` previously returned from
  * `rustls_server_cert_verifier_builder_build`. Calling with NULL is fine. Must not be
```
cpu, line 1264: I think this comment could benefit from an update to mention `rustls_platform_server_cert_verifier` as another source of a verifier to be freed.

tests/client_server.rs
```diff
@@ -103,6 +103,12 @@ fn client_server_integration() {
 
 fn standard_client_tests(valgrind: Option<String>) -> Vec<ClientTest> {
     vec![
+        ClientTest {
+            name: "rustls-platform-verifier",
+            valgrind: valgrind.clone(),
+            env: vec![("RUSTLS_PLATFORM_VERIFIER", "1")],
+            expect_error: true,
```
cpu, line 110: it might be nice to think of a way we could test this against a server with a system trusted certificate. Probably the easiest way is an online test (for e.g. against `example.com`). Maybe a separate `Makefile` PHONY target for this? Something like `make connect-tests` ?

tests/static_libs.rs
```diff
@@ -46,7 +46,16 @@ fn expected_linker_parts() -> &'static [&'static str] {
...
+            "-framework",
+            "CoreFoundation",
+            "-liconv",
+            "-lSystem",
+            "-lc",
+            "-lm",
+        ]
     }
     #[cfg(target_os = "windows")]
     {
```
cpu, line 61: I think our `static_libs.rs` test isn't being run on Windows anymore since I removed `Makefile.windows` in https://github.com/rustls/rustls-ffi/pull/396. Previously we ran the test there, and I had updated the CI for the Windows CMake builds to only run the client/server integration test. I forgot to update that when removing the makefile :sob: I will look into fixing this separately.

In an ideal world this test would have failed for this branch, complaining about the expected list missing `crypt32.lib` and pointing to that as the right fix for the Windows CI issue you observe.

cpu (commented):
Thank you! This is looking good. I had a few suggestions for documentation related fixes but the actual implementation is great :stars: 

> Have not yet taken a closer look at the Windows failures

I did some fiddling of my own and I believe we need `crypt32.lib` added to the Windows `CMakeLists.txt` in the `target_link_libraries` list. 

Looking into that was worthwhile because (as mentioned in a review comment) it uncovered that I accidentally removed our test coverage for the expected Windows link libraries changing :sweat: 

Applying this commit https://github.com/rustls/rustls-ffi/commit/575c80dd4e2e5841ea952d71f6dee1ff9c9aeb15 fixes the Windows build in my testing. It tidies the overall list to match what I get using `RUSTFLAGS=--print native-static-libs cargo build` in CI while also adding `crypt32.lib`.

tests/client_server.rs
amesgen: Good idea, will do :+1:

amesgen: Added a new test to `client_server.rs` by slightly enriching the existing `ClientTest`, but let me know if this is undesirable, happy to also add a new Makefile target :+1:

tests/static_libs.rs
amesgen: Thanks for looking into this, I cherry-picked your 575c80dd4e2e5841ea952d71f6dee1ff9c9aeb15 commit for this PR :+1:

tests/client_server.rs
cpu: I think a separate Makefile target would be preferable if you don't mind reworking it. 

It feels separate from the traditional integration tests because we're testing just the client against unknown server software (vs our own server binary). I also think it could flake on occasion from bad internet weather and other factors out of our control. Being able to quickly disable just the `example.com` part could be helpful.

cpu (approved):
Thanks! The latest revisions look great. I left a comment suggesting we separate the new test but that's a minor detail.

cpu:
> Build+test (clang, 1.61.0, ubuntu-latest) Expected — Waiting for status to be reported
> Build+test (gcc, 1.61.0, ubuntu-latest) Expected — Waiting for status to be reported

This is to be expected from the MSRV update. We'll admin merge and fixup the branch protection rules afterwards.

tests/client_server.rs
amesgen: Should it be run on CI?

cpu: My vote would be yes, but as a separate job after the ones running the Rust-based integration tests like:

```
- name: Platform verifier connect test
  run: make connect-test
```

Where the `connect-test` PHONY target will depend on the `client` target, and run the client binary with the right env flags/args to connect to `example.com` with the verifier feature enabled. We'll consider the exit code being a success as sufficient test coverage.

Does that make sense? It's possible I'm overlooking a detail that could make this less appealing than I think.

amesgen: Added a Makefile target which is run in CI. It currently isn't being run in Windows CI (as that uses cmake), maybe that's fine.

cpu: Sounds OK to me. I haven't done my homework on `CMake` to learn what the equivalent setup would look like.

cpu (approved):

src/cipher.rs
```diff
@@ -1161,9 +1161,29 @@ impl Castable for rustls_server_cert_verifier {
...
+    /// [`rustls-platform-verifier`][].
+    ///
+    /// The verifier can be used in several `rustls_client_config` instances and must be freed by
+    /// the application using `rustls_server_cert_verifier_free` when no longer needed.
+    ///
+    /// [`rustls-platform-verifier`]: https://github.com/rustls/rustls-platform-verifier
+    #[no_mangle]
+    pub extern "C" fn rustls_platform_server_cert_verifier(
+        verifier_out: *mut *mut rustls_server_cert_verifier,
+    ) -> rustls_result {
```
jsha, line 1174: Since the only way this function can return an error is on invalid inputs (i.e. `try_mut_from_ptr_ptr!` returns an error), it's better to just return the new pointer: `-> *mut rustls_server_cert_verifier`. From the README:

> ## NULL
>
> The library checks all pointers in arguments for NULL and will return an error
> rather than dereferencing a NULL pointer. For some methods that are infallible
> except for the possibility of NULL (for instance
> `rustls_connection_is_handshaking`), the library returns a convenient
> type (e.g. `bool`) and uses a suitable fallback value if an input is NULL.

jsha (changes requested):
Generally looks great!

src/cipher.rs
amesgen: Good point, done :+1:

jsha (approved):

cpu:
Thanks @amesgen !

cpu:
> This is to be expected from the MSRV update. We'll admin merge and fixup the branch protection rules afterwards.

I've retroactively fixed the branch protection rules for the new MSRV.

Issue #409, opened 2024-03-31
Investigate error returned for an unknown issuer


cpu:
In a downstream application (https://github.com/curl/curl/issues/13248) that wasn't configured with CA certificates connecting to `https://example.com` provokes an odd `BadSignature` error when I would expect something indicating an unknown issuer.

> rustls_connection_process_new_packets: invalid peer certificate: BadSignature

Filing this as a reminder to look into what's going on here.

cpu:
I looked into this and there was a simple explanation (as is usually the case :laughing:). When the pkg-config build configuration was broken in such a way that it failed to populate the default CA path the rustls.c [`cr_init_backend()`](https://github.com/curl/curl/blob/0469c68f4923f6780a04865c61504f7ed3d83b48/lib/vtls/rustls.c#L372-L375) code in curl would find `verifypeer` was true, but `ca_info_blob` and `ssl_cafile` were falsey. This state would skip over [the construction and configuration of the builder with a root cert store and verifier](https://github.com/curl/curl/blob/0469c68f4923f6780a04865c61504f7ed3d83b48/lib/vtls/rustls.c#L412-L478). 

This means when the connection is later built with `rustls_client_config_builder_build()` it uses [the default verifier](https://github.com/rustls/rustls-ffi/blob/cbbef17f068b3fab2a958a9061432d0da94f9056/src/client.rs#L128) from rustls-ffi, [`NoneVerifier`](https://github.com/rustls/rustls-ffi/blob/cbbef17f068b3fab2a958a9061432d0da94f9056/src/client.rs#L66-L106). The `NoneVerifier`'s `verify_server_cert` fn is hardcoded to return:

```rust
        Err(rustls::Error::InvalidCertificate(
            CertificateError::BadSignature,
        ))
```

I think in this case it I think it would be slightly clearer if it returned `CertificateError::UnknownIssuer` so I opened a PR to make that change: https://github.com/rustls/rustls-ffi/pull/421

PR #421, opened 2024-04-14
client: NoneVerifier UnknownIssuer instead of BadSignature

cpu:
The `NoneVerifier` that's used by default if a rustls client config builder is built without a verifier being specified was configured to return `Error::InvalidCertificate(CertificateError::BadSignature)` from all of its trait methods.

This branch updates the `verify_server_cert()` trait method to instead return `Error::InvalidCertificate(CertificateError::UnknownIssuer)`. This will better match what would happen if you configured an empty root certificate store with a real verifier and is perhaps less confusing to debug than an error indicating a cryptographic signature validation error. `BadSignature` is still used for `verify_tls12_signature()` and `verify_tls13_signature()` where it feels like an appropriate default error. In practice neither of these trait methods come into play in `NoneVerifier` use.

Resolves https://github.com/rustls/rustls-ffi/issues/409

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -76,7 +76,7 @@ impl ServerCertVerifier for NoneVerifier {
         _now: UnixTime,
     ) -> Result<ServerCertVerified, rustls::Error> {
         Err(rustls::Error::InvalidCertificate(
-            CertificateError::BadSignature,
+            CertificateError::UnknownIssuer,
         ))
     }
 


cpu:
Alternatively, maybe `rustls_client_config_builder_build` should error if there hasn't been a verifier configured? The `rustls_client_config_builder_new` docs say: 

    /// This starts out with no trusted roots.
    /// Caller must add roots with rustls_client_config_builder_load_roots_from_file
    /// or provide a custom verifier.
    
If the caller *must* set up a verifier, why not error at build-time instead of verification time with the `NoneVerifier`? :thinking:

ctz:
> Alternatively, maybe `rustls_client_config_builder_build` should error if there hasn't been a verifier configured?

I think I prefer this. Though if that option _doesn't_ eliminate `NoneVerifier` (for some reason) we probably should land this PR too?

cpu:
I started looking at the alternative and unfortunately it's a breaking change: `rustls_client_config_builder_build` is infallible right now and would need to be reworked to use an out pointer and return a `rustls_result`. 

I think my inclination is to file an issue for that, tag it as a breaking change, and then hold for the next rustls update.

jsha (approved):
Looks good to me, thanks!

Issue #423, opened 2024-04-19
Use -fsanitize=address for tests only


jsha:
In #33 I added `-fsanitize=address`, "for demo". But looking again at the Makefile it's turned on anytime `CC` is `clang`. We should fix that, since AddressSanitizer isn't intended for production use.

PR #425, opened 2024-04-19
Avoid ASAN for release builds, use w/ GCC or clang in debug builds

cpu:
This branch resolves https://github.com/rustls/rustls-ffi/issues/423 by:

* Making sure we only use ASAN with debug builds, and switching most of CI to run tests with that profile. Valgrind is left using release builds because ASAN + Valgrind conflict without more care.
* Enabling ASAN for GCC in addition to Clang. GCC has also had ASAN support built-in for a long time. We might as well use it for more builds.
* Adding simple checks in CI that debug builds _are_ using ASAN and release builds _aren't_. Presently this isn't comprehensive, but a quick safety check on Linux hosts using `nm`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,8 @@ PROFILE := release
 DESTDIR=/usr/local
 
 ifeq ($(PROFILE), debug)
-ifeq ($(CC), clang)
 	CFLAGS += -fsanitize=address -fsanitize=undefined
-	LDFLAGS += -fsanitize=address
-endif
+	LDFLAGS += -fsanitize=address -fsanitize=undefined
 endif
 
 ifeq ($(PROFILE), release)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,10 +11,12 @@ CFLAGS := -Werror -Wall -Wextra -Wpedantic -g -I src/
 PROFILE := release
 DESTDIR=/usr/local
 
+ifeq ($(PROFILE), debug)
 ifeq ($(CC), clang)
```
cpu: This falls away in the next commit because we drop the `ifeq` conditional on `$(CC)` but jeez was it depressing to learn that GNU Makefiles don't support boolean conditionals and folks suggest madness like [concatenating your conditions](https://stackoverflow.com/a/6452937) :scream:

jsha (approved):

Issue #31, opened 2021-01-28
Consider removing all the type annotations on let bindings


djc:
In places where these aren't needed, the code looks verbose and unidiomatic to me. How would you feel about removing these? In general I'm pretty sure we can trust the type system, and the compiler will point out when we need annotations (that said, I don't have much experience with FFI bindings, so maybe it's different here).

jsha:
Thanks for the suggestion! I definitely agree we can trust the type system. I use abundant type annotations to make comprehension easier for the future reader. For instance, consider this code:

```
        let mut private_keys = match pkcs8_private_keys(&mut Cursor::new(private_key)) {
            Ok(v) => v,
            _ => match rsa_private_keys(&mut Cursor::new(private_key)) {
                Ok(v) => v,
                Err(_) => return rustls_result::PrivateKeyParseError,
            }
        };
        let private_key = match private_keys.pop() {
            Some(p) => PrivateKey(p),
            None => return rustls_result::PrivateKeyParseError,
        };
```

Even though I wrote this last week, I would be hard pressed to tell you whether private_keys is an `Iterator<Item=PrivateKey>`,  `Vec<PrivateKey>`, or (the right answer) a `Vec<Vec<u8>>`. There are some clues that can guide you - for instance, if you look in the second statement, you can see the call to `.pop()`, which indicates `private_keys` is something with a `.pop()` method, probably a `Vec` or `VecDeque`. And on the next line, you can see struct literal `PrivateKey(p)`, which suggests the elements of private_keys are _not_ of type `PrivateKey`. But this requires the reader to do type inference in their head, which I think is quite hard.

Compare to this version:

```rust
        let mut private_keys: Vec<Vec<u8>> = match pkcs8_private_keys(&mut Cursor::new(private_key)) {
            Ok(v) => v,
            _ => match rsa_private_keys(&mut Cursor::new(private_key)) {
                Ok(v) => v,
                Err(_) => return rustls_result::PrivateKeyParseError,
            }
        };
        let private_key: PrivateKey = match private_keys.pop() {
            Some(p) => PrivateKey(p),
            None => return rustls_result::PrivateKeyParseError,
        };
 ```
 
At each `let` statement, it's quite clear to the reader what types they are looking at. This builds confidence and allows them to spend their mental energy elsewhere. I think this reader aid is especially valuable since we may have C programmers who are new-ish to Rust reading this codebase.

djc:
[...]

PR #432, opened 2024-05-30
project-wide tidying, style updates

cpu:
It might be best to review this commit-by-commit. The total overall changeset should have no functional effect - it's exclusively stylistic changes.

* Removes explicit typing throughout, whenever the compiler can successfully infer the types on its own
* Fixes redundant prefixing throughout, whenever an item was already in-scope without the prefix
* Fixes a couple clippy findings that fell out of ^
* Applies a project-wide formatting pass, including within the `ffi_panic_boundary!` macro bodies (using the approach ctz used in https://github.com/rustls/rustls-ffi/pull/383 and that we're using [in rustls-openssl-compat](https://github.com/rustls/rustls-openssl-compat/blob/main/rustls-libssl/admin/format)).
* Updates `make format` and `make format-check` to enforce rust formatting within the `ffi_panic_boundary!` macros

Resolves https://github.com/rustls/rustls-ffi/issues/31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,6 +60,7 @@ thread_local! {
 
 pub(crate) struct Userdata {
     userdata: *mut c_void,
+    #[cfg(not(feature = "no_log_capture"))]
     log_callback: rustls_log_callback,
 }
 
@@ -84,6 +85,7 @@ impl UserdataGuard {
         UserdataGuard {
             data: Some(Userdata {
                 userdata: u,
+                #[cfg(not(feature = "no_log_capture"))]
                 log_callback: None,
             }),
         }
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -458,8 +458,8 @@ pub(crate) use castable;
 /// This can be used regardless of the [`Castable::Ownership`] as we can make const pointers for
 /// `Box`, `Arc` and ref types.
 pub(crate) fn cast_const_ptr<C>(ptr: *const C) -> *const C::RustType
-where
-    C: Castable,
+    where
```
ctz: This commit is also reformatting these where clauses. Intended?

ctz: Actually this reformatting gets undone by the actual formatting commit later, so maybe this could be just dropped?

ctz (approved):

src/lib.rs
cpu: I think this was a mistake along the way. I'll tidy :+1:

cpu: I force pushed a more careful update that avoids any formatting changes until the very end when the project-wide reformatting is applied.

cpu:
@jsha Would you like me to hold merging this until you've had a chance to review?

jsha:
Nope, go ahead and merge, thanks!

PR #434, opened 2024-06-12
`rustls_version()` integration test

cpu:
Figuring out something to do about the manually sync'd `RUSTLS_CRATE_VERSION` constant in `build.rs` has been on my mind for a while.

As explained in the new `build.rs` comment I was originally thinking about populating the `const` automatically, but I decided against that because it seemed like it would require brittle by-hand parsing, or taking a new dep (_unfortunately, it can't be a development dependency if used in `build.rs`_). Both seemed like bad options so I landed on adding an integration test to catch if we forget to update the constant. This has the added advantage of also replacing an existing looser unit test with one that matches the full expected value from `rustls_version()` without adding any new hardcoded constants to maintain.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ capi = []
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "0.23.4", default-features = false, features = [ "ring", "std", "tls12" ]}
+rustls = { version = "0.23.4", default-features = false, features = ["ring", "std", "tls12"] }
 pki-types = { package = "rustls-pki-types", version = "1", features = ["std"] }
-webpki = { package = "rustls-webpki", version = "0.102.0", default-features = false, features = [ "ring", "std" ] }
+webpki = { package = "rustls-webpki", version = "0.102.0", default-features = false, features = ["ring", "std"] }
 libc = "0.2"
 rustls-pemfile = "2"
 log = "0.4.17"
@@ -54,4 +54,4 @@ name = "rustls"
 filename = "rustls"
 
 [package.metadata.capi.install.include]
-asset = [{from = "src/rustls.h", to = "" }]
+asset = [{ from = "src/rustls.h", to = "" }]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ crate-type = ["lib", "staticlib"]
 
 [dev-dependencies]
 regex = "1.9.6"
+toml = { version = "0.6.0", default-features = false, features = ["parse"] }
 
 [package.metadata.capi.header]
 name = "rustls"
--- a/build.rs
+++ b/build.rs
@@ -3,6 +3,11 @@ use std::io::Write;
 use std::{env, fs, path::PathBuf};
 
 // Keep in sync with Cargo.toml.
+//
+// We don't populate this automatically from the Cargo.toml at build time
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -38,6 +38,7 @@ crate-type = ["lib", "staticlib"]
 
 [dev-dependencies]
 regex = "1.9.6"
+toml = { version = "0.6.0", default-features = false, features = ["parse"] }
```
cpu, line 41: note: the version here, and in the `Cargo.lock` was arranged based on this crate's MSRV. The `toml` stack of crates (`toml`, `toml-edit`, `serde_spanned`, etc) have aggressive MSRV reqs for newer versions. (_Another reason I didn't want them to be primary dependencies!_)

cpu:
@ctz If you had a couple min to spare I'd love a +1 on this to get it out of my PR queue. Thanks!

tests/rustls_version.rs
```diff
@@ -0,0 +1,70 @@
...
+///
+/// In particular this ensures that the Rustls version reported in the rustls-ffi version string
+/// matches the version of the Rustls dependency that rustls-ffi was built with.
+///
+/// If this test starts to fail, you probably forgot to update `RUSTLS_CRATE_VERSION` in
+/// `build.rs`.
+#[cfg_attr(miri, ignore)] // Requires file I/O
+#[test]
+fn rustls_version_match() {
+    // Parse Cargo.toml as a generic TOML Table.
```
ctz, line 20: I think this is a bit nicer than what I would have done, which is to shell out to `cargo metadata` and then squish the output through jq!

build.rs
```diff
@@ -3,6 +3,11 @@ use std::io::Write;
 use std::{env, fs, path::PathBuf};
 
 // Keep in sync with Cargo.toml.
+//
+// We don't populate this automatically from the Cargo.toml at build time
```
ctz, line 7: Shower thought:

If the rust crate's build.rs did `println!("cargo:version_number={}", env!("CARGO_PKG_VERSION"));` then we could simply use `env!("DEP_RUSTLS_VERSION_NUMBER")` here.

Haven't tried that, and it's probably a worse overall solution than what we have in this PR, and also I should have learned my lesson by now rather than relying on underused parts of cargo.

ctz (approved):

tests/rustls_version.rs
cpu: I have to re-learn/forget `jq` syntax every time I reach for it!

build.rs
cpu: Ooo, I forgot that `build.rs` can forward information like that. Seems worth playing with but we'd need a rustls release with that magic in-place to be usable here so I think I'll leave it for a "One Day" follow-up.

tests/rustls_version.rs
```diff
@@ -0,0 +1,70 @@
...
+        File::open(PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("Cargo.toml"))
+            .expect("failed to open Cargo.toml");
+    let mut metadata_content = String::new();
+    metadata_file
+        .read_to_string(&mut metadata_content)
+        .expect("failed to read Cargo.toml");
+    let metadata = metadata_content.parse::<Table>().unwrap();
+
+    // Find the crate version specified in Cargo.toml
+    let pacakge_metadata = metadata["package"]
```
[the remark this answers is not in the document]
cpu: Good catch, thanks! Fixed up.

PR #435, opened 2024-06-19
ci: pin cargo-c version in pkg-config workflow

cpu:
The recent 0.10.0 release seems to be causing a regression. Until we've had a chance to dig into the root cause let's pin the version in-use in CI to keep the build passing.

I also snuck in an update to print the pkg-config version in-use. I was interested in that data when I wasn't sure if it was the cargo-c update that changed things or if there was an unrelated system image update that bork'd it.

### DIFF
--- a/.github/workflows/pkg-config.yaml
+++ b/.github/workflows/pkg-config.yaml
@@ -33,10 +33,13 @@ jobs:
         if: runner.os == 'macos-latest'
         run: brew install pkg-config
 
+      - name: Show pkg-config version
+        run: pkg-config --version
+
       - name: Install cargo-c (Ubuntu)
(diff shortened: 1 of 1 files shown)


ctz (approved):

cpu:
I think I [figured out the issue](https://github.com/lu-zero/cargo-c/issues/379#issuecomment-2179229054). I'll revert this pinning + land a fix once I've gotten a better sense of whether the upstream change was intentional or not (and perhaps if there's a way we could make this test rig less brittle).

lu-zero:
Sorry for the noise, try to set `CARGOFLAGS=--libdir=lib` it should return to the original behaviour

PR #440, opened 2024-07-02
build(deps): bump rustls from 0.23.4 to 0.23.5，fix CVE-2024-32650

gordonwwang:
rustls-0.23.5 fix CVE-2024-32650, see https://nvd.nist.gov/vuln/detail/CVE-2024-32650

We need to bump it for security!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,9 +260,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.4"
+version = "0.23.5"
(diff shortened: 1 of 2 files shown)


ctz:
Thanks for the change!

This is unnecessary for two reasons:

- cargo's semver rules (see https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) means `version = "0.23.4"` here already covers 0.23.5 and later versions. Just run `cargo update`
- this crate does not call the affected function

Issue #366, opened 2023-11-21
CryptoProvider support


jsha:
Rustls' upcoming 0.22 release has support for a [`CryptoProvider`](https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/trait.CryptoProvider.html) trait that can be used to provide alternate crypto backends. We should support it in rustls-ffi. We can think of this as two stages:

# Stage 1

Allow choosing between rustls' first party supported providers, `ring`, and `aws-lc-rs`. Requirements:

 - Compile time
   - It must be possible to build with each of: `ring`, `aws-lc-rs`, and "both `ring` and `aws-lc-rs`".
   - If one of the solo options is selected, no code from the other option should be compiled in.
 - Run time
   - It must be possible to select one of `ring` or `aws-lc-rs` for any given config, including selecting different backends for different configs.

# Stage 2

Allow implementation of third-party CryptoProviders through rustls-ffi.

The path to do this would be to expose each of the 15 traits I mentioned in https://github.com/rustls/rustls/issues/1603. The pattern to implement traits in FFI is to effectively do the same thing as trait objects in Rust: define a vtable and do dynamic dispatch.

 - For each method in the trait, define a corresponding C function signature, including a `void *userdata`.
 - Define a struct whose fields are function pointers with each of those signatures. This is the vtable, and it's usually defined statically, though it doesn't have to be.
 - The "trait object" is the combination of a specific `*userdata` with the vtable.

So, there's a path to exposing all the traits of the CryptoProvider API. But I actually think we shouldn't do this. I think the FFI should go the other way. If there's an existing C library that provides crypto implementations (e.g. boringssl), someone should write Rust code that calls out to that library with FFI in order to provide the traits that `CryptoProvider` wants. This is how the demo [`boring-rustls-provider`](https://github.com/janrueth/boring-rustls-provider) does it.

That leaves the question: say a rustls-ffi users wants to use neither `ring` nor `aws-lc-rs`, but some third thing like `boring-rustls-provider`. How do they do that?

I think it should work like this:

 - rustls-ffi accepts a `*rustls_crypto_provider`.
 - The preconditions for that pointer are "it must have come from a Rust `&'static &'static dyn CryptoProvider`" (the two statics handle the DSTs problem we were just discussing).
 - third party `CryptoProvider` implementations can either expose an FFI method themselves to get a pointer of the appropriate type, or someone else can provide it.
 - the third party implementation will be linked in as a separate library.

[...]

Issue #422, opened 2024-04-15
Make `rustls_client_config_builder_build` fallible, rm `NoneVerifier`


cpu:
Followup from https://github.com/rustls/rustls-ffi/pull/421 and https://github.com/rustls/rustls-ffi/issues/409 

The rustls_client_config_builder_new docs say:

```
/// This starts out with no trusted roots.
/// Caller must add roots with rustls_client_config_builder_load_roots_from_file
/// or provide a custom verifier.
```

If the caller **must** set up a verifier, it seems clearer to error at build-time instead of verification time. 

The `NoneVerifier` is only used for the circumstance where we build a client config without the user providing their own verifier with `rustls_client_config_builder_dangerous_set_certificate_verifier` or `rustls_client_config_builder_set_server_verifier`. If the `ClientConfigBuilder` instead changed the `verifier` field from `Arc<dyn ServerCertVerifier>` to `Option<Arc<dyn ServerCertVerifier>>` and `rustls_client_config_builder_build` returned an error for the case where `verifier` was `None`, we could remove `NoneVerifier` and I think bugs that lead to accidentally failing to set a verifier would be easier to catch/debug. I can't think of a use-case where you would want the `NoneVerifier` behaviour (and users that _do_ want it can implement it themselves, providing their impl as the explicit verifier).

This is a breaking change, so we should hold it until we're ready to make a few similar changes at once.

cpu:
This was done in [0.14.0](https://github.com/rustls/rustls-ffi/releases/tag/v0.14.0) and I forgot to close the issue.

PR #441, opened 2024-07-08
0.14.0-rc1: Adopt crypto provider API, use aws-lc-rs as default provider

cpu:
This branch updates `rustls-ffi` to support the cryptography provider model established upsteam in `rustls` since version [0.22](https://github.com/rustls/rustls/releases/tag/v%2F0.22.0), and later refined in [0.23](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0). By the end of the changeset `rustls-ffi`, the unit tests, the integration tests, and CI will support `aws-lc-rs` or `*ring*` as cryptography providers. We follow the precedent set upstream in Rustls 0.23 and make `aws-lc-rs` the default choice. We _do not_ support a model where `rustls-ffi` is built with **both** crypto providers enabled. The `Makefile`/`CMake` build tooling only allows selecting one or the other.

Notably as this branch introduces a number of breaking API changes the version number is updated from 0.13.0 to 0.14.0-rc1. I've labelled this "rc1" while we figure out one remaining downstream issue in HTTPD mod_tls.

Resolves https://github.com/rustls/rustls-ffi/issues/366

## TODO

* [x] update Rustls 0.23.4 -> 0.23.11
* [x] update `Changelog.md` to describe breaking changes (_note: waiting on review feedback in case there are changes_).
* [x] consider trimming CI matrix down?
* [x] attempt a demo using a 3rd party crypto provider
* [x] stage a downstream update in `curl` - [done](https://github.com/curl/curl/compare/master...cpu:curl:cpu-rustls-ffi-0.14/ci)
* [ ] stage a downstream update in `mod-tls` - [wip](https://github.com/apache/httpd/compare/trunk...cpu:httpd:cpu-rustls-ffi-0.14)


cpu:
I recommend reviewing this commit-by-commit. I've taken extra care to make sure each commit builds/tests cleanly along the way.

cpu:
> attempt a demo using a 3rd party crypto provider

I put together a rough proof of concept for this. I don't plan to "productionize" it at this time, but it was sufficient for proving the concept.

[rustls-post-quantum-ffi](https://github.com/cpu/rustls-post-quantum-ffi) is a new crate that I made that wraps up the Rust [`rustls-post-quantum`](https://crates.io/crates/rustls-post-quantum) provider. It exposes one fn, [`const void *rustls_post_quantum_ffi_crypto_provider(void)`](https://github.com/cpu/rustls-post-quantum-ffi/blob/6b62d4af62114566df3b135909de8fb76b0888dd/src/rustls-post-quantum-ffi.h#L9-L18), returning a `void*` to what is an `Arc<CryptoProvider>` under the hood.

I used that to [quickly extend the rustls-ffi `client.c` demo](https://github.com/cpu/rustls-ffi/commit/5fa63c58e4d347c7888b254ab0643e60019e6445) so that it would use `rustls_post_quantum_ffi_crypto_provider()` in place of the AWS LC RS backend.

With that in place we can build `client` and observe it doing post-quantum KEX with Cloudflare's test server:
```
[daniel@blanc:~/Code/Rust/rustls-post-quantum-ffi]$ cargo capi install --prefix=/tmp/rustls-ffi
<snipped>
    Finished `release` profile [optimized] target(s) in 0.08s
  Installing pkg-config file
  Installing header file
  Installing static library
  Installing shared library

[daniel@blanc:~/Code/Rust/rustls-ffi]$ cargo capi install --prefix=/tmp/rustls-ffi
<snipped>
    Finished `release` profile [optimized] target(s) in 18.19s
    Building pkg-config files
  Populating uninstalled header directory
  Installing pkg-config file
  Installing header file
  Installing static library
  Installing shared library

[daniel@blanc:~/Code/Rust/rustls-ffi]$ tree /tmp/rustls-ffi/
/tmp/rustls-ffi/
├── include
│   ├── rustls.h
│   └── rustls-post-quantum-ffi.h
└── lib
    ├── librustls.a
    ├── librustls-post-quantum-ffi.a
    ├── librustls-post-quantum-ffi.so -> librustls-post-quantum-ffi.so.0.1.0
    ├── librustls-post-quantum-ffi.so.0.1.0
    ├── librustls.so -> librustls.so.0.14.0
    ├── librustls.so.0.14.0
    └── pkgconfig
        ├── rustls.pc
        └── rustls-post-quantum-ffi.pc

[daniel@blanc:~/Code/Rust/rustls-ffi]$ PKG_CONFIG_PATH=/tmp/rustls-ffi/lib/pkgconfig make --file=Makefile.pkg-config PROFILE=debug CRYPTO_PROVIDER=aws_lc_rs
<snipped>

[daniel@blanc:~/Code/Rust/rustls-ffi]$ RUSTLS_PLATFORM_VERIFIER=1 LD_LIBRARY_PATH=/tmp/rustls-ffi/lib ./target/client pq.cloudflareresearch.com 443 /cdn-cgi/trace 2>/dev/null | grep "kex"
kex=X25519Kyber768Draft00
kex=X25519Kyber768Draft00
kex=X25519Kyber768Draft00
```

cpu:
@ctz @jsha I'll be out of office next week. Is it possible you folks could take a look at this branch during that time so I can address feedback when I'm back? I think this is one where we want both reviews before merging. Thanks!

cpu:
> stage a downstream update in curl

Here's a [WIP branch](https://github.com/curl/curl/compare/master...cpu:curl:cpu-rustls-ffi-0.14-wip/ci). For the time being I've left all of the `vtls/rustls.c` code using the interfaces that assume a clear default provider is available. That means whoever builds/installs the `librustls.so`/`librustls.a` that curl links choses the crypto provider. There are potentially more complicated options we could pursue where `curl` sets the default provider based on its own configuration, but its not clear if that's the direction the downstream project will want to go or not. Probably merits discussion.

cpu:
> stage a downstream update in `mod-tls`

This one needs a bit more work, but here's [the start of a HTTPD branch that builds mod-tls w/ 0.14.0](https://github.com/apache/httpd/compare/trunk...cpu:httpd:cpu-rustls-ffi-0.14-wip-rebase). Like the `curl` branch for now I've left everything using the default crypto provider.

The experience made me think that it would be nice to offer a simplified certified key construction and supported ciphersuite list construction that assumes the default provider. These were two places where the mod-tls diff required getting a handle on the default explicitly and passing it in. It'd be more convenient for rustls-ffi to do that work for you. I'll push a revision for this when I'm back in office.

> cpu force-pushed the cpu-366-choose-your-own-crypto-adventure branch from d29400b to b07551e

I noticed a `Makefile` omission where building with `CRYPTO_PROVIDER=ring` was causing `aws-lc-rs` to be built as well. The force push was for that fix.

src/connection.rs
```diff
@@ -389,29 +389,37 @@ impl rustls_connection {
...
     pub extern "C" fn rustls_connection_get_negotiated_ciphersuite(
         conn: *const rustls_connection,
-    ) -> *const rustls_supported_ciphersuite {
+    ) -> u16 {
         ffi_panic_boundary! {
-            let conn = try_ref_from_ptr!(conn);
-            let negotiated = match conn.negotiated_cipher_suite() {
+            match try_ref_from_ptr!(conn).negotiated_cipher_suite() {
+                Some(cs) => u16::from(cs.suite()),
+                None => 0,
```
ctz, line 402: 💭 0x0000 is `TLS_NULL_WITH_NULL_NULL` which is a good value to use for this. Another option would be 0xffff (in the reserved range) but that is less nice from C, as it's truthy.

src/connection.rs
```diff
@@ -389,29 +389,37 @@ impl rustls_connection {
         }
     }
 
-    /// Retrieves the cipher suite agreed with the peer.
-    /// This returns NULL until the ciphersuite is agreed.
-    /// The returned pointer lives as long as the program.
+    /// Retrieves the cipher suite identifier agreed with the peer.
```
ctz, line 392: Maybe say that the identifier is the standard IANA numbering and link to https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4 ?

src/crypto_provider.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn build_provider(self) -> CryptoProvider {
+        let cipher_suites = match self.cipher_suites.is_empty() {
+            true => self.base.cipher_suites.clone(),
+            false => self.cipher_suites,
+        };
+
+        // Unfortunately we can't use the `..` syntax to fill in the rest of the provider
+        // fields, because we're working with `Arc<CryptoProvider>` as the base,
+        // not `CryptoProvider`.
+        CryptoProvider {
```
ctz, line 45: Perhaps `..self.base.unwrap_or_clone()`?

src/server.rs
```diff
@@ -57,22 +57,17 @@ arc_castable! {
...
-            // default protocol versions.
-            let base = ServerConfig::builder_with_provider(
-                rustls::crypto::ring::default_provider().into(),
-            )
-            .with_safe_default_protocol_versions()
-            .unwrap();
             let builder = ServerConfigBuilder {
-                base,
+                provider: CryptoProvider::get_default().cloned(),
+                versions: vec![&rustls::version::TLS12, &rustls::version::TLS13],
```
ctz, line 70: `rustls::DEFAULT_VERSIONS.to_vec()` perhaps?

src/client.rs
```diff
@@ -94,26 +94,20 @@ impl ServerCertVerifier for NoneVerifier {
...
-            // default protocol versions.
-            let base = ClientConfig::builder_with_provider(
-                rustls::crypto::ring::default_provider().into(),
-            )
-            .with_safe_default_protocol_versions()
-            .unwrap();
             let builder = ClientConfigBuilder {
-                base,
+                provider: CryptoProvider::get_default().cloned(),
+                versions: vec![&rustls::version::TLS12, &rustls::version::TLS13],
```
ctz, line 118: ditto `DEFAULT_VERSIONS`

src/crypto_provider.rs
```diff
@@ -269,6 +309,24 @@ pub extern "C" fn rustls_supported_ciphersuites_free(
...
+box_castable! {
+    /// A signing key that can be used to construct a certified key.
+    // Note(XXX): we box cast an arc over the dyn trait per the pattern described
+    //   in our docs[0] for dynamically sized types.
+    //   [0]: <https://github.com/rustls/rustls-ffi/blob/main/CONTRIBUTING.md#dynamically-sized-types>
+    pub struct rustls_signing_key(Arc<dyn SigningKey>);
+}
+
+impl rustls_signing_key {
+    /// Frees the `rustls_signing_key`. This is safe to call with `NULL`, or multiple times.
```
ctz, line 335: "multiple times" not sure that is correct? a second call would be a UAF

src/acceptor.rs
```diff
@@ -747,10 +747,11 @@ mod tests {
...
         // Sort to ensure consistent comparison
         signature_schemes.sort();
-        assert_eq!(
-            &signature_schemes,
-            &[1025, 1027, 1281, 1283, 1537, 2052, 2053, 2054, 2055]
-        );
+        #[cfg(feature = "aws_lc_rs")] // aws-lc-rs includes P-521.
+        let expected_schemes = &[1025, 1027, 1281, 1283, 1537, 1539, 2052, 2053, 2054, 2055];
+        #[cfg(all(feature = "ring", not(feature = "aws_lc_rs")))]
+        let expected_schemes = &[1025, 1027, 1281, 1283, 1537, 2052, 2053, 2054, 2055];
```
ctz, line 753: Preexisting, but I'd find 4 character hex (eg `0x0401`) more readable, and thats how iana documents them.

src/crypto_provider.rs
cpu: A good idea, but unfortunately [`Arc::unwrap_or_clone()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.unwrap_or_clone) was only stabilized in 1.76+ and this crate's MSRV is 1.64.

I'll add a TODO for a future cleanup opportunity.

src/connection.rs
cpu: Replaced to use `TLS_NULL_WITH_NULL_NULL` (_which sounds a bit like a terrible dinner order. "Yeah, can I get NULL with a side of NULL, extra NULL and no mayo?"_)

src/connection.rs
cpu: Good idea. Done.

src/client.rs
cpu: Done.

src/server.rs
cpu: Done.

src/crypto_provider.rs
cpu: Oof. Good catch, thank you. Fixed up.

src/acceptor.rs
cpu: I tacked on a separate commit to tidy this up: https://github.com/rustls/rustls-ffi/pull/441/commits/99728c05fff90d337b9814dc0e827619d4092577

I used the `rustls::SignatureScheme` enum instead of raw hex literals. I think that's the best option for readability and keeping everything in-sync with one source of truth.

cpu:
>  consider trimming CI matrix down?

A full run is ~9m. That feels reasonable-ish to me, but it's also probably not necessary to build/test with both clang/gcc for each provider/platform choice. I'm going to leave this as-is until there's additional feedback to consider.

src/crypto_provider.rs
ctz: I think there are a few other instances of this remaining.

cpu: Argh, sorry about that. I should have done a self-review pass. 

I think I've caught the remaining instances.

jsha (changes requested):
Reviewed up through 31aed7a169fdd0efa15f73a6009056062ff04c66. Looking great so far.

src/crypto_provider.rs
```diff
@@ -0,0 +1,273 @@
...
+    builder: *mut rustls_crypto_provider_builder,
+) {
+    ffi_panic_boundary! {
+        free_box(builder);
+    }
+}
+
+/// Return the `rustls_crypto_provider` backed by the `*ring*` cryptography library. The
+/// caller owns the returned `rustls_crypto_provider` and must free it using
+/// g`rustls_crypto_provider_free`.
```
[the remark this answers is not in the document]
jsha: I see in a later commit this was fixed.

src/crypto_provider.rs
```diff
@@ -0,0 +1,273 @@
...
+    try_mut_from_ptr, try_mut_from_ptr_ptr, try_ref_from_ptr, try_ref_from_ptr_ptr, try_slice,
+    try_take,
+};
+
+box_castable! {
+    /// A `rustls_crypto_provider` builder.
+    pub struct rustls_crypto_provider_builder(Option<CryptoProviderBuilder>);
+}
+
+/// A builder for customizing a `CryptoProvider`. Can be used to install a process-wide default.
```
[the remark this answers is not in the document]
jsha: Another way to improve on this might be to rename `_new` to `_new_from_default`. Having the name end in just `_new` makes it seem like it's the default thing to use first, when really they have approximately equal relevance to most people.

jsha:
Finished reviewing. Two last thoughts:

 - I think the test matrix should include a case that builds both ring and aws-lc-rs into the same binary, since that's supported.
 - Ideally the version string should say which crypto providers the package was built against, though that can be a followup.

src/connection.rs
```diff
@@ -389,29 +390,38 @@ impl rustls_connection {
...
-            let negotiated = match conn.negotiated_cipher_suite() {
+            match try_ref_from_ptr!(conn).negotiated_cipher_suite() {
+                Some(cs) => u16::from(cs.suite()),
+                None => u16::from(TLS_NULL_WITH_NULL_NULL),
+            }
+        }
+    }
+
+    /// Retrieves the cipher suite name agreed with the peer.
+    /// This returns "" until the ciphersuite is agreed.
```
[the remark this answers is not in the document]
cpu: Fixed. I cribbed the language already used by `rustls_supported_ciphersuite_get_name`: "The lifetime of the `rustls_str` is the lifetime of the program, it does not need to be freed."

src/crypto_provider.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn build_provider(self) -> CryptoProvider {
+        let cipher_suites = match self.cipher_suites.is_empty() {
+            true => self.base.cipher_suites.clone(),
+            false => self.cipher_suites,
+        };
+
+        // Unfortunately we can't use the `..` syntax to fill in the rest of the provider
+        // fields, because we're working with `Arc<CryptoProvider>` as the base,
+        // not `CryptoProvider`.
+        // TODO(XXX): once MSRV is 1.76+, use `Arc::unwrap_or_clone`.
```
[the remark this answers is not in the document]
cpu: Done: https://github.com/rustls/rustls-ffi/issues/450

src/crypto_provider.rs
```diff
@@ -0,0 +1,273 @@
...
+            secure_random: self.base.secure_random,
+            key_provider: self.base.key_provider,
+        }
+    }
+}
+
+/// Constructs a new `rustls_crypto_provider_builder` using the process-wide default crypto
+/// provider as the base crypto provider to be customized. When this function returns
+/// `rustls_result::Ok` a pointer to the `rustls_crypto_provider_builder` is written to
+/// `boulder_out`. Returns `rustls_result::NoDefaultCryptoProvider` if no default provider
```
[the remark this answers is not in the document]
cpu: Fixed! Still have "boulder" on the mind after all these years I guess! :laughing:

src/crypto_provider.rs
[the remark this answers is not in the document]
cpu: Fixed in the original commit to keep things tidy.

src/client.rs
```diff
@@ -74,26 +74,34 @@ impl ServerCertVerifier for NoneVerifier {
...
         &self,
-        _message: &[u8],
-        _cert: &CertificateDer,
-        _dss: &DigitallySignedStruct,
+        message: &[u8],
+        cert: &CertificateDer,
+        dss: &DigitallySignedStruct,
     ) -> Result<HandshakeSignatureValid, Error> {
-        Err(Error::InvalidCertificate(CertificateError::BadSignature))
+        verify_tls12_signature(
```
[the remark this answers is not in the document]
cpu: > Here you change it to call out to verify_tls12_signature. Why is that?

We now need to carry a `Arc<CryptoProvider>` internal to the `NoneVerifier` in order to implement `supported_verify_schemes()`. After doing that it became possible to implement the signature verification by deferring to the provider (similar to the `tlsclient-mio` example's [`NoCertificateVerification` verifier](https://github.com/rustls/rustls/blob/70d39bcb5a2118066e4fde27c6dce7c335884c4f/examples/src/bin/tlsclient-mio.rs#L342-L393)).

In retrospect it doesn't seem all that useful so I've backed out that part of the diff and left the `NoneVerifier` producing errors for signature verification.

src/crypto_provider.rs
```diff
@@ -219,6 +222,43 @@ pub extern "C" fn rustls_crypto_provider_ciphersuites(
     }
 }
 
+/// Load a private key from the provided PEM content using the crypto provider.
+///
+/// private_key` must point to a buffer of `private_key_len` bytes, containing
```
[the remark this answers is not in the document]
cpu: Fixed. Thanks!

src/crypto_provider.rs
```diff
@@ -219,6 +222,43 @@ pub extern "C" fn rustls_crypto_provider_ciphersuites(
...
+/// a PEM-encoded private key. The exact formats supported will differ based on
+/// the crypto provider in use. The default providers support PKCS#1, PKCS#8 or
+/// SEC1 formats,
```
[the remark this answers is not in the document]
cpu: Fixed.

src/crypto_provider.rs
```diff
@@ -272,6 +312,25 @@ pub extern "C" fn rustls_supported_ciphersuites_free(
     }
 }
 
+box_castable! {
+    /// A signing key that can be used to construct a certified key.
+    // Note(XXX): we box cast an arc over the dyn trait per the pattern described
```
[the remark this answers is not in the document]
cpu: I removed the XXX to leave a normal note comment. 

I'm not sure where I picked this up but I have a habit of using `TODO(XXX)` or `NOTE(XXX)` when I want to leave an IDE-highlighted note/todo but don't intend to do it myself (e.g. `TODO(@cpu)`). I think it's [somewhat common](https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments) but doesn't feel important here.

src/crypto_provider.rs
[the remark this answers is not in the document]
cpu: > So you can't, for instance, for rustls_provider_builder_new, then call rustls_crypto_provider_builder_build_as_default on the result, because either the first call or the second call will error.

That's correct. 

> I think the intent is that you do _new_with_base, followed by _build_as_default to set the process-wide default, and then for subsequent tweaks (e.g. on a per-connection basis), you can do _new. Is that right? 

Yup! You got it. The only extra nuance is that if only one provider is enabled, the default is implicit and nothing needs to explicitly set it up-front. I'll touch on this more in a separate reply to your question about adding CI coverage for builds with both ring/aws-lc-rs enabled (spoiler: I've tried to avoid supporting that for rustls-ffi).

>  If so I'd love to see it explained in a comment for the type, and maybe for each of the functions. 

I've taken a crack at expressing the usual "workflows" for the builders in additions to the comments. PTAL!

>  Another way to improve on this might be to rename _new to _new_from_default

That's a good idea. I implemented this suggested rename.

cpu:
> I think the test matrix should include a case that builds both ring and aws-lc-rs into the same binary, since that's supported.

I should make this more clear in the PR description, but I've tried to _not_ support this use-case. It's technically true that you can build the `rustls-ffi` Rust components manually activating both `--features=ring` and `--features=aws_lc_rs`, but the `Makefile`, `Makefile.pkg-config`, and CMake build configurations only support choosing one or the other (defaulting to `aws_lc_rs` if not specified) and will never activate both features at once. IMO those are the only supported ways to build `rustls-ffi` and so we don't support builds w/ both providers.

I did that on purpose because I couldn't come up with a use-case where you would want _both_ crypto providers built in, and it makes life harder for consuming applications. When both features are enabled the various builder fns that rely on a default will error unless a process-wide default has been explicitly set. If only one of the two features are enabled, the process is implicit and doesn't require any updates in the consumer code. As a concrete example of this the WIP `curl` diff I staged ([diff](https://github.com/curl/curl/compare/master...cpu:curl:cpu-rustls-ffi-0.14-wip/ci)) requires minimal changes under this model. Notably it _doesn't_ need new code to install either `rustls_aws_lc_rs_crypto_provider()` or `rustls_ring_crypto_provider()` as the default.

WDYT? Do you think I should revisit this aspect of the design?

cpu:
>  IMO those are the only supported ways to build rustls-ffi and so we don't support builds w/ both providers.

Maybe this is an oversimplification w.rt. `cargo-c` builds and it is worth having coverage? :thinking:

src/crypto_provider.rs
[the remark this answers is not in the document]
jsha: Ah, good to know. In my headcanon "XXX" means something like the "TKTK" journalists use: change this before submission. I'll interpret it with your meaning when I see it in your code. 👍🏻

jsha:
> we don't support builds w/ both providers.

That's great, I support that. This makes me realize we need an update to the README discussing building with ring. That could be where we specifically document that building with both is unsupported.

jsha (approved):

cpu:
@ctz Can I bug you for a second review? I'm out today but can finish up the changelog and other misc bits on ~Thurs if this looks good to everyone.

Cargo.toml
```diff
@@ -26,7 +26,7 @@ aws_lc_rs = ["rustls/aws_lc_rs", "webpki/aws_lc_rs"]
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "0.23.4", default-features = false, features = ["std", "tls12"] }
+rustls = { version = "0.23.11", default-features = false, features = ["std", "tls12"] }
```
ctz, line 29: Commit message says 0.23.12, this says 0.23.11.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-ffi"
-version = "0.13.0"
+version = "0.14.0"
```
ctz, line 3: Do we have any breaking C API changes? I didn't notice any.

ctz: Sorry, the answer is "yes".

Cargo.toml
cpu: Good catch, thanks! rebase mixup :-) All fixed.

ctz (approved):

cpu:
> This makes me realize we need an update to the README discussing building with ring. That could be where we specifically document that building with both is unsupported.

Added a README update to discuss the crypto provider support, choosing one, and to mention both at once is unsupported.

>  update Changelog.md to describe breaking changes (note: waiting on review feedback in case there are changes).

Added extensive coverage of the added/breaking changes/removed updates to the `CHANGELOG.md`.

> The experience made me think that it would be nice to offer a simplified certified key construction and supported ciphersuite list construction that assumes the default provider.

I reworked the `rustls_signing_key` update so that:

* `rustls_certified_key_build()` retains the `private_key` and `private_key_len` arguments and uses the default crypto provider to create a `rustls_signing_key` to use for the certified key (or will error if not set).
* a new `rustls_certified_key_build_with_signing_key()` fn is added that allows providing a `rustls_signing_key` constructed with a `rustls_crypto_provider` in place of the `private_key` and `private_key_len` arguments of `rustls_certified_key_build()`. The latter uses this function internally.

I also added a new `rustls_default_supported_ciphersuites()` fn to return the `rustls_supported_ciphersuites` of the default crypto provider (or an error if not set).

I think these two changes will make life easier for the downstream users that just want to work with the defaults and not handle a `rustls_crypto_provider` explicitly.

> https://github.com/rustls/rustls-ffi/issues/422  - "Make rustls_client_config_builder_build fallible, rm NoneVerifier"

Lastly, I remembered this issue I had filed as a breaking change and rolled it into this branch. The crypto provider work already made `rustls_client_config_builder_build` fallible so it was straight forward to remove the `NoneVerifier`. A handful of tests needed to be updated to explicitly set a provider to avoid the new `RUSTLS_RESULT_NO_SERVER_CERT_VERIFIER` error returned by the builder operation when no verifier is set.

@jsha Did you want to give any of these updates another pass or is this safe to merge/release? 

@jsha @ctz Thank you both for the reviews! Hopefully this is the last big API update for a while now that we have more parity between this repo and the Rustls 0.21+ crypto provider API.

cpu:
> cpu marked this pull request as draft 6 minutes ago

I think I need a couple more tweaks here (and test coverage) for consumers that want to use the implicit default provider. I'm seeing a failure mode downstream in `curl` where building a server cert verifier fails unexpectedly with this branch. I think the issue is that the verifier is built _before_ a client config. When setting up the rustls-ffi verifier builder we query the process default provider, but it's `None` at this point because no configs have been built to have the Rustls-internal `get_default_or_install_from_crate_features()` do the required magic to install the default implied by unambiguous feature selection. When we try to build the verifier, we find there's no default provider and err.

Working on a fix.

cpu:
> Working on a fix.

The core issue here is that Rustls only [sets the process default based on feature selection](https://github.com/rustls/rustls/blob/d484e412ec2653a79765cb8f716049df56ef801e/rustls/src/crypto/mod.rs#L248-L263) when calling `ClientConfig::builder()`, `ServerConfig::builder()`, `WebPkiClientVerifier::builder()` or `WebPkiServerVerifier::builder()`. 

That makes offering an API like the `rustls_default_supported_ciphersuites()` function in this patch set challenging. We want this function to act like upstream Rustls and work without having to explicitly register a default provider if we can install one on behalf of the user unambiguously based on crate features. However, since `CryptoProvider::get_default_or_install_from_crate_features()` is private we can't. We don't want to error and make the user explicitly install a default before calling it since there's an obvious default to use. We could add a hacky workaround like constructing a throw away `ClientConfig::builder()` and taking advantage of the side-effect it has but that's pretty gross :-/

A similar problem happens w.r.t loading certified keys. If you try to do this with the default provider workflow and haven't built a client or server config builder yet there isn't a process default provider and we can't set one from rustls-ffi based on the implied unambiguous default.

I've worked around the pain this causes in some places by changing where the builders are instantiated in a local WIP branch but I think to offer the best API we should make a case for Rustls exporting `CryptoProvider::get_default_or_install_from_crate_features()`. I'll see about pulling out a pitch for this in the coming days.

I was able to get curl building & passing its rustls test suite with my hacky branch but I think we have the time to get this right before pushing out a 0.14 release with an awkward API update.

cpu:
> I've worked around the pain this causes in some places by changing where the builders are instantiated in a local WIP branch but I think to offer the best API we should make a case for Rustls exporting CryptoProvider::get_default_or_install_from_crate_features(). I'll see about pulling out a pitch for this in the coming days.

It occurs to me as an intermediate step we can reproduce the logic local to rustls-ffi. I'm going to experiment with that.

cpu:
> It occurs to me as an intermediate step we can reproduce the logic local to rustls-ffi. I'm going to experiment with that.

I think this works well and I'm happy with the shape of things again :face_exhaling: 

Summary of changes:

* I moved dropping the `NoneVerifier` earlier in the diff since its less interesting
* We get our own new `rustls-ffi` internal [`get_default_or_install_from_crate_features()` fn](https://github.com/cpu/rustls-ffi/blob/84675823dedad7b9bad98d6ba4cc3d330249b6ac/src/crypto_provider.rs#L410-L447) in `crypto_provider.rs`
* The functions where we were previously using `CryptoProvider::get_default()` internal to various builder constructors are switched to use this function.
* I dropped all of the `ensure_default()` bits in the Rust unit tests - the implied process-wide default is used and we exercise the tests with both `CRYPTO_PROVIDER=aws_lc_rs` and `CRYPTO_PROVIDER=ring`.
* The `client.c` and `server.c` diffs are _much_ smaller and no longer do as much involved `rustls_crypto_provider` work, instead using the process default provider.

I rebased my [wip `curl` branch](https://github.com/cpu/curl/tree/cpu-rustls-ffi-0.14/ci) and its [Rustls test suite](https://github.com/cpu/curl/actions/runs/10291084638/job/28482574978) is now passing. The [actual API update](https://github.com/curl/curl/commit/273b90c51e16e9d1a68b2bc698dab45201f6e00c) is nice and small.

Still working on the Apache HTTPD `mod-tls` update. I'll keep this in draft until that's squared away.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ links = "rustls_ffi"
 rust-version = "1.64"
 
 [features]
+default = ["aws_lc_rs"]
```
ctz, line 15: fwiw I think the rustls featuring naming was a mistake on my part, and it should've been `aws-lc-rs` all along. Not sure if we want to continue with that for consistency, or use hyphens from here on?

README.md
```diff
@@ -22,7 +22,63 @@ to provide the cryptographic primitives.
...
+
+### Choosing a provider
+
+#### Make
+
+When building with the `Makefile`, or example `Makefile.pkg-config` specify
+a `CRYPTO_PROVIDER` as a makefile variable. E.g.:
+
+* `make` to build with the default (`aws-lc-rs`).
+* `make CRYPTO_PROVIDER=aws-lc-rs` to build with `aws-lc-rs` explicitly.
```
ctz, line 44: hm, this disagrees with the makefile (`aws_lc_rs` vs `aws-lc-rs`), ditto other mentions below.

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+requirements.
+
+[Rustls 0.23.12]: https://github.com/rustls/rustls/releases/tag/v%2F0.23.12
+[`aws-lc-rs`]: https://github.com/aws/aws-lc-rs
+
+### Added
+
+* A new `rustls_crypto_provider` type has been added to represent
+  `rustls::CryptoProvider` instances.
+  * The current process-siwde default crypto provider (if any) can be retrieved
```
ctz, line 20:
```suggestion
  * The current process-wide default crypto provider (if any) can be retrieved
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+    with `rustls_crypto_provider_default()`.
+  * If rustls-ffi was built with `aws-lc-rs`, (`DEFINE_AWS_LC_RS` is true), then
+    `rustls_aws_lc_rs_crypto_provider()` can be used to retrive the `aws-lc-rs`
+    provider.
+  * If rustls-ffi was built with `*ring*`, (`DEFINE_RING` is true), then
```
ctz, line 25:
```suggestion
  * If rustls-ffi was built with `ring`, (`DEFINE_RING` is true), then
```

I would avoid the stylised name in code fonts that suggest literal uses.

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+
+* `rustls_server_config_builder_new()`, `rustls_client_config_builder_new()`,
+  `rustls_web_pki_client_cert_verifier_builder_new()`, and
+  `rustls_web_pki_server_cert_verifier_builder_new()`, and
+  `rustls_certified_key_build` functions now use the process
+  default crypto provider instead of being hardcoded to use `*ring*`.
+
+* `rustls_server_config_builder_new_custom()` and
+  `rustls_client_config_builder_new_custom()` no longer take custom
+  ciphersuites as an argument. Instead they requires providing
```
ctz, line 104:
```suggestion
  ciphersuites as an argument. Instead they require providing
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+* Error specificity for revoked certificates was improved.
+
+### Removed
+
+* The `ALL_CIPHER_SUITES` and `DEFAULT_CIPHER_SUITES` constants and associated
+  functions (`rustls_all_ciphersuites_len()`,
+  `rustls_all_ciphersuites_get_entry()`, `rustls_default_ciphersuites_len()` and
+  `rustls_default_ciphersuites_get_entry()`) have been
+  removed. Ciphersuite support is dictated by the `rustls_crypto_provider`. 
+  * Use `rustls_default_supported_ciphersuites()` to retrive
```
ctz, line 135:
```suggestion
  * Use `rustls_default_supported_ciphersuites()` to retrieve
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+   * See the function documentation for more information on recommended
+     workflows.
+
+* A new `rustls_signing_key` type has been added to represent a private key
+  that has been parsed by a `rustls_crypto_provider` and is ready to use for
+  cryptographic operations.
+   * Use `rustls_crypto_provider_load_key()` to load a `signing_key` from
+     a buffer of PEM data using a `rustls_crypto_provider`.
+   * Use `rustls_certified_key_build_with_signing_key()` to build
+     a `rustls_ceritifed_key` with a PEM cert chain and a `rustls_signing_key`.
```
ctz, line 58:
```suggestion
     a `rustls_certified_key` with a PEM cert chain and a `rustls_signing_key`.
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+
+* A new `rustls_crypto_provider` type has been added to represent
+  `rustls::CryptoProvider` instances.
+  * The current process-siwde default crypto provider (if any) can be retrieved
+    with `rustls_crypto_provider_default()`.
+  * If rustls-ffi was built with `aws-lc-rs`, (`DEFINE_AWS_LC_RS` is true), then
+    `rustls_aws_lc_rs_crypto_provider()` can be used to retrive the `aws-lc-rs`
+    provider.
+  * If rustls-ffi was built with `*ring*`, (`DEFINE_RING` is true), then
+    `rustls_ring_crypto_provider()` can be used to retrive the `aws-lc-rs`
```
ctz, line 26:
```suggestion
    `rustls_ring_crypto_provider()` can be used to retrieve the `aws-lc-rs`
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+[`aws-lc-rs`]: https://github.com/aws/aws-lc-rs
+
+### Added
+
+* A new `rustls_crypto_provider` type has been added to represent
+  `rustls::CryptoProvider` instances.
+  * The current process-siwde default crypto provider (if any) can be retrieved
+    with `rustls_crypto_provider_default()`.
+  * If rustls-ffi was built with `aws-lc-rs`, (`DEFINE_AWS_LC_RS` is true), then
+    `rustls_aws_lc_rs_crypto_provider()` can be used to retrive the `aws-lc-rs`
```
ctz, line 23:
```suggestion
    `rustls_aws_lc_rs_crypto_provider()` can be used to retrieve the `aws-lc-rs`
```

jsha:
> @jsha Did you want to give any of these updates another pass or is this safe to merge/release?

Please go ahead without waiting for further reviews from me - thanks!

README.md
cpu: This fell away with the change to `aws-lc-rs`. I've done a pass throughout and I don't think there are any `aws_lc_rs` instances left (except in module paths, e.g. `rustls::crypto::aws_lc_rs`).

CHANGELOG.md
cpu: Thanks! Fixed.

CHANGELOG.md
cpu: Fixed throughout (though I left a historic mention in older release notes).

CHANGELOG.md
cpu: Fixed.

CHANGELOG.md
cpu: The only thing worse than a silly typo is a silly typo copy-pasted a bunch of times :upside_down_face: 

Thanks for catching. Fixed throughout.

CHANGELOG.md
cpu: Fixed.

ctz (approved):

cpu:
> stage a downstream update in mod-tls

I've been working on this more ([wip branch](https://github.com/apache/httpd/compare/trunk...cpu:httpd:cpu-rustls-ffi-0.14-wip-2024-08-27?expand=1)) and have all the `mod_tls` tests passing save one, [`test/modules/tls/test_10_session_id.py::TestSessionID::test_tls_10_session_id_12`](https://github.com/cpu/httpd/blob/trunk/test/modules/tls/test_10_session_id.py). I'm not sure what's up with this one, I wouldn't have expected any changes w.r.t session resumption. I will dig deeper.

I think I also need to revisit some of my `httpd` diff. In my current version I updated [`tls_proto_init`](https://github.com/cpu/httpd/blob/85f0712a92106ba8ea759036e181ad9ec61ae08e/modules/tls/tls_proto.c#L420-L474) to collect up the `apr_hash_set` of `rustls_supported_ciphersuite` instances from the default crypto provider, but I think the lifetime of the `rustls_supported_ciphersuite` pointers are tied to the `rustls_supported_ciphersuites` object. 

In our `server.c`/`client.c` examples this is fine because we peel out a `rustls_supported_ciphersuite` from the `rustls_supported_ciphersuites` and construct a `crypto_provider_builder` and `crypto_provider` with the `rustls_supported_ciphersuite` before calling `rustls_supported_ciphersuites_free` ([relevant helper code](https://github.com/cpu/rustls-ffi/blob/5b44b9f455ee32b85c198871386e8dd4b8272e32/tests/common.c#L389-L459)). In the HTTPD arrangement the ciphersuites are stored in the hashset but not used to construct anything until later-on in `tls_core.c`'s `init_outgoing_connection` and `init_incoming` functions. That means we can't free the `rustls_supported_ciphersuites` at the end of `tls_proto_init` or those pointers are dangling. To unblock progress I added a [very hacky commit](https://github.com/cpu/httpd/commit/85f0712a92106ba8ea759036e181ad9ec61ae08e) to leak the struct since otherwise segfaults ensued :skull: 

I'm not sure the best architectural fix for this wrinkle yet. Perhaps the `tls_proto_conf_t` struct should be changed to hold onto the `rustls_supported_ciphersuites` directly? I think that struct is a singleton that lives as long as the server and isn't freed (?). Alternatively should we think of a way to make this easier in rustls-ffi?

I'm also not sure if either the `test_10_session_id.py` test failure, or the HTTPD supported ciphersuites architectural issue mentioned above should be considered blockers for this release. My intuition right now is that the answer is yes, and that fixing both is the best way to get confidence this release is ready. Divergent opinions welcome.

cpu:
> I'm not sure the best architectural fix for this wrinkle yet. Perhaps the tls_proto_conf_t struct should be changed to hold onto the rustls_supported_ciphersuites directly?

I think this turns out to be a crummy direction to pursue because the `tls_proto_conf_t` struct only holds members allocated from an APR memory pool that (I believe) is freed based on pool-level semantics. It'd be tricky to arrange `rustls_supported_ciphersuites_free()` to be called for a new `rustls_supported_ciphersuites` member.

>  Alternatively should we think of a way to make this easier in rustls-ffi? The ciphersuite impls are 'static so in theory this shouldn't be so painful.

This experience pushed me into realizing the pointers we give to supported ciphersuites need to be associated with the lifetime of the crypto provider they came from (default or otherwise). This also alleviates the need for the `rustls_supported_ciphersuites` type, which was awkwardly trying to have a lifetime separate from the `rustls_crypto_provider` by cloning the `CryptoProvider.cipher_suites` `Vec`.

For the HTTPD use-case where you're working with the default crypto provider the lifetime ends up being equivalent to `'static` because the `Arc<CryptoProvider>` held by Rustls for the process-wide default won't be freed for the lifetime of the process. For the custom provider use-case you'll need to make sure the pointers to supported ciphersuites you get from the provider don't outlast the provider, but that is reasonable (and avoids having to manage the lifetime of a separate `rustls_supported_ciphersuites` collection).

I've reworked this aspect of the changeset and updated the changelog accordingly.

> I've been working on this more ([wip branch](https://github.com/apache/httpd/compare/trunk...cpu:httpd:cpu-rustls-ffi-0.14-wip-2024-08-27?expand=1)) and have all the mod_tls tests passing save one, [test/modules/tls/test_10_session_id.py::TestSessionID::test_tls_10_session_id_12](https://github.com/cpu/httpd/blob/trunk/test/modules/tls/test_10_session_id.py) ([example failure](https://github.com/cpu/httpd/actions/runs/10585065560/job/29330819375)). 

I'm still pretty stumped by this. I was able to verify that the rustls-ffi `server.c` example works as the test expects with `openssl s_client -reconnect -tls_12` seeing only the same `Session-ID` across the 6 negotiations (both with the "default" server config, and with a customized one using `RUSTLS_CIPHERSUITE` to customize down to one supported ciphersuite). This suggests to me the issue isn't with `rustls-ffi` but with the `httpd` diff. However, I'm also seeing the same failure mode with my local `httpd` build using `rustls-ffi` 0.13.0 (???). Very bizarre. I think I'm missing something...

cpu:
> I think I'm missing something...

As expected, I was missing something :laughing: I see the expected output w/ `httpd@trunk` and `rustls-ffi@0.13.0` once I updated my config to include the right mod:

> [Wed Aug 28 21:10:13.939911 2024] [tls:warn] [pid 1385824:tid 1385824] AH10348: session cache [shmcb:mod_tls-sesss(64000)] could not be initialized, will continue without session one. Since this will impact performance, consider making use of the 'TLSSessionCache' directive. The error was: cache type 'shmcb' not supported (known names: ). Maybe you need to load the appropriate socache module (mod_socache_shmcb?).
 
I'm still not sure what's causing the divergence with my WIP branch and 0.14.0 :thinking:

cpu:
> Cargo: version 0.13.0 -> 0.14.0-rc1 

I switched the version number from 0.14.0 to 0.14.0-rc1. Despite a bunch of debugging time I still haven't been able to get to the bottom of the HTTPD `mod_tls` TLS 1.2 resumption issue. Ctz agreed to take a look to see if he can spot what I'm missing. So far my debugging suggests the issue is in my HTTPD patch and not the `rustls-ffi` code, so I think merging this as-is and calling it a release candidate will make getting help easier. If Ctz comes up empty I'll beg/plead/bribe Icing@Apache to take a peek. When we get to the bottom of the problem we can cut a 0.14 with full confidence our two major downstream projects will be able to update w/o issue.

> 39 successful and 11 expected checks

Admin merging so I can fixup the required check names afterwards.

cpu:
> so I can fixup the required check names afterwards.

Done.

PR #443, opened 2024-07-19
docs: clarify freeing client/server builder directly

cpu:
Previously the docs for the `rustls_client_config_builder` and `rustls_server_config_builder` structs seemed to imply the only safe way to dispose of a builder was to build it, and then free the resulting config (e.g. this is a pattern [curl uses for its rustls-ffi integration](https://github.com/curl/curl/blob/1f877b0fba8881e1a7b8a2f1c7e9a1f9ae78831c/lib/vtls/rustls.c#L486-L487) ). It's also possible (and clearer for most error-handling paths) to free the builder directly.

This commit updates the relevant docs to make this alternative option clearer.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -31,9 +31,11 @@ box_castable! {
     /// A client config being constructed. A builder can be modified by,
     /// e.g. rustls_client_config_builder_load_roots_from_file. Once you're
     /// done configuring settings, call rustls_client_config_builder_build
-    /// to turn it into a *rustls_client_config. This object is not safe
-    /// for concurrent mutation. Under the hood, it corresponds to a
-    /// `Box<ClientConfig>`.
+    /// to turn it into a *rustls_client_config. Alternatively, if an error
+    /// occurs or, you don't wish to build a config, call
+    /// `rustls_client_config_builder_free` to free the builder directly.
+    /// This object is not safe for concurrent mutation. Under the hood,
+    /// it corresponds to a `Box<ClientConfig>`.
     /// <https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html>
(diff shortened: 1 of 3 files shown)


cpu:
> rustls-ffi / Clippy nightly (optional) (pull_request) Failing after 19s

I looked at fixing this one but the simple fix (renaming the allow) is mutually exclusive with the non-nightly `clippy` task. For now it seems better to allow the optional nightly job to fail and the required non-nightly to use the old name in its allow.

cpu:
> For now it seems better to allow the optional nightly job to fail and the required non-nightly to use the old name in its allow.

Here's one alternative to consider: https://github.com/rustls/rustls-ffi/pull/444

src/client.rs
```diff
@@ -31,9 +31,11 @@ box_castable! {
     /// A client config being constructed. A builder can be modified by,
     /// e.g. rustls_client_config_builder_load_roots_from_file. Once you're
     /// done configuring settings, call rustls_client_config_builder_build
-    /// to turn it into a *rustls_client_config. This object is not safe
-    /// for concurrent mutation. Under the hood, it corresponds to a
-    /// `Box<ClientConfig>`.
+    /// to turn it into a *rustls_client_config. Alternatively, if an error
```
ctz, line 34: This is following the prevailing style, but I think some paragraphs would be beneficial to split up the text. eg, I would make "Alternatively, if an error..." and "This object is not safe for..." their own paragraphs.

ctz (approved):

src/client.rs
cpu: Good call. I fixed up the instances being touched in this PR.

I think this repo would benefit from a project-wide refresh of the doc strings but it's a bit too much of a lift for me to start this morning.

PR #446, opened 2024-07-30
Remove authors from Cargo.toml

jsha:
Per RFC 3052: https://rust-lang.github.io/rfcs/3052-optional-authors-field.html

Also follows the example of the main rustls repo.


cpu (approved):
LGTM, thanks!

PR #447, opened 2024-07-30
doc: fix comment on WebPKI and platform verifier

jsha:
Since the webpki crate is a specific implementation (which is chosen by the platform verifier crate on some platforms), it was confusing to call it out by name. Instead, simplify the comment to say it uses "the default behavior for the current platform."

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -1055,8 +1055,9 @@ box_castable! {
 }
 
 impl rustls_server_cert_verifier {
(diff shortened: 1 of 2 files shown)


cpu (approved):
Thank you, this change makes sense to me :+1: 

The build was failing because `src/rustls.h` was out-of-date. I took the liberty of pushing a fix for that, squashed into your original commit. Hope that's OK.

ctz (approved):

PR #456, opened 2024-09-05
pkg-config.yml: fix typo

ctz:
The "Install pkg-config (macOS)" step was always skipped (see eg https://github.com/rustls/rustls-ffi/actions/runs/10701958885/job/29669050520) because "runner.os" has possible values "Linux", "Windows", "macOS". I think `matrix.os` was intended here.

(However the job worked anyway, so maybe this step could be removed?)

### DIFF
--- a/.github/workflows/pkg-config.yaml
+++ b/.github/workflows/pkg-config.yaml
@@ -29,10 +29,6 @@ jobs:
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Install pkg-config (macOS)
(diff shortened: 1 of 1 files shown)


cpu (approved):
> (However the job worked anyway, so maybe this step could be removed?)

I think maybe that's a better option (but will approve this fix in case you disagree). I must have overlooked that [macos-latest includes pkg-config](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#utilities) in the base image.

Thanks!

cpu (approved):

PR #458, opened 2024-09-09
crypto_provider: expose a way to get CSRNG data

cpu:
This branch adds a `rustls_crypto_provider_random()` fn for filling a buffer with cryptographically secure random data using a  specific `rustls_crypto_provider`, and `rustls_default_crypto_provider_random()` for the same with the process-wide default. Doing so requires adding a `try_slice_mut!` macro to accompany the existing `try_slice!` macro. It's similar in function but returns a `&mut` slice using `slice::from_raw_parts_mut()` instead of `slice::from_raw_parts()`.

This should help with downstream issues like https://github.com/curl/curl/pull/14770 where `rustls-ffi` stands out as the only TLS `vtls` backend in `curl` that can't provide this functionality.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -679,6 +679,18 @@ macro_rules! try_slice {
 
 pub(crate) use try_slice;
 
+macro_rules! try_slice_mut {
+    ( $ptr:expr, $count:expr ) => {
+        if $ptr.is_null() {
+            return $crate::panic::NullParameterOrDefault::value();
+        } else {
+            unsafe { slice::from_raw_parts_mut($ptr, $count) }
+        }
+    };
+}
+
+pub(crate) use try_slice_mut;
+
 macro_rules! try_callback {
     ( $var:ident ) => {
         match $var {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -672,7 +672,7 @@ macro_rules! try_slice {
         if $ptr.is_null() {
             return $crate::panic::NullParameterOrDefault::value();
         } else {
-            unsafe { slice::from_raw_parts($ptr, $count as usize) }
+            unsafe { slice::from_raw_parts($ptr, $count) }
         }
     };
 }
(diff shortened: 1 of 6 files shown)


src/rustls.h
```diff
@@ -674,26 +675,12 @@ typedef uint32_t (*rustls_session_store_put_callback)(rustls_session_store_userd
                                                       const struct rustls_slice_bytes *key,
                                                       const struct rustls_slice_bytes *val);
 
-/**
- * Rustls' list of supported protocol versions. The length of the array is
- * given by `RUSTLS_ALL_VERSIONS_LEN`.
- */
```
cpu, line 680: Hm, I think I must have dropped these comments by mistake in my 0.14 prep branch and not re-run `cbindgen` pre-merge. Will fix.

cpu: > Hm, I think I must have dropped these comments by mistake in my 0.14 prep branch 

Nevermind, just using the wrong `cbindgen` version on this computer. The comments in `src/enum.rs` are still there.

src/lib.rs
```diff
@@ -679,6 +679,18 @@ macro_rules! try_slice {
 
 pub(crate) use try_slice;
 
+macro_rules! try_slice_mut {
+    ( $ptr:expr, $count:expr ) => {
+        if $ptr.is_null() {
+            return $crate::panic::NullParameterOrDefault::value();
+        } else {
+            unsafe { slice::from_raw_parts_mut($ptr, $count as usize) }
```
ctz, line 687: Can we avoid this `as usize`, to prevent unsafety if someone calls it with (for example) `-1isize`? Let's assume $count is already the right type?

ctz: I guess the same comment also applies to try_slice!

cpu: Done. I also made the change to the original `try_slice!` macro I lifted this from.

ctz (approved):

cpu:
This feels small enough to not merit blocking on a review from Jsha.

PR #461, opened 2024-09-10
cipher: allow getting the proto of a ciphersuite

cpu:
This commit adds a `rustls_supported_ciphersuite_protocol_version()` fn for getting the IANA registered protocol version identifier supported by a given `rustls_supported_ciphersuite`. This avoids downstream users having to use `rustls_supported_ciphersuite_get_name()` and then matching on the protocol version prefix in that identifier (e.g. https://github.com/curl/curl/pull/14840). This is equivalent to the upstream [`SupportedCipherSuite.version()`](https://docs.rs/rustls/latest/rustls/enum.SupportedCipherSuite.html#method.version) fn, but tailored to FFI users.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -19,6 +19,7 @@ use rustls_pemfile::{certs, crls};
 use webpki::{RevocationCheckDepth, UnknownStatusPolicy};
 
 use crate::crypto_provider::{rustls_crypto_provider, rustls_signing_key};
+use crate::enums::rustls_tls_version;
 use crate::error::{self, map_error, rustls_result};
 use crate::rslice::{rustls_slice_bytes, rustls_str};
 use crate::{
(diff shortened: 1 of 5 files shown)


src/cipher.rs
```diff
@@ -100,6 +100,18 @@ pub extern "C" fn rustls_supported_ciphersuite_get_name(
     }
 }
 
+/// Returns the IANA registered protocol version identifier of the ciphersuite.
+///
+/// See also `RUSTLS_ALL_VERSIONS`.
```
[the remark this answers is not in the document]
ctz: I think we have `RUSTLS_TLS_VERSION_TLSV1_2` and co already?

src/cipher.rs
```diff
@@ -100,6 +100,18 @@ pub extern "C" fn rustls_supported_ciphersuite_get_name(
...
+#[no_mangle]
+pub extern "C" fn rustls_supported_ciphersuite_protocol(
+    supported_ciphersuite: *const rustls_supported_ciphersuite,
+) -> u16 {
```
ctz, line 109: Return `rustls_tls_version` perhaps?

ctz (approved):

src/cipher.rs
```diff
@@ -100,6 +100,18 @@ pub extern "C" fn rustls_supported_ciphersuite_get_name(
     }
 }
 
+/// Returns the IANA registered protocol version identifier of the ciphersuite.
+///
+/// See also `RUSTLS_ALL_VERSIONS`.
+#[no_mangle]
+pub extern "C" fn rustls_supported_ciphersuite_protocol(
```
ctz, line 107: I would be minded to have `version` somewhere in the name of this.

src/cipher.rs
cpu: Ah! I forgot we had that. Good catch, thank you

cpu: Switched to use this enum. I also had to add an `Unknown` (0x0000) variant to make things play nice with the `ffi_panic_boundary!` code.

src/cipher.rs
cpu: Renamed to `rustls_supported_ciphersuite_protocol_version()` (my kingdom for namespaces :weary: )

ctz (approved):

cpu:
Feels small enough to land without additional reviews. Happy to revisit if there's further feedback.

PR #465, opened 2024-09-20
client/server: support for KeyLog trait, SSLKEYLOGFILE

cpu:
For debugging purposes it's quite helpful to be able to log session secrets to a file specified by the `SSLKEYLOGFILE` env var, for example to use with Wireshark to decrypt session traffic.

This commit adds two methods to rustls-ffi for both client and server configurations to facilitate this:

1. `rustls_server_config_builder_set_key_log_file()` and `rustls_client_config_builder_set_key_log_file()` enable using the Rustls `KeyLogFile` implementation of the `KeyLog` trait. This option simply honours the `SSLKEYLOGFILE` env var and spits out a NSS formatted key log file appropriate for use with Wireshark and other tools that support this format.

2. `rustls_server_config_builder_set_key_log()` and `rustls_client_config_builder_set_key_log()` enable providing C callbacks that will be invoked to decide which secrets are logged, and to do the logging. This allows for fine-grained control over how secrets are logged and may be more appropriate for applications that already handle this task for other TLS backends ([e.g. curl](https://github.com/curl/curl/blob/master/lib/vtls/keylog.c)).

The client and server examples are updated to optionally use these new features. If the `SSLKEYLOG` env. var is set, both will use the `_set_key_log_file()` fns to set up the standard file based key logging. If the `STDERRKEYLOG` env var is set then both will use the `_set_key_log()` fns to set up custom callbacks that will print the hex-encoded secret data to stderr as a simple demonstration.

See the upstream [`rustls::KeyLog` trait](https://docs.rs/rustls/latest/rustls/trait.KeyLog.html#) and [`rustls::KeyLogFile`](https://docs.rs/rustls/latest/rustls/struct.KeyLogFile.html) implementation for more detail.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,8 +9,8 @@ use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified, Server
 use rustls::client::ResolvesClientCert;
 use rustls::crypto::{verify_tls12_signature, verify_tls13_signature, CryptoProvider};
 use rustls::{
-    sign::CertifiedKey, ClientConfig, ClientConnection, DigitallySignedStruct, Error,
-    ProtocolVersion, SignatureScheme, SupportedProtocolVersion,
+    sign::CertifiedKey, ClientConfig, ClientConnection, DigitallySignedStruct, Error, KeyLog,
+    KeyLogFile, ProtocolVersion, SignatureScheme, SupportedProtocolVersion,
 };
 
 use crate::cipher::{rustls_certified_key, rustls_server_cert_verifier};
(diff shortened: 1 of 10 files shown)


ctz (approved):
LGTM!

cpu:
@jsha Is this a branch you'd like to review pre-merge?

cpu:
> jsha Is this a branch you'd like to review pre-merge?

I'm going to merge this but will handle any feedback separately if it arrives. I'm not planning to cut a release imminently so there's time to address any issues.

PR #478, opened 2024-10-21
WIP: opt-in rustls-ffi FIPS support

cpu:
## FIPS feature

Using `make FIPS=true` with the Makefiles, or `cmake -DFIPS="true" -S . -B build` with the Windows cmake build will activate the `aws-lc-rs` feature of `rustls-ffi`, and the `rustls/fips` feature of Rustls.

On MacOS and Windows this requires some additional build tooling (Golang and Ninja). See [the rustls manual](https://docs.rs/rustls/latest/rustls/manual/_06_fips/index.html) and the [aws-lc-rs-fips-sys crate](https://crates.io/crates/aws-lc-fips-sys) for more information.

## API additions

* Ability to instantiate the FIPS default `crypto_provider` using a new function `rustls_default_fips_provider()`, available only when the fips feature is activated.

* Ability to determine if a given `crypto_provider` is in FIPS mode using a new function `rustls_crypto_provider_fips()`.

* Ability to determine if a given `rustls_client_config` would create connections that are FIPS compatible with a new function
  `rustls_client_config_fips()`. 
  
* Ability to determine if a given `rustls_server_config` would create connections that are FIPS compatible with a new function
  `rustls_server_config_fips()`.

* Ability to determine if a given `rustls_connection` was created from a `rustls_client_config` that was FIPS enabled with a new function `rustls_connection_fips()`. Doing equivalent for a server connection is not presently supported upstream (see https://github.com/rustls/rustls/pull/2174).

## TODO

* [ ] presently the Mac and Windows FIPS-enabled builds fail with unresolved symbol errors when building the client/server examples
* [ ] there's no FIPS specific integration tests.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -613,6 +613,24 @@ impl rustls_connection {
         }
     }
 
+    /// Returns true if the `rustls_connection` was made with a `rustls_client_config`
+    /// that is FIPS compatible.
+    ///
+    /// This is different from `rustls_crypto_provider_fips` which is concerned
+    /// only with cryptography, whereas this also covers TLS-level configuration that NIST
+    /// recommends, as well as ECH HPKE suites if applicable.
+    #[no_mangle]
+    pub extern "C" fn rustls_connection_fips(conn: *const rustls_connection) -> bool {
+        ffi_panic_boundary! {
+            let conn = try_ref_from_ptr!(conn);
+            match &conn.conn {
+                rustls::Connection::Client(c) => c.fips(),
+                // TODO(XXX): investigate why there isn't a ServerConnection::fips().
+                _ => false,
+            }
+        }
+    }
+
     /// Free a rustls_connection. Calling with NULL is fine.
     /// Must not be called twice with the same value.
     #[no_mangle]
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -99,6 +99,32 @@ jobs:
       - name: Integration tests
         run: make PROFILE=debug CERT_COMPRESSION=true integration
 
+  fips:
+    name: FIPS
+    runs-on: "${{ matrix.os }}"
+    continue-on-error: true # TODO(@cpu): remove this after debugging MacOS failures.
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install golang for aws-lc-fips-sys on macos
+        if: runner.os == 'MacOS'
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22.2"
+      - name: Install nightly rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Unit tests
+        run: make FIPS=true test
+      - name: Integration tests
+        run: make FIPS=true integration
+
   test-windows-cmake-debug:
     name: Windows CMake, Debug configuration
     runs-on: windows-latest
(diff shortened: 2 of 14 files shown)


cpu:
> update rustls 0.23.13 -> 0.23.15 

Pulling this out into a separate PR since I have some TODOs for this one: https://github.com/rustls/rustls-ffi/pull/479

cpu:
@justsmth Can I trouble you for your immediate thoughts?

In [MacOS CI](https://github.com/cpu/rustls-ffi/blob/660480daa171ec131bb526023f87ad2057e3da01/.github/workflows/test.yaml#L102-L126) we're using [this makefile](https://github.com/cpu/rustls-ffi/blob/660480daa171ec131bb526023f87ad2057e3da01/Makefile) to build an `.a` archive library with `rustls-ffi`, depending on `rustls` & `aws-lc-rs` (both with `fips` enabled), that is then [statically linked](https://github.com/cpu/rustls-ffi/blob/660480daa171ec131bb526023f87ad2057e3da01/Makefile#L75-L79) to the `client.c` and `server.c` examples from the [`tests/`](https://github.com/rustls/rustls-ffi/blob/main/tests/) directory. It [works OK](https://github.com/rustls/rustls-ffi/actions/runs/11446833819/job/31846870340) on Linux, but there are unresolved `_aws_lc_fips_*` prefixed symbols for the Mac build (similar on Windows but that involves `cmake` and so has a few more moving parts). The [symbol resolution errors](https://github.com/rustls/rustls-ffi/actions/runs/11446833819/job/31846871362) happen _after_ building `rustls-ffi.a`, while building the client/server binaries.

I'm going to revive a Windows desktop env to try and dig deeper on my own but if you have any ideas I'd be grateful!

cpu:
I believe the root cause of the trouble here is that I didn't realize the aws-lc FIPS module only supports static linking **on Linux**. On macOS the module .dylib needs to be used (https://github.com/aws/aws-lc-rs/issues/495). On Windows, there are two `.dll`'s that need to be [distributed with the application](https://aws.github.io/aws-lc-rs/requirements/windows.html#fips-build). 

Both are limitations we can surmount for the CI integration testing of a FIPS-enabled client/server test binary, but I need to think about the best way to pull out a predictable file path for both the `.dylib` and `.dll`'s produced by the `cargo build` of `rustsl-ffi` with the dependent crates in fips-mode. 

Taking Windows as an example, right now they're under paths like:
```
./target/release/build/aws-lc-fips-sys-7dc15b55e6834182/out/build/artifacts/aws_lc_fips_0_12_13_rust_wrapper.dll
./target/release/build/aws-lc-fips-sys-7dc15b55e6834182/out/build/artifacts/aws_lc_fips_0_12_13_crypto.dll
```

I'm not sure the right way to jig a `target_link_libraries` cmake rule for the test binary `CMakeList.txt` that could predict the right paths for the `.dll` artifacts 🤔 

@ctz Do you have any experience with this?

ctz:
> @ctz Do you have any experience with this?

I'm afraid not :(

cpu:
> I'm afraid not :(

Np! I will keep plugging away at this, more because I'm overdue to better understand the existing Windows build than because I think it's super important to get working.

Worst case I'll land Linux in CI and we can circle back on MacOS/Windows testing of the FIPS builds later on.

cpu:

I backed out the CI bits for MacOS and Windows. You can still build `rustls-ffi` in FIPS mode on those platforms, but since linking the `client`/`server` binaries to the FIPS module is more involved on those platforms it's untested in CI at the moment (but verified to work manually).

I've learned _a lot_ (some would say: too much) about `cmake` over the weekend and I think I'd prefer to:

* Land this as-is
* Rework the build system using what I've learned about `cmake` (I have Mac/Linux building w/ `cmake` on a separate branch so we can avoid Windows being the special duck).
* Add back the FIPS coverage for MacOS/Windows after the build rework makes it a bit more tractable.

WDYT? @jsha @ctz

src/connection.rs
```diff
@@ -613,6 +613,24 @@ impl rustls_connection {
...
+    /// This is different from `rustls_crypto_provider_fips` which is concerned
+    /// only with cryptography, whereas this also covers TLS-level configuration that NIST
+    /// recommends, as well as ECH HPKE suites if applicable.
+    #[no_mangle]
+    pub extern "C" fn rustls_connection_fips(conn: *const rustls_connection) -> bool {
+        ffi_panic_boundary! {
+            let conn = try_ref_from_ptr!(conn);
+            match &conn.conn {
+                rustls::Connection::Client(c) => c.fips(),
+                // TODO(XXX): investigate why there isn't a ServerConnection::fips().
```
ctz, line 628: now we did #481, this should be available so we can drop this TODO and its mention in the commit message

cpu: Thanks, forgot to circle back on that one. All fixed.

ctz (approved):

cpu:
@jsha Is this a branch I should hold for your review? :bow:

PR #484, opened 2024-11-06
ci: cancel in-progress GH runs for same PR || ref

cpu:
If you're pushing new work it's generally not useful to have CI churning on the old work. Cancel it and get cooking on the new stuff ASAP. See [github docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-concurrency-groups) for more information. 

h/t https://github.com/rustls/rustls/pull/2197 & @ctz for the config.


cpu:
Let's give this more time to bake in the main repo first.

PR #485, opened 2024-11-06
client-side encrypted client hello (ECH) support

cpu:
## Client ECH support

This branch adds initial support for customizing a `rustls_client_config_builder` instance to use encrypted client hello
(ECH). Presently this is only supported with the `aws-lc-rs` crypto provider.

In sum it adds:

* New `rustls_hpke` type for representing a collection of HPKE suites as   an opaque struct.
* New `rustls_supported_hpke()` function for returning a const pointer to a `rustls_hpke` if available. With the aws-lc-rs backend this is a wrapper around `rustls::crypto::aws_lc_rs::hpke::ALL_SUPPORTED_SUITES`. For other providers it returns NULL.
* New `rustls_client_config_builder_enable_ech_grease` client config builder function for configuring ECH GREASE using a `rustls_hpke`. This chooses the HPKE suite at random, and so brings in a dep on the `rand` crate. We could come up with a simpler selection mechanism if avoiding the dependency is preferred.
* New `rustls_client_config_builder_enable_ech` client config builder function for configuring ECH using a `rustls_hpke` and a TLS encoded ECH config.

Additionally the `client.c` example is updated to add `RUSTLS_ECH_GREASE` and `RUSTLS_ECH_CONFIG` env vars. The former when non-zero configures ECH GREASE if possible, the latter configures ECH with the TLS encoded config list read from the provided file path. 

To facilitate a reliable cross-platform way to fetch ECH configs a small `ech-fetch` test program is added that accepts a domain and (optionally) a file name and writes the TLS encoded ECH config for that domain to the file after performing the appropriate DNS-over-HTTPS lookup. This requires a new dev-only dep on HickoryDNS and Tokio.

## Testing

Test locally with:

```
make
cargo test --test ech_fetch research.cloudflare.com /tmp/research.cloudflare.com.ech.configs.bin
RUSTLS_ECH_CONFIG_LIST=/tmp/research.cloudflare.com.ech.configs.bin \
  RUSTLS_PLATFORM_VERIFIER=1 \
  ./target/client research.cloudflare.com 443 /cdn-cgi/trace 2>/dev/null \
    | grep "sni="
```

You should see output that includes:
```
sni=encrypted
```

You may have to run this a few times to avoid an error of the form: "content length header not found". The reason for this is that the existing `client.c` HTTP handling is somewhat primitive and can't handle chunked responses like this server generates. 

The existing client code only successfully handles an HTTP response if either:
1. We get the full plaintext and read a connection close from the server in the first iteration of `send_request_and_read_response`. In this case we dump the plaintext to stdout without processing any HTTP headers and so it "works" with chunked encoding responses (mostly by accident).
2. We buffer plaintext until we've found an HTTP body, and then find the content header in the header data, using that to determine how to proceed. This fails for the two ECH servers I typically test with (`defo.ie` and `research.cloudflare.com`) because both use chunked encoding and so never send a content length header.

This buggy handling can be replicated without ECH under the right conditions, but _with_ ECH, it's much more frequent that we miss the happy accident case (number 1 above.).

Because of this I've not added CI coverage of running the test client. I think it makes more sense to land this as-is and break out a separate refactoring of the client.c HTTP handling. The code is a little bit messy and could use a more thoughtful rework (maybe as a state machine?).

## TODO

* [x] Figure out a solution for Rustls HPKE suites with MSRV 1.71 - presently using 1.71 or 1.72 on this branch produces an [internal compiler error](https://gist.github.com/cpu/35dc574853a2ad185ba00ded256b20e2). See https://github.com/rust-lang/rust/issues/133578
* [x] ~~Add CI coverage for client.c test w/ ECH - blocked on making `client.c` reliable in more circumstances.~~ Tracking in https://github.com/rustls/rustls-ffi/issues/498

### DIFF
--- a/src/rustls.h
+++ b/src/rustls.h
@@ -62,6 +62,8 @@ enum rustls_result {
   RUSTLS_RESULT_NO_DEFAULT_CRYPTO_PROVIDER = 7016,
   RUSTLS_RESULT_GET_RANDOM_FAILED = 7017,
   RUSTLS_RESULT_NO_CERT_RESOLVER = 7018,
+  RUSTLS_RESULT_HPKE_ERROR = 7019,
+  RUSTLS_RESULT_BUILDER_INCOMPATIBLE_TLS_VERSIONS = 7020,
   RUSTLS_RESULT_NO_CERTIFICATES_PRESENTED = 7101,
   RUSTLS_RESULT_DECRYPT_ERROR = 7102,
   RUSTLS_RESULT_FAILED_TO_GET_CURRENT_TIME = 7103,
@@ -281,6 +283,15 @@ typedef struct rustls_crypto_provider rustls_crypto_provider;
  * A `rustls_crypto_provider` builder.
  */
 typedef struct rustls_crypto_provider_builder rustls_crypto_provider_builder;
+
+/**
+ * A collection of supported Hybrid Public Key Encryption (HPKE) suites.
+ *
+ * `rustls_hpke` can be provided to `rustls_client_config_builder_enable_ech` and
+ * `rustls_client_config_builder_enable_ech_grease()` to customize a
+ * `rustls_client_config_builder` to use Encrypted Client Hello (ECH).
+ */
+typedef struct rustls_hpke rustls_hpke;
 
 /**
  * An alias for `struct iovec` from uio.h (on Unix) or `WSABUF` on Windows.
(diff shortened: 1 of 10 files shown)


cpu:
> cpu force-pushed the cpu-client-ech branch from f5b09bd to 5ee21ff 

just rebasing, no substantive iteration yet.

cpu:
> The MSRV tests may have shaken out now?

Yup!

> The other option is to hit cloudflare's DNS API? 

I gave this more thought and realized that it's beneficial to do as much lifting as possible in Rust code because it's cross-platform and avoids me having to fight through Powershell on Windows (:nauseated_face:). I think the Hickory DNS helper util is probably the best option with that in mind.

> I would be minded to minimise this API as much as possible. I haven't fully thought this through, but how about:
> ...

This was helpful input. I implemented the API you suggested and it was much simpler. I collapsed everything into one commit as a result. PTAL.

> Something's a bit buggy with client.c and I'm seeing some requests fail with missing content length in the response. I'm wondering if we're reading a partial response and need to do more buffering? Unclear: haven't done any investigation yet.

I figured this out and wrote an update in the PR description.

TLDR: the existing `client.c` `send_request_and_read_response()` function isn't quite right. It needs a rework to support chunked transfer encoding. I started working on that in this branch and got something crude enough that it was functional but realized this area of the code needs a pretty substantial amount of love. If folks are open to it I'd like to defer that to its own follow-up PR.

src/rustls.h
```diff
@@ -1740,6 +1751,56 @@ rustls_result rustls_client_config_builder_set_key_log(struct rustls_client_conf
                                                        rustls_keylog_log_callback log_cb,
                                                        rustls_keylog_will_log_callback will_log_cb);
 
+/**
+ * Configure the client for Encrypted Client Hello (ECH).
+ *
+ * This requires providing a DER encoded list of ECH configurations that should
```
cpu, line 1757: reminder to self: these aren't DER. Need to do a consistency pass. It's a TLS encoded structure.

cpu:
> Figure out a solution for MSRV 1.71 - presently using 1.71 or 1.72 on this branch produces an [internal compiler error](https://gist.github.com/cpu/35dc574853a2ad185ba00ded256b20e2). I haven't investigated upstream yet. For now I switched CI to use 1.73 so I can chase this down separately.

I was able to reproduce the problem upstream in Rustls building the ech client example with 1.71 and 1.72. I reported the ICE to the Rust team here: https://github.com/rust-lang/rust/issues/133578

We can either block on a resolution for this upstream, land a `#[used]` workaround in Rustls and block on that, or declare MSRV 1.73 for this repo as I've done in this commit. I think which choice will be best likely depends on what kind of response 133578 generates.

src/rustls.h
cpu: fixed.

cpu:
> Figure out a solution for Rustls HPKE suites with MSRV 1.71 - presently using 1.71 or 1.72 on this branch produces an [internal compiler error](https://gist.github.com/cpu/35dc574853a2ad185ba00ded256b20e2). See https://github.com/rust-lang/rust/issues/133578

The upstream Rust project isn't interested in fixing bugs in old compilers (sensible!) so I think we're left with ~two options:

* Adding workarounds upstream in Rustls and then cutting a point release. This doesn't seem like a good trade-off. We're not 100% confident in the cause, or the workarounds, and there isn't any indication other projects have bumped into this ICE so far. Either due to the conditions required, or ECH related features being fairly niche.
* Increasing rustls-ffi's MSRV to 1.73. This is easy to do and requires no upstream changes. The userbase for rustls-ffi is considerably smaller than Rustls and I expect much more able to take the MSRV increase.

I vote for the latter option and have pushed a corresponding update.

cpu:
@ctz Would you be willing to +1 this branch?

I pulled the CI improvements out into a follow-up issue https://github.com/rustls/rustls-ffi/issues/498 There's work in-flight to unblock this test coverage but I'd like to merge this before waiting on that stuff if possible.

ctz (approved):

cpu:
I think this is self-contained enough and has been up for review long enough that I'm comfortable merging with just Ctz's review.

Happy to iterate further if there's any new feedback. I don't plan to cut a release tag for a bit so there's still time to adjust.

> Build+Test (ubuntu-latest, clang, 1.71, aws-lc-rs) Expected — Waiting for status to be reported

Going to admin merge this and then fix the MSRV in the branch rules (it's now 1.73)

cpu:
> and then fix the MSRV in the branch rules

Done.

Issue #503, opened 2024-12-13
client.c test ECH support updates


cpu:
Upstream we made some tweaks to the `ech-client.rs` example (https://github.com/rustls/rustls/pull/2278) that motivates making a few changes to this repo's equivalent.

* The `ech_fetch.rs` util should be updated to write `EchConfigListBytes` files for each of the HTTPS records fetched, not just the first.
* The `client.c` test should be updated to accept a directory of ECH config list files in an env var instead of a single file.
* We might want to make some API changes to make it easier to filter many config lists into one working client config.

cpu:
Fix is https://github.com/rustls/rustls-ffi/pull/504

> The client.c test should be updated to accept a directory of ECH config list files in an env var instead of a single file.

I decided it was easier to make the env var a comma separated list of file paths.

> We might want to make some API changes to make it easier to filter many config lists into one working client config.

It turns out the existing API is fine. It's pretty close to how you do this w/ the Rust API as well: try to convert each into an ECH mode. Stop for the first good one. Error if none are good.

PR #504, opened 2024-12-14
tests: support multiple HTTPS RRs for ECH configs

cpu:
Similar to a change in the [upstream Rustls ech-client.rs demo](https://github.com/rustls/rustls/pull/2278) we want to be able to process _multiple_ HTTPS records for a given domain, and look at each ECH config list from each record for a potential compatible config. Follow-up to https://github.com/rustls/rustls-ffi/pull/485.

Mechanically this means:

1. Updating the `test/ech_fetch.rs` helper to support writing multiple `.bin` files when there are multiple HTTPS records w/ ECH configs. The tool now outputs to stdout a comma separated list of the files it writes to make it easier to use with the `client.c` example.

2. Updating the `tests/client.c` example to treat the `RUSTLS_ECH_CONFIG_LIST` env var as a comma separated list of ECH config lists. We now loop through each and only fail if all of the provided files are unable to be used to configure the client config with a compatible ECH config. 

Doing string manipulation with C remains "a delight". For Windows compat we achieve tokenizing the string by the comma delim with a define to call either [`strtok_r`](https://linux.die.net/man/3/strtok_r) with GCC/clang, or [`strtok_s`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strtok-s-strtok-s-l-wcstok-s-wcstok-s-l-mbstok-s-mbstok-s-l?view=msvc-170) with MSCV.

You can test this update with:

```
ECH_CONFIG_LISTS=$(cargo test --test ech_fetch -- curves1-ng.test.defo.ie /tmp/curves1-ng.test.defo.ie)
RUSTLS_PLATFORM_VERIFIER=1 RUSTLS_ECH_CONFIG_LIST="$ECH_CONFIG_LISTS" ./cmake-build-debug/tests/client curves1-ng.test.defo.ie 443 /echstat.php?format=json
```

If you're unlucky and the first HTTPS record served is the one with invalid configs you should see output like the following showing the client skipping over the `.1` config list and using the `.2` one instead:

```
client[188911]: no compatible/valid ECH configs found in '/tmp/curves1-ng.test.defo.ie.1'
client[188911]: using ECH with config list from '/tmp/curves1-ng.test.defo.ie.2'
```

This logic will also need to be ported into #497. I'll do that shortly but expect we can land this PR first against the pre-existing `client.c` and then catch 497 up afterwards.

Resolves https://github.com/rustls/rustls-ffi/issues/503

### DIFF
--- a/tests/client.c
+++ b/tests/client.c
@@ -474,7 +474,9 @@ main(int argc, const char **argv)
     config_builder = rustls_client_config_builder_new();
   }
 
-  if(getenv("RUSTLS_ECH_GREASE")) {
+  const char *rustls_ech_grease = getenv("RUSTLS_ECH_GREASE");
+  const char *rustls_ech_config_list = getenv("RUSTLS_ECH_CONFIG_LIST");
+  if(rustls_ech_grease) {
     const rustls_hpke *hpke = rustls_supported_hpke();
     if(hpke == NULL) {
       fprintf(stderr, "client: no HPKE suites for ECH available\n");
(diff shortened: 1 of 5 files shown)


tests/client.c
```diff
@@ -489,39 +491,82 @@ main(int argc, const char **argv)
...
+        continue;
+      }
+
+      // Try to enable ECH with the config list. This may error if none
+      // of the ECH configs are valid/compatible.
+      result =
+        rustls_client_config_builder_enable_ech(config_builder,
+                                                (uint8_t *)ech_config_list_buf,
+                                                ech_config_list_len,
+                                                hpke);
```
ctz, line 548: I got myself in a bit of a pickle here wondering if `ech_config_list_buf` lives long enough.

`rustls_client_config_builder_enable_ech` says "The caller maintains ownership of the ECH config list TLS bytes", but not really if there is any lifetime requirement for it. From reading the docs I thought there was, but after reading the code I think there isn't.

ctz (approved):

tests/client.c
cpu: > From reading the docs I thought there was, but after reading the code I think there isn't.

Thanks for calling out that confusion. Presently the docs say:

> The caller maintains ownership of the ECH config list TLS bytes and `rustls_hpke` instance.

What if I amended it to say:

> The caller maintains ownership of the ECH config list TLS bytes and `rustls_hpke` instance. Both may be freed without affecting the `rustls_client_config_builder`.

cpu: I tacked on a commit that used this language:

> The lifetime of the `ech_config_list_bytes` and `rustls_hpke` instance are not tied to the lifetime of the `rustls_client_config_builder` and can be freed at any time after this function returns without affecting the builder.

ctz: Maybe "This function does not retain any reference to `ech_config_list_bytes`"? But I think any clarification is an improvement.

PR #505, opened 2024-12-16
Split up modules, support no built-in providers, small tidying

### LINKED ISSUE #231: Split up cipher.rs into cipher_suites.rs and certificates.rs
About half the functionality in cipher.rs is actually related to certificates, not cipher suites, so we should split it up.

cpu:
* Splits up `cipher.rs` and `lib.rs`. The former was quite large, and the latter should ideally be just mod definitions and not substantial code.

* Adds support for building rustls-ffi without a built-in provider. Mechanically this was always possible, but `cargo test` wouldn't work without some effort. The `client`/`server` examples continue to enforce that a provider is supplied. This build configuration only makes sense for downstream consumers that want to bring their own FFI-ready crypto provider.

* Tidies up Rust imports to match the Rustls convention.

* Removes single variant `rustls_result` enum imports.

* Fixes `cargo capi test` by ignoring the version test in this configuration.

* Removes a small acceptor test helper that wasn't carrying its weight.

Resolves #231


.github/workflows/test.yaml
```diff
@@ -298,6 +298,9 @@ jobs:
       - name: cargo test (debug; ring; -Z minimal-versions)
         run: cargo -Z minimal-versions test --no-default-features --features=ring --locked
 
+      - name: carg test (debug; no provider; -Z minimal-versions)
```
ctz, line 301: s/carg/cargo/

.github/workflows/test.yaml
```diff
@@ -298,6 +298,9 @@ jobs:
...
+        run: cargo -Z minimal-versions test --no-default-features --locked
```
ctz, line 302: it seems a little odd to do the -Z minimal-versions thing with both no-default-features *and* default features? I would think if default-features works, then no-default-features also must (since it must have the same or fewer dependencies in play?)

ctz (approved):

.github/workflows/test.yaml
cpu: Yeah I agree it's odd. I was mainly looking to (ab)use an existing `cargo test` invocation. I'll split it into a separate better described item.

.github/workflows/test.yaml
cpu: This fell away with a better name ("Test no built-in provider build") in a different area of CI away from the minimal versions stuff.

.github/workflows/test.yaml
cpu: I added a `cargo test --no-default-features --locked` step to the main CI job along with a descriptive comment about why existing test steps weren't sufficient to cover this particular use-case where you want no built-in providers because you bring your own 3rd party one.

cpu:
Since this is just moving existing code around, making cosmetic tweaks, or build-gating tests I'm going to merge w/o waiting for further review. Always happy to iterate further.

PR #515, opened 2024-12-23
license: Fix missed LICENSE-ISC link

yedayak:
This seems to have been missed in https://github.com/rustls/rustls-ffi/pull/509


cpu (approved):
Thanks for catching this :-)

PR #516, opened 2024-12-24
Two small breaking API changes before next release

### LINKED ISSUE #179: Change rustls_server_connection_get_server_name to return rustls_str
This function was written before we started using rustls_str, and it uses the pattern of writing to a caller-provided buffer. That pattern is quite awkward to use, and rustls_str fits better, so we should use that instead.

Note: #178 (lifetime issues for rustls_connection' getters) may get in the way, but I think the SNI field has the same set-once property we rely on for the other getters.

### LINKED ISSUE #229: Remove rustls_result return value from rustls_server_config_builder_set_persistence
This function is currently: https://github.com/rustls/rustls-ffi/blob/78736fb730ffedc6bbd8acf70aa3ab08ba2be467/src/server.rs#L654-L675

Note that the only possible error return cases are that the inputs were NULL. Our API guidelines state that in such a case, we don't bother to return a rustls_result (and make the user check it), but instead do nothing and return void.

cpu:
Knocking out a couple small issues labelled `next-major-release` since we're preparing a 0.15 in `main`. Sure do appreciate having `let .. else` stable in our MSRV :revolving_hearts: 

Resolves #179
Resolves #229 

### server: SNI rustls_connection getter ret rustls_str
The `rustls_server_connection_get_server_name()` API function previously wrote the SNI value to a caller-provided buffer as it predates the `rustls_str` pattern.

This commit updates it to return a `rustls_str`. Invalid parameters or incomplete connection state are returned as empty strs.

The documented lifetime matches that of other `rustls_connection` accessors based on a "set-once" property we assume from upstream.

### server: remove rustls_result from set_persistence

Previously `rustls_server_config_builder_set_persistence()` returned a `rustls_result` just to indicate if a required parameter was `NULL`.

Our API guidelines instead recommend this function do nothing and return void for these circumstances.

### DIFF
--- a/librustls/src/rustls.h
+++ b/librustls/src/rustls.h
@@ -2365,22 +2365,22 @@ rustls_result rustls_server_connection_new(const struct rustls_server_config *co
                                            struct rustls_connection **conn_out);
 
 /**
- * Copy the server name from the server name indication (SNI) extension to `buf`.
+ * Returns a `rustls_str` reference to the server name sent by the client in a server name
+ * indication (SNI) extension.
  *
- * `buf` can hold up  to `count` bytes, and the length of that server name in `out_n`.
+ * The returned `rustls_str` is valid until the next mutating function call affecting the
+ * connection. A mutating function call is one where the first argument has type
+ * `struct rustls_connection *` (as opposed to `const struct rustls_connection *`). The caller
+ * does not need to free the `rustls_str`.
  *
- * The string is stored in UTF-8 with no terminating NUL byte.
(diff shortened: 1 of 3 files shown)


cpu:
> cpu requested review from ctz and jsha now

This one should be pretty easy to review if you folks have a few min to spare. Thanks!

ctz (approved):

cpu:
These feels small enough to merge with 1 review. Thanks for taking a look ctz :bow:

Issue #507, opened 2024-12-17
FFI for rustls-post-quantum provider crate


cpu:
Rustls presently offers post quantum hybrid [X25519MLKEM768](https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem/02/) via a separate crate that augments the aws-lc-rs `CryptoProvider` with the post-qc algorithms, [rustls-post-quantum](https://github.com/rustls/rustls/tree/main/rustls-post-quantum). The plan is to eventually roll this into the main crate (https://github.com/rustls/rustls/issues/2056).

Previously I prototyped a FFI wrapper around this provider in [cpu/rustls-post-quantum-ffi](https://github.com/cpu/rustls-post-quantum-ffi) as a means of verifying external crypto providers worked with the FFI crypto provider API in this crate.

I think it makes sense to roll the FFI wrapper for the post-qc provider into the main rustls-ffi crate and offer it behind a default-off opt-in feature ala `fips` or `cert-compression`. Once the `rustls-post-quantum` crate is rolled into the main rustls crate and offered by default we can do the same for `rustls-ffi`.

cpu:
Related: https://github.com/rustls/rustls/pull/2288

PR #520, opened 2024-12-29
rustls 0.23.22, opt-in prefer-post-quantum feature flag

cpu:
### rustls 0.23.20 -> 0.23.22
Notably this brings in support for post-quantum-secure key exchange using X25519MLKEM768 with the `aws-lc-rs` crypto provider. Support is enabled by default, but at a low negotiation priority. A subsequent commit will expose the feature required to make it the most preferred KX alg.

See the [upstream release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.22) for more information:

### add prefer-post-quantum feature flag
When explicitly activated (it is not a default) this feature flag activates the matching upstream `rustls/prefer-post-quantum` feature flag, making the `aws-lc-rs` crypto provider prefer `X25519MLKEM768`.

Since the default provider is augmented with `prefer-post-quantum` the existing `client.c` and `server.c` examples require no changes to benefit from the `-DPREER_POST_QUANTUM=on` CMake option. CI is updated to test a preferred post-QC secure key exchange using `client.c` and  `pq.cloudflareresearch.com`, checking that the output reports the correct KX on Windows/MacOS/Linux. This is done in the daily tests job since it relies on an external host. 

Here's an [exemplar run](https://github.com/cpu/rustls-ffi/actions/runs/13161321339) from a manual invocation on this branch.

Resolves https://github.com/rustls/rustls-ffi/issues/507

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -130,3 +130,61 @@ jobs:
         shell: bash
         run: |
           grep 'sni=encrypted' ech-test.log
+
+  prefer-pq:
+    name: "Post-quantum (${{ matrix.os }})"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Install nightly toolchain
+        uses: dtolnay/rust-toolchain@nightly
+
+      - name: Install cargo-c (Ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        env:
+          LINK: https://github.com/lu-zero/cargo-c/releases/latest/download
+          CARGO_C_FILE: cargo-c-x86_64-unknown-linux-musl.tar.gz
+        run: |
+          curl -L $LINK/$CARGO_C_FILE | tar xz -C ~/.cargo/bin
+      - name: Install cargo-c (macOS)
+        if: matrix.os == 'macos-latest'
+        env:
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yaml
```diff
@@ -135,6 +148,12 @@ jobs:
           cmake --build build --target ech-test > ech-test.log
           grep 'sni=encrypted' ech-test.log
 
+      - name: Run PQ connect test
```
cpu, line 151: This will need to be reconciled with https://github.com/rustls/rustls-ffi/pull/517 depending on merge order.

cpu: Done.

cpu:
I think this should be updated to take the same approach as the [last update in rustls#2288](https://github.com/rustls/rustls/pull/2288#issuecomment-2575413951) - e.g. including the hybrid KX by default at the end of the supported KX list, and then promoting it to most-preferred with the feature flag enabled.

I'll do that shortly.

cpu:

This should be ready for review now. The general approach has changed a little bit from the first draft based on what landed upstream in 0.23.22. This side is simpler now, so that's nice :stars: I've updated the PR desc accordingly.

ctz (approved):

.github/workflows/daily-tests.yml
```diff
@@ -130,3 +130,61 @@ jobs:
         shell: bash
         run: |
           grep 'sni=encrypted' ech-test.log
+
+  prefer-:
```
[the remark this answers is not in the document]
cpu: Oops, I think this was an unintentional find/replace snafu. Fixed.

cpu:
Since this ended up not requiring any code changes in `librustls` I'm going to merge with the one review. Thanks!

Issue #526, opened 2025-01-16
Support installing from cmake, or emit an error when it's tried


cpu:
If folks miss the note that `cmake` is only required for the client/server example it might be tempting to try and use cmake to install rustls-ffi.

We're in bad shape for that case right now because it:

a) doesn't work! 
b) doesn't emit any kind of error!

Making it actually work might require work towards https://github.com/rustls/rustls-ffi/issues/519 since we're presently "installing" the library to a fixed location in the example binary build dir as a pre-req for the client/server binaries. 

Making it emit an error should be straight forward and an easy UX win.

Issue #519, opened 2024-12-28
Generalize CMake for examples to use external librustls


cpu:
As a follow-up to https://github.com/rustls/rustls-ffi/pull/506 it would be nice to remove the need for the 3rd party https://github.com/cpu/rustls-ffi-test repo for smoke-testing built artifacts.

The existing client/server examples would be perfect for this, except that the `CMakeLists.txt` they use unconditionally build `librustls` from scratch using `cargo-c` pointed to output into the cmake build root.

It would be better if we generalized this so that if an existing `librustls` was available (e.g. from being discovered via `pkg-config`) the build would prefer to use that copy of the lib instead of building it anew. This would let us use the test binaries with the pre-built artifacts during the testing phase of the binary artifact release CI process.

PR #532, opened 2025-01-20
Small CMake related tweaks

cpu:
### add doc header to CMakeLists.txt
At least one person convinced themselves CMake is the required/preferred way to build librustls despite the README documenting the opposite. Let's try to be even more overt about it by adding a header to the root `CMakeLists.txt`.

In the near future I'd like to lift all of the CMake + test C code into a separate crate for even clearer separation.

### rm CI dep on rustls-ffi-test for artifact test
Previously the `artifacts.yaml` CI job relied on an external repo that had a simple C/CMake application depending on librustls. When verifying artifacts we cloned this repo and tested a build against the artifact.

This commit reworks the main repo's test application code/build to support this use-case. The main tweak required is to support _not_ building librustls from src, but instead finding it with pkg-config. We gate this behind a new `FORCE_SYSTEM_RUSTLS` option. The artifact test CI is updated to use this option. We want to leave the old default because it's the most convenient for developers of librustls: you want to hack on the Rust code and have your C test code builds pick up the changes without extra ceremony.

Annoyingly for the pkg-config workflow with the release artifacts zips (not the `deb`) we have to patch the `.pc` prefix to an absolute dir to get this working nicely, which means handling Windows specially.

Similarly, for yet to be determined reasons, on Win32 we have to add some extra `target_link_libraries` that don't propagate automatically through the `.pc` config (_would be nice to sort this out one day_). For now we just shift the existing Win32 specific `target_link_libraries` invocation up and use it for both `pkg-config` librustls and build-it-ourselves librustls.

Resolves https://github.com/rustls/rustls-ffi/issues/519

### abort on CMake install w/ clear error

This updates the librustls `CMakeLists.txt` to abort on install. It's only intended to be used for the C examples, which we _don't_ want anyone to install anywhere!

Use cargo capi for librustls, not cmake.

```
$ cmake --install build
-- Install configuration: "Release"
CMake Error at build/cmake_install.cmake:46 (message):
  librustls installation via CMake is not supported.  Use 'cargo capi
  install' instead.

  See:
  https://github.com/rustls/rustls-ffi?tab=readme-ov-file#build-rustls-ffi
```

Resolves https://github.com/rustls/rustls-ffi/issues/526

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -163,10 +163,10 @@ jobs:
           - os: macos-13
             artifact: rustls-ffi-x86_64-macos
     steps:
-      - name: Checkout rustls-ffi-test sources
+      - name: Checkout sources
         uses: actions/checkout@v4
         with:
-          repository: 'cpu/rustls-ffi-test'
+          persist-credentials: false
       - name: Download rustls-ffi artifact
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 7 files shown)


cpu:
> cpu marked this pull request as draft now

I had removed the main CI yaml while iterating on this but it looks like I broke the normal CI workflow's build on Windows with the `target_link_libraries` change. I'll fix this shortly. Folks can hold review until then if desired.

cpu:
> looks like I broke the normal CI workflow's build on Windows with the target_link_libraries change. I'll fix this shortly.

All fixed. Just needed a small adjustment for consistent `target_link_options` form between two call-sites on the Win32 path.

@ctz If you could spare time for a review it would be appreciated (no big rush). Thanks!

cpu:
@jsha Would you have time to review this? :bow:

.github/workflows/artifacts.yaml
```diff
@@ -176,19 +176,29 @@ jobs:
       # the correct location that we extracted the archive. This seems more reliable
       # than using `--define-prefix` - it seems to tack an extra 'lib/' subcomponent
       # onto the include path that breaks the build.
-      - name: Fix pkg-config prefix
-        # We use bash shell explicitly to avoid PowerShell on Windows and to ensure we have 'sed'.
+      - name: Fix pkg-config prefix (UNIX)
+        if: matrix.os != 'windows-latest'
```
ctz, line 180: nit: I think `if: runner.os != 'Windows'` is a more direct expression of intent, in this hellhole of yaml and godawful nonsense

.github/workflows/artifacts.yaml
```diff
@@ -176,19 +176,29 @@ jobs:
...
-              sed -i '' "s|prefix=.*|prefix=${{ matrix.artifact }}|" ${{ matrix.artifact }}/lib/pkgconfig/rustls.pc
+              sed -i '' "s|prefix=.*|prefix=$(pwd)/${{ matrix.artifact }}|" ${{ matrix.artifact }}/lib/pkgconfig/rustls.pc
               ;;
             *)
-              sed -i "s|prefix=.*|prefix=${{ matrix.artifact }}|" ${{ matrix.artifact }}/lib/pkgconfig/rustls.pc
+              sed -i "s|prefix=.*|prefix=$(pwd)/${{ matrix.artifact }}|" ${{ matrix.artifact }}/lib/pkgconfig/rustls.pc
               ;;
           esac
+      - name: Fix pkg-config prefix (Windows)
+        if: matrix.os == 'windows-latest'
```
ctz, line 193:
```suggestion
        if: runner.os == 'Windows'
```

ctz (approved):

.github/workflows/artifacts.yaml
cpu: Good call. I fixed this in a separate commit because it turns out to be a pattern that I was using in the `test.yaml` and `daily-tests.yaml` workflows as well. I've switched all instances to use `runner.os` instead.

.github/workflows/artifacts.yaml
cpu: Fixed.

Issue #528, opened 2025-01-20
fips feature: both fips and non-fips aws-lc-sys is compiled and present in the library


xnox:
Using 7bb2ae7130b7c622324ab414fdf3a223543b5a96

Compiling using

```
cargo auditable capi install --features fips --locked --prefix=/usr --destdir "${{targets.contextdir}}"
```

Observing:

```
2025/01/20 14:34:20 WARN + cargo auditable capi install --no-default-features --features fips --locked '--prefix=/usr' --destdir /home/build/melange-out/rustls-ffi-fips
2025/01/20 14:34:20 WARN     Updating crates.io index
2025/01/20 14:34:23 WARN  Downloading crates ...

2025/01/20 14:34:23 WARN   Downloaded aws-lc-rs v1.12.0
2025/01/20 14:34:23 WARN   Downloaded aws-lc-fips-sys v0.13.0
2025/01/20 14:34:24 WARN   Downloaded aws-lc-sys v0.24.0
2025/01/20 14:34:24 WARN   Downloaded ring v0.17.8

2025/01/20 14:34:25 WARN    Compiling aws-lc-rs v1.12.0
2025/01/20 14:34:35 WARN    Compiling aws-lc-sys v0.24.0
2025/01/20 14:34:44 WARN    Compiling aws-lc-fips-sys v0.13.0

2025/01/20 14:36:46 WARN     Finished `release` profile [optimized] target(s) in 2m 26s
2025/01/20 14:36:46 WARN     Building pkg-config files
2025/01/20 14:36:46 WARN   Populating uninstalled header directory
2025/01/20 14:36:47 WARN   Installing pkg-config file
2025/01/20 14:36:47 WARN   Installing header file
2025/01/20 14:36:47 WARN   Installing static library
2025/01/20 14:36:47 WARN   Installing shared library
```

The compilation does not have ring, but does have two copies of aws-lc-*sys modules

```
$ find target/ -name '*.rlib' | grep -e libaws -e libring
target/x86_64-unknown-linux-gnu/release/deps/libaws_lc_sys-2525757ee4a12aa0.rlib
target/x86_64-unknown-linux-gnu/release/deps/libaws_lc_rs-5126331562e8f1b8.rlib
target/x86_64-unknown-linux-gnu/release/deps/libaws_lc_fips_sys-c5f3330d855b5592.rlib
```

But it does appear to have both aws-lc-sys and aws-lc-fips-sys symbols

```
[...]

PR #548, opened 2025-03-17
Cargo: rustls .22 -> .25, webpki 0.102.8 -> 0.103, rustls-platform-verifier 0.5 -> 0.5.1

cpu:
Updates rustls to 0.23.25, webpki 0.102.8 to 0.103, and rustls-platform-verifier to 0.5.1 (for compat w/ the updated rustls/webpki deps).

Alongside this update, we tidy our features slightly:

* The webpki "aws_lc_rs" feature is replaced with "aws-lc-rs". The underscore variant was an early mistake we undid with 0.103.
* The rustls-ffi "fips" feature now activates "rustls/fips" and "webpki/aws-lc-rs-fips". This avoids having both aws-lc-sys and aws-lc-fips-sys in FIPS builds.

These updates introduce a few new `rustls::CertificateError::*` variants (with more context) that we take care to map to our existing `rustls_error` variants for the respective underlying cause. Some various `error.rs` tidying comes along for the ride.

Resolves https://github.com/rustls/rustls-ffi/issues/528

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,7 +831,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


ctz:
I think we might want to extend this code? https://github.com/rustls/rustls-ffi/blob/b762f51a616d79b1d061fbd66c0b376736b57b51/librustls/src/error.rs#L373-L387

cpu:
> I think we might want to extend this code?

Thanks, that was a good catch.

I pushed a bunch of small commits to the end of this branch to fix that mapping and to clean up a bunch of low-hanging style nits that were bothering me in `error.rs`. There's still room for improvement, but it's in a more tolerable state now.

djc (approved):

ctz (approved):

PR #550, opened 2025-03-20
Cargo: hickory-resolver 0.25.0-alpha.4 -> 0.25

cpu:


djc (approved):
I think some of the Hickory dependency management isn't quite right yet, otherwise looks okay.

cpu:
> I think some of the Hickory dependency management isn't quite right yet

D'oh. Thankfully in this repo it's just used in [a tools crate](https://github.com/rustls/rustls-ffi/tree/main/tools) that's excluded from the workspace & not published. Happy to hear it's getting fixed upstream anyway, thank you.

PR #551, opened 2025-03-22
`rustls_error`: correct output buffer length handling

ctz:
This function does not need to keep an output buffer byte spare to zero-terminate the string, as it doesn't terminate the output.  So don't do that.

As an effect, this fixes the integer underflow followed by buffer overflow for `rustls_error` with an empty output buffer. `test_rustls_error_into_empty_buffer` is a regression test for that case.

(found while reviewing a use of this function in curl)


cpu (approved):
Good catch :+1:

djc (approved):

PR #554, opened 2025-03-25
Cargo: update ring 0.17.8 -> 0.17.14

cpu:
Updates the Cargo.lock for `*ring*` to clear [RUSTSEC-2025-0009](https://rustsec.org/advisories/RUSTSEC-2025-0009)

I don't believe it to be an issue that affects rustls-ffi in practice because:

1. ring is optional, and not the default crypto provider
2. there's no exposed API surface that would lead to `ring::aead::quic` usage
3. for `ring::aead::{AES_128_GCM, AES_256_GCM}`, the TLS protocol limits prevent the volume of data required to hit the panic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -203,9 +203,9 @@ checksum = "325918d6fe32f23b19878fe4b34794ae41fc19ddbe53b10571a4874d44ffd39b"
 
 [[package]]
 name = "cc"
-version = "1.2.5"
+version = "1.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c31a0499c1dc64f458ad13872de75c0eb7e3fdb0e67964610c914b034fc5956e"
+checksum = "1fcb57c740ae1daf453ae85f16e37396f672b039e00d9d866e07ddb24e328e3a"
 dependencies = [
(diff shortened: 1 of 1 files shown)


cpu:
I can't see why Dependabot skipped this one :thinking: That might be worth some further debugging.

djc (approved):

djc:
Perhaps we should have a cargo-deny run here, too?

cpu:
> Perhaps we should have a cargo-deny run here, too?

I'm onboard. I think ~only `webpki` has that today. Probably worth looking at both `rustls/rustls` and this repo.

kpcyrd:
FWIW, this is what I use to invoke cargo-deny, takes about 24s in total.

You're very welcome to copy/use/adapt this.

```yaml
  deny:
    runs-on: ubuntu-24.04
    steps:
    - uses: actions/checkout@v4
      with:
        persist-credentials: false
    - name: Run cargo deny
      run: |
        docker run --rm -v "$PWD:/src" -w /src alpine:edge sh -c '
        set -e
        apk add cargo cargo-deny
        exec cargo deny check
        '
```

djc:
I usually use the first-party GitHub Action.

PR #556, opened 2025-03-30
Added function rustls_connection_last_error_msg

tomas-stejskal-abra:
Hello,

I'm using rustls-ffi for an experimental project to implement TLS in an ERP system written in Delphi. One challenge I've encountered is the lack of detailed error messages.

This patch allows me to obtain the last error message originating from rustl_connection_ functions.

Regards,
Tom Stejskal


djc:
But note that some types already store the last error (in order to fuse the state), so not sure just storing an extra field is the best solution here.

PR #559, opened 2025-04-12
update semver compat. deps, fix cbindgen CI diff check

cpu:
Notably fixes two dev-dep RUSTSEC issues:

* https://github.com/advisories/GHSA-pg9f-39pc-qf8g for crossbeam-channel
* https://github.com/advisories/GHSA-rr8g-9fpq-6wmg for tokio

Along the way I also dug into the `main` CI failure that's been flagging a `cbindgen` diff for the `rustls.h` header. I believe this is an upstream regression and so to fix it we have to temporarily pin our `nightly` version. I've opened an upstream issue for this (https://github.com/rust-lang/rust/issues/139715).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.83"
+version = "0.1.88"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
+checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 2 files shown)


cpu:
> wip: try pinning nightly rust?

This seems to have done the trick, but is an unsatisfying solution.

djc:
Where do the guards come from? Is that related to the nightly somehow? Maybe there is a bug in a specific nightly?

cpu:
> Where do the guards come from? Is that related to the nightly somehow?

The guards come from mapping cargo features to `#define`'s. That in turn relies on expanding the crate content with specific features enabled. It's controlled in the `cbindgen.toml` file: https://github.com/rustls/rustls-ffi/blob/4d1d5d85a54d71057983bdbf4eb5bf5c7e107c1b/librustls/cbindgen.toml#L13-L21

cpu:

I think this is good to go. I've opened an upstream issue for what I believe is the `nightly` regression that broke our workflow. Depending on how that's triaged we can revisit the change here.

Cargo.lock
```diff
@@ -922,21 +925,20 @@ dependencies = [
...
+version = "0.12.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "32cf62eb4dd975d2dde76432fb1075c49e3ee2331cf36f1f8fd4b66550d32b6f"
+checksum = "a9321642ca94a4282428e6ea4af8cc2ca4eac48ac7a6a4ea8f33f76d0ce70926"
 dependencies = [
  "crossbeam-channel",
  "crossbeam-epoch",
  "crossbeam-utils",
- "once_cell",
+ "loom",
```
djc, line 935: Pretty sure loom is not supposed to be a lib dependency. IIRC most users guard with a `cfg` flag.

djc (approved):

Cargo.lock
cpu: `cargo tree | grep loom` and `cargo tree --invert loom` both produce no results. I assume this is something transitive and not our issue?

cpu: It looks like it's hickory-resover -> moka -> loom ?

cpu: https://github.com/moka-rs/moka/issues/501

PR #561, opened 2025-04-16
Build artifacts on ubuntu-22

ctz:
There are comments here motivating the choice of 20.04 as being
an older (= more compatible) glibc.  It is unclear whether that
is a current or vestigial requirement. (Thoughts @cpu ?)

If it's a current requirement, we can instead continue to build on ubuntu-20.04
by using docker hosted on ubuntu-latest.

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -31,7 +31,7 @@ jobs:
           curl -L "$env:LINK/$env:CARGO_C_FILE" -o cargo-c-windows-msvc.zip
           powershell -Command "Expand-Archive -Path cargo-c-windows-msvc.zip -DestinationPath $env:USERPROFILE\\.cargo\\bin -Force"
 
-      - name: Build rusts-ffi
+      - name: Build rustls-ffi
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/artifacts.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
           curl -L "$env:LINK/$env:CARGO_C_FILE" -o cargo-c-windows-msvc.zip
           powershell -Command "Expand-Archive -Path cargo-c-windows-msvc.zip -DestinationPath $env:USERPROFILE\\.cargo\\bin -Force"
 
-      - name: Build rusts-ffi
+      - name: Build rustls-ffi
```
cpu, line 34: Oops :-)

cpu (approved):
> It is unclear whether that is a current or vestigial requirement. (Thoughts @cpu ?)

It's generally something that still applies in the sense that building on 22.04 will produce binaries with a higher minimum supported glibc than doing the same on 20.04. 

IIRC I was just trying to make the release binaries as flexible as possible where it was convenient, and it wasn't something informed by deep research into the glibc version distribution in the wild or anything.

With that in mind it feels reasonable to me to follow the LTS deprecation that the runners are going through. If someone desperately needs support for older glibc they can always build from source.

PR #564, opened 2025-04-28
Skip push CI for most branches

djc:
My impression is that the common practice of running CI for both `push` and `pull_request` can be pretty expensive/slow while adding almost no value. Restrict branch CI to `main` and branches prefixed with `rel-` or `ci/`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,6 +5,7 @@ permissions:
 
 on:
   push:
+    branches: ['main', 'rel-*', 'ci/*']
   pull_request:
   merge_group:
   schedule:


.github/workflows/test.yaml
```diff
@@ -5,6 +5,7 @@ permissions:
 
 on:
   push:
+    branches: ['main', 'rel-*', 'ci/*']
   pull_request:
```
ctz, line 9:
```suggestion
  pull_request:
  workflow_dispatch:
```

This would mean, in the case where some maintainer really does want to run CI for something without opening a PR, they can do that by pushing the branch and then running it manually (this enables the UI for that).

ctz (approved):

djc:
> > Restrict branch CI to main and branches prefixed with rel- or ci/.
> 
> Do you plan to do this in the other repos, e.g. `rustls/rustls`? I think it would be valuable to do this consistently, at least in the magic prefix/suffix to get it to happen for a branch push intentionally. E.g. `rustls-platform-verifier` is [using `*_dev`](https://github.com/rustls/rustls-platform-verifier/blob/9c4c17d6950fe9a368443cf754197b442c93b87f/.github/workflows/ci.yml#L5), can we standardize on that?

Would @ctz's suggestion of adding `workflow_dispatch` alleviate your concerns? Sounds like that might be helpful.

djc:
> > 
> 
> Do you plan to do this in the other repos, e.g. `rustls/rustls`? I think it would be valuable to do this consistently, at least in the magic prefix/suffix to get it to happen for a branch push intentionally. E.g. `rustls-platform-verifier` is [using `*_dev`](https://github.com/rustls/rustls-platform-verifier/blob/9c4c17d6950fe9a368443cf754197b442c93b87f/.github/workflows/ci.yml#L5), can we standardize on that?

Yes, I can probably do this for the other repos -- agreed on the value of consistency. Do you prefer using `*_dev`? I like that `ci/` is a little more explicit about the goal (and I sort of generally prefer `/` or `-` over `_`, which might be silly), but don't feel too strongly (and given that I just open PRs directly my workflow wouldn't be impacted).

cpu:
> Do you prefer using *_dev? I like that ci/ is a little more explicit about the goal

No strong preference here. `ci/` seems OK if you like that one. I don't think anyone will mind a change from `*_dev` on the rustls-platform-verifier repo, and I don't think we restrict CI by branch name in any other org repos presently.

cpu (approved):

PR #566, opened 2025-05-06
rustls 0.23.25 -> 0.23.27

cpu:
The Rustls [0.23.26](https://github.com/rustls/rustls/releases/tag/v%2F0.23.26) release didn't add any interesting API surface, but the [0.23.27](https://github.com/rustls/rustls/releases/tag/v%2F0.23.27) release did make some changes that are worth updates here:

* A new `rustls_client_connection_new_alpn()` function is added to match the new [`ClientConnection::new_with_alpn()`](https://docs.rs/rustls/0.23.27/rustls/client/struct.ClientConnection.html#method.new_with_alpn) fn.
* Our `map_invalid_certificate_error()` fn is updated to handle the new `CertificateError::InvalidPurposeContext` error.
* The `prefer-post-quantum` feature flag is promoted to enabled-by-default to match the upstream Rustls change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,11 +101,10 @@ platforms see the upstream documentation:
 #### Post-Quantum X25519MLKEM768 Key Exchange
 
 Post-quantum-secure key exchange using [X25519MLKEM768][] is supported when using the `aws-lc-rs` 
-cryptography provider. At this time default support places `X25519MLKEM768` at a lower negotiation priority. 
+cryptography provider and offered by default at the highest priority.
 
-By enabling the `prefer-post-quantum` feature flag the `X25519MLKEM768` key exchange will be used as the most
-preferred key exchange algorithm. We expect to add this feature to the crate's default features in a future
-release.
+By disabling the `prefer-post-quantum` feature flag the `X25519MLKEM768` key exchange will be
+offered at a lower negotiation priority.
 
 [X25519MLKEM768]: https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem
 
--- a/librustls/src/client.rs
+++ b/librustls/src/client.rs
@@ -768,22 +768,19 @@ impl rustls_client_config {
                 }
                 CStr::from_ptr(server_name)
             };
-            let config = try_clone_arc!(config);
-            let conn_out = try_mut_from_ptr_ptr!(conn_out);
-
             let Ok(server_name) = server_name.to_str() else {
                 return rustls_result::InvalidDnsNameError;
             };
             let Ok(server_name) = server_name.try_into() else {
                 return rustls_result::InvalidDnsNameError;
             };
-            let client = ClientConnection::new(config, server_name).unwrap();
 
-            // We've succeeded. Put the client on the heap, and transfer ownership
-            // to the caller. After this point, we must return rustls_result::Ok so the
-            // caller knows it is responsible for this memory.
-            let c = Connection::from_client(client);
-            set_boxed_mut_ptr(conn_out, c);
+            set_boxed_mut_ptr(
+                try_mut_from_ptr_ptr!(conn_out),
+                Connection::from_client(
+                    ClientConnection::new(try_clone_arc!(config), server_name).unwrap(),
+                ),
+            );
             rustls_result::Ok
         }
     }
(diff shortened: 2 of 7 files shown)


librustls/src/client.rs
```diff
@@ -770,14 +770,11 @@ impl rustls_client_config {
...
-                Err(std::str::Utf8Error { .. }) => return rustls_result::InvalidDnsNameError,
-            };
-            let server_name = match server_name.try_into() {
-                Ok(sn) => sn,
-                Err(_) => return rustls_result::InvalidDnsNameError,
+
+            let Ok(server_name) = server_name.to_str() else {
+                return rustls_result::InvalidDnsNameError;
             };
+            let Ok(server_name) = server_name.try_into() else { return rustls_result::InvalidDnsNameError; }
```
djc: Nit: no rustfmt because macro? I think this should get a similar style as the `server_name` one.

librustls/src/client.rs
```diff
@@ -768,20 +768,19 @@ impl rustls_client_config {
...
-            let client = ClientConnection::new(config, server_name).unwrap();
-
-            // We've succeeded. Put the client on the heap, and transfer ownership
-            // to the caller. After this point, we must return rustls_result::Ok so the
-            // caller knows it is responsible for this memory.
-            let c = Connection::from_client(client);
-            set_boxed_mut_ptr(conn_out, c);
+            let Ok(server_name) = server_name.try_into() else {
+                return rustls_result::InvalidDnsNameError;
+            };
```
djc: Nit: ahh, so this belongs in the previous commit.

librustls/src/client.rs
```diff
@@ -784,6 +787,74 @@ impl rustls_client_config {
             rustls_result::Ok
         }
     }
+
+    /// Create a new client `rustls_connection` with custom ALPN protocols.
```
djc, line 791: Did you consider having `_new()` call `_new_alpn()` like we do in Rust-land? Seems nice to avoid the duplication?

djc (approved):

README.md
```diff
@@ -101,11 +101,10 @@ platforms see the upstream documentation:
...
 
 Post-quantum-secure key exchange using [X25519MLKEM768][] is supported when using the `aws-lc-rs` 
-cryptography provider. At this time default support places `X25519MLKEM768` at a lower negotiation priority. 
+cryptography provider and offered by default at the highest priority.
 
-By enabling the `prefer-post-quantum` feature flag the `X25519MLKEM768` key exchange will be used as the most
-preferred key exchange algorithm. We expect to add this feature to the crate's default features in a future
-release.
+By disabling the `prefer-post-quantum` feature flag the `X25519MLKEM768` key exchange will be used as the most
+preferred key exchange algorithm.
```
ctz, line 107: I think the sense is inverted here:

```suggestion
By disabling the `prefer-post-quantum` feature flag the `X25519MLKEM768` key exchange will be offered as the least
preferred key exchange algorithm.
```

ctz (approved):

librustls/src/client.rs
cpu: I was avoiding that initially to not require an extra `clone()` for the `_new()` path, but since that will happen internal to rustls anyway in retrospect it's better to avoid the duplication. I'll adjust.

librustls/src/client.rs
cpu: We do [a special `sed` dance](https://github.com/cpu/rustls-ffi/blob/4cf5a4b77065fb48ff18c06322a31414f667e2b8/librustls/cmake/format.cmake#L1-L7) in this repo (and `rustls-openssl-compat`) to make rustfmt work inside the macros.

It's easy to forget :'( In this case it looks like I probably ran vanilla `cargo fmt` and then later `cmake --target format-fix` with the `sed` version.

All fixed!

librustls/src/client.rs
cpu: Exactly. Fixed.

librustls/src/client.rs
cpu: Reworked so both versions share a common impl.

README.md
cpu: Thanks for catching that. Fixed!

PR #577, opened 2025-06-24
website: Remove backdrop-filter: blur from items

yedayak:
This causes a significant slowdown and lag during scrolling in firefox on my machine, and it seems like this isn't an issue only on my machine: https://old.reddit.com/r/Frontend/comments/186u5wy/hows_the_performance_with_css_blur_filter_and/

### DIFF
--- a/website/static/style.css
+++ b/website/static/style.css
@@ -95,7 +95,6 @@ code {
 
 .item {
   background: rgba(255, 255, 255, 0.07);
-  backdrop-filter: blur(2px);
   border: 1px solid rgba(255, 255, 255, 0.1);
   border-radius: 1rem;
   margin: 1.5rem 0;


cpu (approved):
Thanks! Definitely not worth a perf hit for this :-)

ctz (approved):
Will look into doing this on the rustls.dev site too. Though I'd say that site is a bit more aesthetic-oriented than this one.

yedayak:
> Will look into doing this on the rustls.dev site too. Though I'd say that site is a bit more aesthetic-oriented than this one.

I looked a bit and could find a page there long enough to actually trigger an issue for me (specifically scrolling a large amount at once was an issue in ffi.rustls.dev)

PR #579, opened 2025-07-03
Take updated dependencies

ctz:
Notably this means `rustls_platform_server_cert_verifier_with_provider` can now return NULL. Avoiding that would mean using an out param and returning a `rustls_result`, but would be a major version bump. Thoughts?

(I checked and curl uses `rustls_platform_server_cert_verifier` and is therefore suitably fallible.)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -414,6 +414,9 @@ jobs:
   miri:
     name: Miri
     runs-on: ubuntu-latest
+    env:
+      # allows platform-verifier initialisation to inspect filesystem
+      MIRIFLAGS: -Zmiri-disable-isolation
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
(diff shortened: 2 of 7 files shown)


djc:
> Notably this means `rustls_platform_server_cert_verifier_with_provider` can now return NULL. Avoiding that would mean using an out param and returning a `rustls_result`, but would be a major version bump. Thoughts?

Probably not worth doing a major bump for? Do we have a deprecation mechanism here? Stick with 0.5 until rustls 0.24 comes out?

.github/workflows/test.yaml
```diff
@@ -414,6 +414,9 @@ jobs:
   miri:
     name: Miri
     runs-on: ubuntu-latest
+    env:
+      # allows platform-verifier initialisation to inspect filesystem
+      MIRIFLAGS: -Zmiri-disable-isolation
```
cpu, line 419: Huh, TIL this exists. I think there are a handful of pre-existing tests annotated to skip running under miri because they do simple file I/O. Separate from this PR maybe it's worth revisiting those with this flag enabled?

cpu (approved):
> Probably not worth doing a major bump for? 

Yeah, probably not, but the README's ABI stability claims mean consumers that take our .so need to treat every release as a major bump. I'd personally be inclined to wait for more changes to pile up in `main` to do any kind of release (or revisiting if there's an explicit request from downstream).

> Do we have a deprecation mechanism here?

I think `cbindgen` supports translating `#[deprecated]` into `__attribute__((deprecated))` annotations in the .h if that's what you meant :thinking:

.github/workflows/test.yaml
ctz: Did a pass and reenabled as many as possible. Unfortunately that was only one, because miri is pretty allergic to ffi/crypto and most other tests involve that.

djc (approved):

PR #581, opened 2025-07-08
docgen: skip over cpp attributes before decl

ctz:
Aiming to fix https://github.com/rustls/rustls-ffi/actions/runs/16144845989

(getting the deprecation message into the docs would be a nice to do in the future)


ctz:
https://github.com/rustls/rustls-ffi/actions/runs/16146216499 (nb deploy is intended to not work from a branch)

PR #589, opened 2025-07-30
librustls: update rustls 0.23.29 -> 0.23.31

cpu:
There were no upstream changes that motivate API updates in this crate:

* [0.23.31](https://github.com/rustls/rustls/releases/tag/v%2F0.23.31)
* [0.23.30](https://github.com/rustls/rustls/releases/tag/v%2F0.23.30)

### DIFF
--- a/librustls/Cargo.toml
+++ b/librustls/Cargo.toml
@@ -30,7 +30,7 @@ prefer-post-quantum = ["aws-lc-rs", "rustls/prefer-post-quantum"]
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "0.23.31", default-features = false, features = ["std", "tls12"] }
+rustls = { version = "=0.23.31", default-features = false, features = ["std", "tls12"] }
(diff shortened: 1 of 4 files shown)


librustls/Cargo.toml
```diff
@@ -30,7 +30,7 @@ prefer-post-quantum = ["aws-lc-rs", "rustls/prefer-post-quantum"]
 
 [dependencies]
 # Keep in sync with RUSTLS_CRATE_VERSION in build.rs
-rustls = { version = "0.23.29", default-features = false, features = ["std", "tls12"] }
+rustls = { version = "0.23.31", default-features = false, features = ["std", "tls12"] }
```
ctz, line 33: Ah, I had previously updated the lockfile in #588 -- should this version spec be a "=0.23.31", if we want `RUSTLS_CRATE_VERSION` to be accurate? Or alternatively we could alter the `librustls/tests/rustls_version.rs` test to look at the lockfile instead? (Happy to do the legwork on the latter.)

ctz (approved):

PR #606, opened 2025-09-16
Update dependencies

djc:
tree-sitter stuff still can't be updated, see

- https://github.com/tree-sitter-grammars/tree-sitter-markdown/pull/198

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1210,7 +1210,6 @@ version = "0.15.0"
 dependencies = [
  "libc",
  "log",
- "regex",
  "rustls",
  "rustls-platform-verifier",
  "rustls-webpki",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1408,11 +1408,11 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "0.6.9"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf41e0cfaf7226dca15e8197172c295a782857fcb97fad1808a166870dee75a3"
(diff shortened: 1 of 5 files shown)


tools/tests/rustls_header_version.rs
```diff
@@ -15,27 +11,10 @@ use tree_sitter::{Parser, Query, QueryCursor};
...
-        .as_table()
-        .expect("missing package metadata");
-    let crate_version = package_metadata["version"]
-        .as_str()
-        .expect("missing crate version");
-
     let version_in_header = version_in_header();
     assert_eq!(
-        crate_version, version_in_header,
+        env!("CARGO_PKG_VERSION"),
```
ctz, line 16: unfortunately this is a different package to librustls.

djc: Reverted that change for now.

djc:
This seems pretty surprising:

```rust
error[E0599]: no method named `copysign` found for type `f64` in the current scope
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/toml-0.9.6/src/value.rs:899:27
    |
899 |             value = value.copysign(1.0);
    |                           ^^^^^^^^ method not found in `f64`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `toml` (lib) due to previous error
```

Screenshot 2025-09-16 at 13 34 25

Screenshot 2025-09-16 at 13 34 45

Uhhh... what am I missing?

djc:
> > Uhhh... what am I missing?
> 
> Looks like toml without the `std` feature actually has a MSRV of 1.84 because it requires that method from `core`: https://doc.rust-lang.org/beta/releases.html#version-1840-2025-01-09

Ohh, nice work figuring that one out.

ctz (approved):

PR #649, opened 2026-06-22
Take semver-compatible dependency updates

djc:


cpu:
I think you need to bump `rustls` separately (and match the build.rs) because it's a `=` dep in `librustls`: https://github.com/rustls/rustls-ffi/blob/5ecaf1b06a7bd60fa2710d8bf3bb1e5cff046430/librustls/Cargo.toml#L32-L33

ctz:
> [rustls-ffi / Miri (pull_request)](https://github.com/rustls/rustls-ffi/actions/runs/28022437942/job/82941831162?pr=649)Failing after 9m

This actually comes from the rustls update (specifically https://github.com/rustls/rustls/commit/a612901957ad6f8c2e6a61e79dc776ef1004b087). Basically ~all of aws-lc-rs public API requires some FFI code, and now this test (via the above commit) calls into aws-lc-rs earlier.

cpu:
> Basically ~all of aws-lc-rs public API requires some FFI code, and now this test (via the above commit) calls into aws-lc-rs earlier.

I tacked on a commit that adds a miri exclude to two tests. I did it up-front because on my system I was seeing [both fail w/ tip-of-main under miri](https://gist.github.com/cpu/a3259683e215caaff3dd1e83244cc47b) from `rustls-platform-verifier`'s transitive deps doing unsupported I/O on Linux even without the rustls update. Disabling isolation seemed to make the tests hang, and we'll need to disable them anyway for the aws-lc-rs update & early FFI :shrug:

cpu (approved):

djc:
I was thinking we might use graviola for Miri instead... but the kids are home from school (due in part to the heat) so not much desk time for me.

cpu:
> I was thinking we might use graviola for Miri instead...

I think we'll still be running into issues with `test_config_builder()` + `rustls_platform_verifier` wanting to do I/O under miri on Linux. I think there would also be some more work required to wire up Graviola in this crate.

The other test I ignored (`rustls_version_match()`) probably shouldn't have ever run under Miri. The value there seems pretty dubious. 

None of that is to say I'm not supportive of finding a way to maximize the value of using Miri in this crate (or extending Graviola support), just that I think it's better done as a separate follow-up task to unblock keeping `main` building clean + sync'd with latest rustls.

ctz:
I don't think miri will support graviola either unfortunately. While Graviola is 100% rust, miri's lack of support for FFI is the same underlying problem as inline assembly and intrinsics.

PR #653, opened 2026-07-12
server: add rustls_server_config_builder_set_send_tls13_tickets

### LINKED ISSUE #652: Expose a way to set ServerConfig::send_tls13_tickets (disable TLS 1.3 session ticket issuance)
## The ask

rustls-ffi currently exposes no way to set `ServerConfig::send_tls13_tickets` (or otherwise disable TLS 1.3 NewSessionTicket issuance from the server side). `ServerConfigBuilder` in `librustls/src/server.rs` has fields for `provider`, `versions`, `verifier`, `cert_resolver`, `session_storage`, `alpn_protocols`, `ignore_client_order`, `key_log` — but no `ticketer` / `send_tls13_tickets` — and `rustls_server_config_builder_build` never touches `config.send_tls13_tickets`, so it stays at the rustls default of 2.

A `rustls_server_config_builder_set_num_tickets(builder, n)` (or equivalent) would cover it.

## Why it matters: benchmark equivalence

For TLS-library performance comparison benchmarks, the NewSessionTicket issuance cost after a full handshake is a measurable, non-trivial distortion. rustls upstream already established this in PR #2187 ("server: default send_tls13_tickets 4 -> 2"):

> the extra 2 tickets cost us 8%/6% (client/server) of a normal resumption, or ~0%/2% of a full handshake. This also manifests when comparing default configurations between TLS libraries due to the high cost of creating new tickets ... It also avoids folks repeatedly tripping into this during benchmarking.

Both peer libraries benchmarkers compare against expose a knob to suppress this:

- OpenSSL: `SSL_CTX_set_num_tickets(ctx, 0)`
- rustls (Rust API): `server.send_tls13_tickets = 0`

rustls-ffi is the one path that can't. A C/Zig/Python benchmark driving rustls through the FFI cannot get a clean full-1-RTT-handshake measurement that is equivalent to its OpenSSL/rustls-Rust peers — the rustls-ffi handshake row measures "handshake + 2 NewSessionTicket issuance" while the peers measure "handshake only."

The workaround of building a fresh `rustls_client_config` + `rustls_server_config` per connection prevents *resumption* (the client offering a PSK) but does **not** prevent *ticket issuance* (the server still generates and sends 2 tickets). It only solves half the problem.

## Context

- Client-side resumption override is also missing (the `ClientConfigBuilder` has no `resumption` field; `rustls_client_config_builder_build` leaves `config.resumption` at the rustls default). Tracked partially in #293 for client session storage. This issue is specifically about the server-side ticket count, which is the load-bearing one for handshake benchmark equivalence.
- This came up while evaluating whether to drive rustls from a Zig benchmark harness via rustls-ffi instead of a standalone Rust program. The inability to disable ticket issuance is the one thing forcing the benchmark to stay a Rust program.

## What would unblock

Either:
1. `rustls_server_config_builder_set_num_tickets(*mut builder, usize)` mirroring `SSL_CTX_set_num_tickets`, or
2. Exposing `send_tls13_tickets` as a builder field settable before `build()`.

Happy to draft a PR if there's appetite for a specific approach.

mattrobenolt:
Closes #652.

> [!NOTE]  
> This PR was AI generated using GLM 5.2.

## What

Adds `rustls_server_config_builder_set_num_tickets(builder, n)` to expose `ServerConfig::send_tls13_tickets` through the C API. Mirrors the existing `rustls_server_config_builder_set_ignore_client_order` pattern: a new `Option<usize>` field on `ServerConfigBuilder`, a setter, and application at `build()` time. Defaults to `None` (rustls's own default of 2 applies), so this is a strictly additive, no-behavior-change-by-default feature.

## Why

For TLS-library performance comparison benchmarks, NewSessionTicket issuance after a full handshake is a measurable distortion. rustls upstream established this in [PR #2187](https://github.com/rustls/rustls/pull/2187) ("server: default send_tls13_tickets 4 -> 2"):

> the extra 2 tickets cost us 8%/6% (client/server) of a normal resumption, or ~0%/2% of a full handshake. This also manifests when comparing default configurations between TLS libraries due to the high cost of creating new tickets ... It also avoids folks repeatedly tripping into this during benchmarking.

Both peer libraries expose a knob to suppress this:

- OpenSSL: `SSL_CTX_set_num_tickets(ctx, 0)`
- rustls (Rust API): `server.send_tls13_tickets = 0`

rustls-ffi was the one path that couldn't, so a C/Zig/Python benchmark driving rustls through the FFI could not get a handshake measurement equivalent to its peers — the rustls-ffi handshake row measured "handshake + 2 NewSessionTicket issuance" while peers measured "handshake only." The workaround of building a fresh config per connection prevents *resumption* (client offering PSK) but not *ticket issuance* (server still generates and sends the tickets).

## Changes

- `librustls/src/server.rs`: `send_tls13_tickets: Option<usize>` field on `ServerConfigBuilder` (init `None` in both `_new` and `_new_custom`), `rustls_server_config_builder_set_num_tickets` setter, applied in `rustls_server_config_builder_build`. Test `test_server_config_builder_set_num_tickets` covering both `0` and a non-zero value.
- `librustls/src/rustls.h`: vendored header declaration.

## Verification

- `cargo test --no-default-features --features ring --locked` — 35 passed (incl. new test)
- `cargo fmt --check` — clean
- `cargo clippy --no-default-features --features ring --all-targets -- -D warnings -A unknown-lints` — no warnings in changed code

(No `cargo-c`/capi build run locally; the C examples need cmake which isn't in my env. The header change is a straightforward declaration mirroring `set_ignore_client_order`.)

## Notes

- No effect on TLS 1.2 connections (`send_tls13_tickets` is TLS 1.3 only in rustls).
- The name `set_num_tickets` mirrors OpenSSL's `SSL_CTX_set_num_tickets` for discoverability.
- This came out of evaluating whether to drive rustls from a Zig benchmark harness via rustls-ffi; the inability to disable ticket issuance was the one blocker (see #652).

### DIFF
--- a/librustls/src/server.rs
+++ b/librustls/src/server.rs
@@ -60,6 +60,7 @@ pub(crate) struct ServerConfigBuilder {
     session_storage: Option<Arc<dyn StoresServerSessions + Send + Sync>>,
     alpn_protocols: Vec<Vec<u8>>,
     ignore_client_order: Option<bool>,
+    send_tls13_tickets: Option<usize>,
     key_log: Option<Arc<dyn KeyLog>>,
 }
 
@@ -98,6 +99,7 @@ impl rustls_server_config_builder {
                 session_storage: None,
                 alpn_protocols: vec![],
                 ignore_client_order: None,
+                send_tls13_tickets: None,
                 key_log: None,
             };
             to_boxed_mut_ptr(builder)
@@ -153,6 +155,7 @@ impl rustls_server_config_builder {
                 session_storage: None,
                 alpn_protocols: vec![],
                 ignore_client_order: None,
+                send_tls13_tickets: None,
                 key_log: None,
             };
             set_boxed_mut_ptr(builder_out, builder);
@@ -271,6 +274,22 @@ impl rustls_server_config_builder {
         }
     }
 
+    /// Set the number of TLS 1.3 NewSessionTickets sent after a full handshake.
+    ///
+    /// Setting this to 0 disables session ticket issuance entirely.
+    /// <https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets>
+    #[no_mangle]
+    pub extern "C" fn rustls_server_config_builder_set_send_tls13_tickets(
+        builder: *mut rustls_server_config_builder,
+        n: usize,
+    ) -> rustls_result {
+        ffi_panic_boundary! {
+            let config = try_mut_from_ptr!(builder);
+            config.send_tls13_tickets = Some(n);
+            rustls_result::Ok
+        }
+    }
+
     /// Set the ALPN protocol list to the given protocols.
     ///
     /// `protocols` must point to a buffer of `rustls_slice_bytes` (built by the caller)
@@ -375,6 +394,9 @@ impl rustls_server_config_builder {
             config.alpn_protocols = builder.alpn_protocols;
             if let Some(ignore_client_order) = builder.ignore_client_order {
                 config.ignore_client_order = ignore_client_order;
+            }
+            if let Some(send_tls13_tickets) = builder.send_tls13_tickets {
+                config.send_tls13_tickets = send_tls13_tickets;
             }
 
             if let Some(key_log) = builder.key_log {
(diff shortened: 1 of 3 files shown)


librustls/src/server.rs
```diff
@@ -60,6 +60,13 @@ pub(crate) struct ServerConfigBuilder {
     session_storage: Option<Arc<dyn StoresServerSessions + Send + Sync>>,
     alpn_protocols: Vec<Vec<u8>>,
     ignore_client_order: Option<bool>,
+    // Number of TLS 1.3 NewSessionTicket messages to send after a full
+    // handshake. rustls' default is 2. Setting 0 disables ticket issuance,
+    // which is useful for benchmark harnesses that want to measure a clean
+    // full 1-RTT handshake without NewSessionTicket issuance cost (mirroring
+    // OpenSSL's `SSL_CTX_set_num_tickets(ctx, 0)`).
+    // <https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets>
```
djc: Let's not have the big docstring here.

librustls/src/server.rs
```diff
@@ -271,6 +280,29 @@ impl rustls_server_config_builder {
         }
     }
 
+    /// Set the number of TLS 1.3 NewSessionTicket messages the server will send
```
djc: Please take the guidelines for comments into account:

https://github.com/rustls/rustls/pull/3130

librustls/src/server.rs
cpu: There also isn't a pattern in this codebase of citing equivalent OpenSSL libssl APIs, let's lose that.

librustls/src/server.rs
```diff
@@ -271,6 +280,29 @@ impl rustls_server_config_builder {
...
+    /// after a full 1-RTT handshake. rustls' default is 2.
+    ///
+    /// Setting this to 0 disables session ticket issuance entirely, so each
+    /// handshake is a clean full handshake with no post-handshake
+    /// NewSessionTicket work. This mirrors OpenSSL's
+    /// `SSL_CTX_set_num_tickets(ctx, 0)` and is useful for benchmark harnesses
+    /// comparing TLS libraries on equivalent full-handshake work.
```
cpu: I think it's sufficient here to say that "Setting this to 0 disables session ticket issuance entirely". 

Like my other comment I'd rather not cite OpenSSL APIs, and I don't think the benchmarking use-cae is important enough to capture in the API docs.

cpu (commented):
Can you squash your commits as well? Thanks

mattrobenolt:
Thanks @djc @cpu. Pushed an update addressing all the feedback:

- Removed the docstring on the `send_tls13_tickets` struct field (it is `pub(crate)`, not publicly exposed, so no doccomment per the RFC 1574 guidelines).
- Trimmed the setter doc comment to a short summary + the "Setting this to 0 disables session ticket issuance entirely" behavior, matching the style of `rustls_server_config_builder_set_ignore_client_order`. Dropped the OpenSSL `SSL_CTX_set_num_tickets` citation and the benchmarking rationale from the API docs.
- Regenerated/aligned the vendored `rustls.h` declaration to match cbindgen output.
- Squashed into a single commit (e0a4c3d).

librustls/src/server.rs
djc: This first sentence should fit on a single line, as described by the link.

djc (approved):

librustls/src/server.rs
```diff
@@ -820,6 +842,64 @@ mod tests {
...
+            builder,
+            &certified_key,
+            1,
+        );
+        rustls_server_config_builder::rustls_server_config_builder_set_num_tickets(builder, 4);
+        let mut config = null();
+        let result =
+            rustls_server_config_builder::rustls_server_config_builder_build(builder, &mut config);
+        assert_eq!(result, rustls_result::Ok);
+        {
```
cpu, line 900: Missing an `assert!(!config.is_null());` here to match the one on L875

librustls/src/server.rs
```diff
@@ -271,6 +274,22 @@ impl rustls_server_config_builder {
         }
     }
 
+    /// Set the number of TLS 1.3 NewSessionTicket messages the server will send
+    /// after a full handshake. Setting this to 0 disables session ticket
+    /// issuance entirely.
+    /// <https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets>
+    #[no_mangle]
+    pub extern "C" fn rustls_server_config_builder_set_num_tickets(
```
cpu: I would lightly prefer we match the upstream Rustls API here and call this `rustls_server_config_builder_set_send_tls13_tickets()`. I think that's generally what we've done with other fields like `ignore_client_order` and makes the TLS 1.3 specific aspect more apparent.

mattrobenolt:
Pushed another update (831186e) addressing the latest feedback:

- @djc: the setter doc summary is now a single line, per RFC 1574 §"Summary sentence".
- @cpu: renamed to `rustls_server_config_builder_set_send_tls13_tickets` to match the upstream `ServerConfig::send_tls13_tickets` field name, consistent with how `set_ignore_client_order` mirrors `ignore_client_order`. Updated the header and regenerated `api.json`.
- @cpu: added the missing `assert!(!config.is_null());` to the second build in the test.

Also renamed the test function to match and confirmed the header still regenerates cleanly with cbindgen.

cpu (approved):
Thanks

PR #636, opened 2026-04-12
Upgrade rustls-platform-verifier to 0.7, rustls 0.23.38

djc:

PR #655, opened 2026-07-14
Rustls 0.23.42, RFC 9149 API, check ALPN fix, and cbindgen update

cpu:
This started as a straight update of the rustls dep to 0.23.42 and adding the new APIs, but grew a bugfix to https://github.com/rustls/rustls-ffi/pull/636 I noticed along the way, and a `cbindgen` update so I don't have to keep remembering to use an older pinned version.

Beyond the expected change in formatting/ordering from a new `cbindgen` I also had to hack around the tree-sitter-c grammar's inability to cope with a preprocessor conditional in an enum definition in order for the `docgen` tool to not barf on the new `.h`. I think the solution I landed on isn't too gross overall and unblocks the `cbindgen` update.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -276,7 +276,7 @@ jobs:
         # reliably matched to CI. There can be non-semantic differences in
         # output between point releases of cbindgen that will fail this check
         # otherwise.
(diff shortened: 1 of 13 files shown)


djc (approved):

djc:
Nice!

